PR #12, opened 2015-07-09
added support methods for the with-structure

zack-vii:
allows for:

> with Tree('mytree',-1, 'edit') as tree:
>     ...
> instead of:
> try:
>    ...
> except:
>    tree.quit()

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,22 @@
-commit 96c366aa7c317cf76cc9d8432efd1422505f1252 (HEAD, tag: refs/tags/hudson-alpha-release-13, refs/remotes/origin/alpha, refs/heads/alpha)
+commit 1d22fc8549cb10204b5b04ed459fb6ad4571e13b (HEAD, tag: refs/tags/hudson-alpha-release-14, refs/remotes/origin/alpha, refs/heads/alpha)
+Author: Tom Fredian <twf@psfc.mit.edu>
+Date:   Thu Jul 9 14:03:34 2015 -0400
+
+    output signing dialog to stdout
+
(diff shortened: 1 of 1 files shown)


tfredian:
Thanks for the "with-structure" idea. I extended the feature to work with any method of tree opening and to close the tree after the with clause. If the tree is open for edit when the with clause completes and there are not exceptions raised it will write the changes to the tree before closing. If there are exceptions it will discard any changes to the tree.

I did not pull the java related commits as that will need to be reviewed by Gabriele. We appreciate your suggestions and implementation of those suggestions. It would be easier to pull them (I think) if you separated the commits into separate branches so we could pull the commits related to different features separately. This could also just be my lack of git expertise but in handling this pull request I pulled it into a branch and then cherry-pick'ed the python commit from this branch into the alpha branch and will suggest that Gabriele does the same with the java related commits after reviewing them.

Thanks!
Tom

zack-vii:
Hey, yes my experience with git is as weak, possibly weaker even. turned out 
that Tree.quit() is not the right way to finish up anyway. it would better be 
tree.**del**  as is closes all open sessions. quit ever raises an error if the 
tree has not been opened for edit.
What I wanted to do, is creating individual pull-requests. However, git merged 
them for me. Thank you git =)

Well, if it helps I can redo the pull-request. or you fix it quickly by hand
...
19    # support for the with-structure
20    def **enter**(self):
21        return self
22    def **exit**(self, type, value, traceback):
23 -       return self.quit()
23 +      return self.**del**
24
25    def **del**(self):
...

sorry for the inconvenience
cheers
Timo

PR #18, opened 2015-07-23
set the location of the open dialog relative to it parent

zack-vii:
Prevents, that the tree-open dialog is at a fixed position (100,100 from
the upper left corner). It now opens at a fixed relative position to the
main window's upper left corner. This way, it is much more convenient,
especially if you use large or multiple screens.


zack-vii:
rebased

PR #19, opened 2015-07-23
fixed argument to encounter a java compiler warning

zack-vii:
prevents warning during compilation:
Tree.java:693: warning: non-varargs call of varargs method with inexact argument type for last parameter;
cast to Object for a varargs call
cast to Object[] for a non-varargs call and to suppress this warning


zack-vii:
rebased

PR #23, opened 2015-08-13
_decriptor move unicode to bottom

zack-vii:
this way an error thrown by unicode part is not swallowed by try-structure without affecting the rest


zack-vii:
I just stumbled over it as a TEXT node did not return any data in python but there was clearly data in jTraverser and jScope. It only occurred if non-ascii characters where in the string as any of the German Umlaute äöüÄÖÜ ... but it seems to go deeper as __TreeGetRecord in _treeshr TreeGetRecord already returns no data.

I will close this one as I forgot to create a new branch anyway:
maybe consider PULL #27

PR #24, opened 2015-08-13
treenode.py -syntax and other fixes

zack-vii:
- unused import: threading
- missing object: Tree() -> _tree.Tree()
- redeclaration of getNumSegments(self)
- missing quotes: _treeshr -> '_treeshr'
- miss-spelled var name: R_mimport -> _mimport


tfredian:
Thanks. Didn't pull this one either but instead removed the threading import and fixed the _tree.Tree(). The rest was in the CachedTreeNode class which is obsolete. I removed that.

PR #27, opened 2015-08-13
new treat for unicode

zack-vii:
does not fail on isinstance(value,unicode) as enclosed by try-except but will throw errors within the block

### DIFF
--- a/mdsobjects/python/_descriptor.py
+++ b/mdsobjects/python/_descriptor.py
@@ -116,7 +116,10 @@ class descriptor(_C.Structure):
                 self.addToCache(value)
                 return
         try:
-          if isinstance(value,unicode):
+            isunicode = isinstance(value,unicode)
+        except:
+            isunicode = False
+        if isunicode:
             str_d=descriptor_string(value)
             d=_C.cast(_C.pointer(str_d),_C.POINTER(descriptor)).contents
             self.length=d.length
@@ -124,8 +127,7 @@ class descriptor(_C.Structure):
             self.pointer=d.pointer
             self.addToCache(value)
             return
-        except:
-          pass
+      
         if isinstance(value,str):
             str_d=descriptor_string(value)
             d=_C.cast(_C.pointer(str_d),_C.POINTER(descriptor)).contents


zack-vii:
Hey Tom,

it seems during that major update from 131 to 132, a few things got broken. for 
one, the unicode issue - pythons TreeNode.record returns <no data>. And the 
TreeNode.tree property when passed down from the node to a direct call of a 
python function.

\TEST::TOP:MYSIGNAL          :USAGE     = SIGNAL
                                                   :RECORD  = 
"testfun(TOP:MYSIGNAL)"

\TEST::TOP:MYTEXT                :USAGE   =TEXT
                                                    :RECORD = "abcäöü" = 
"abc\703\644\703\666\703\674"

## testfun.py

def testfun(node)
     return(str(node.tree))
-----------------------------eof

Tditest returns:

TREEOPEN("TEST",-1)

MYSIGNAL
'Tree("TEST",-1,"Normal")' --> alpha_release-7-0-131
"None"                                   --> alpha_release-7-0-132

MYTEXT
"abc\703\644\703\666\703\674" --> alpha_release-7-0-131
"abc\703\644\703\666\703\674" --> alpha_release-7-0-132

but in python:

import MDSplus

tree= MDSplus.Tree('test',-1)

node=tree.getNode('MYTEXT')

node.data()

"abcäöü"       --> alpha_release-7-0-131
None              --> alpha_release-7-0-132

the  .data() of the MYSIGNALL node crashes python...

I will keep digging, but maybe you arready hae a feeling where something might 
ha gone wrong.
Cheers
-Timo

zack-vii:
Good morning Tom,

if fiddled with the treenode bug a little more but for this one I will need your 
help. I turned out the order in _descriptor is important. The more generic types 
should be checked last as e.g. a Nid is a subclass of int. therefore int should 
be checked for after Nid in order to preserve the Nid properties.
I did that and now out side of python () the code work as expected (PULL #31). 
however, calling the node MYSIGNAL from inside a python environment will still 
freeze python.
I tracked it down to _tdishr. The TdiShr.TdiCompile(...) call does not return if 
a TreeNode or TreePath is involved.

So it seems its an issue with the library I will continue checking different 
systems. Right now I am using Windows 10, 64bit. I will report back if this 
issue is system-related.

cheers
-Timo

PR #29, opened 2015-08-14
optimized SimpleWaveData on MdsDataProvider

zack-vii:
Stressed out the method of tdi-variable-caching. This way the the access to the node is reduced to one time only compared to four if only Y-data is provided (LocalDataProvider). This results in a major speed up if the signal from the node is generated by a time-consuming method e.g. a python web interface.
Some code clean up.
Could be intensified as so far I did not cross a case where v_idx was not 0. Consider replacing "_jscope_"+(var_idx  ) and "_jscope_"+(var_idx+1) with _jscope_y and _jscope_x.


zack-vii:
Don't pull but consider or wait for final
sorry about that. but I just realize how the rest works
needs bug fixing

zack-vii:
There have been too many more changes.
Check them out on
https://github.com/zack-vii/mdsplus/tree/zack

If desired, I will do further pull requests

GabrieleManduchi:
Hi Timo,
   I need to make some checks when I come back to work (beginning Sepember) 
in order to make sure not to affect the new resampling mechanism of jScope.

```
           Gabriele
```

zack-vii:
Hey Gabriele,
no worries, there is no rush. I will work with the workaround.
Once you are back i might have some questions about the code.
So far, have a nice vacation. =)

-Timo

PR #32, opened 2015-08-15
javascope.bat: added compiling error handling

zack-vii:
made use of the %ERRORLEVEL% of compiling process to conditionally abort


zack-vii:
#47

PR #34, opened 2015-08-21
matlab package

zack-vii:
-added mdsgetmsg for translating tdi message ids to messages
-status and result .. throw error on error if status is not a return
value
-used 'smart indent' on all files
-compareIP is just a strcmp -> cmpstr
-NATIVEvalue can be subfunction of mdsvalue


joshStillerman:
Timo -
  can you break this up into separate requests?  A few things:
1 - the indent should be separate or not done, since it has the effect of changing every file, making it hard to understand the history.
2 - NATIVEvalue is used outside of MDSValue
3 - The mex file is deprecated and should be removed 
4 - The CompareIP should do more than a string compare.
5 - What is the ‘binary file’ ?

So if you could break up your changes into separate requests, I would be happy to pull them or respond to them.  I don’t fee git fluent enough to cherry pick them;

-Josh

joshStillerman:
Timo -
  I have been trying to deal with these changes and am finding it very difficult.  I really appreciate the work you are doing, and most of the changes are for the better.

Could you withdraw this pull request and make separate pull requests for the changes you want to make ?

A few other things -
  I really don't like that you are not indenting the lines after the function declarations. So if I was going to merge these changes, I would have to extract the good semantic/syntactic changes from your edits.  If I was more facile with GIT this might not be a problem, but....

My comment on mex did not mean to remove the mex, just not type / work on it.  Some people (hopefully no new people) are still using. it.

Thanks,

Josh

zack-vii:
sure, I thought you already closed it as I did individual pull requests. for the other changes.. this is just my private final version.. I will close this one so you don't get further mails if I change anything

PR #35, opened 2015-08-24
javamds.c: Fix 32 bit swap for getAllFrames

zack-vii:
same as for getIntegerArray

### DIFF
--- a/javamds/javamds.c
+++ b/javamds/javamds.c
@@ -783,12 +783,12 @@ JNIEXPORT jbyteArray JNICALL Java_jScope_LocalDataProvider_getAllFrames
       break;
     case 4:
       for (i = 0; i < nSamples; i++) {
-	tmp = buf[2 * i];
-	buf[2 * i] = buf[2 * i + 3];
-	buf[2 * i + 3] = tmp;
-	tmp = buf[2 * i + 1];
-	buf[2 * i + 1] = buf[2 * i + 2];
-	buf[2 * i + 2] = tmp;
+	tmp = buf[4 * i];
+	buf[4 * i] = buf[4 * i + 3];
+	buf[4 * i + 3] = tmp;
+	tmp = buf[4 * i + 1];
+	buf[4 * i + 1] = buf[2 * i + 2];
+	buf[4 * i + 2] = tmp;
       }
       break;
     }


joshStillerman:
Gabriele -
  This looks fine to me.  If you want me to deal with pulling it, let me know.

-Josh

zack-vii:
it could be pulled already..
however getAllFrames does not really support floating data (float or double) yet.
Is it possible to append the dtype value to the getInfo array. The information 
is needed to discriminate between float and int32 (both 4bit).
I would like to display int32 data as RGB(A) formatted image and floating data 
with indexed colormap.

-timo

GabrieleManduchi:
Hi Timo,
    back to work now.   Could you remind me all the pending 
actions/issues with jScope image management, so that I can proceed?

```
                             Gabriele
```

On 25/08/2015 16:11, zack-vii wrote:

> it could be pulled already..
> however getAllFrames does not really support floating data (float or 
> double) yet.
> Is it possible to append the dtype value to the getInfo array. The 
> information
> is needed to discriminate between float and int32 (both 4bit).
> I would like to display int32 data as RGB(A) formatted image and 
> floating data
> with indexed colormap.
> 
> -timo
> 
> On 8/25/2015 3:45 PM, Josh Stillerman wrote:
> 
> > Gabriele -
> > This looks fine to me. If you want me to deal with pulling it, let 
> > me know.
> > 
> > -Josh
> > 
> > > On Aug 24, 2015, at 7:19 PM, zack-vii notifications@github.com 
> > > wrote:
> > > 
> > > same as for getIntegerArray
> > > 
> > > You can view, comment on, or merge this pull request online at:
> > > 
> > > https://github.com/MDSplus/mdsplus/pull/35
> > > https://github.com/MDSplus/mdsplus/pull/35
> > > Commit Summary
> > > 
> > > javamds.c: Fix 32 bit swap for getAllFrames
> > > File Changes
> > > 
> > > M javamds/javamds.c
> > > https://github.com/MDSplus/mdsplus/pull/35/files#diff-0 (12)
> > > Patch Links:
> > > 
> > > https://github.com/MDSplus/mdsplus/pull/35.patch
> > > https://github.com/MDSplus/mdsplus/pull/35.patch
> > > https://github.com/MDSplus/mdsplus/pull/35.diff
> > > https://github.com/MDSplus/mdsplus/pull/35.diff
> > > —
> > > Reply to this email directly or view it on GitHub
> > > https://github.com/MDSplus/mdsplus/pull/35.
> > 
> > —
> > Reply to this email directly or view it on GitHub
> > https://github.com/MDSplus/mdsplus/pull/35#issuecomment-134590948.
> 
> —
> Reply to this email directly or view it on GitHub 
> https://github.com/MDSplus/mdsplus/pull/35#issuecomment-134598203.

## 

Gabriele Manduchi

Istituto Gas Ionizzati del CNR
Consorzio RFX - Associazione EURATOM/ENEA sulla Fusione
Corso Stati Uniti 4, 35127 Padova - Italy
ph +39-049-829-5039/-5000 fax +39-049-8700718
mailto:gabriele.manduchi@igi.cnr.it, http://www.igi.cnr.it

zack-vii:
Hi Gabriele,
hope you had some good time.
Well, I still cannot compile the MDSplus stuff for windows yet (just posix) as I 
have difficulties to get the right development environment. How is yours set up? 
Are you using linux + mingw. if you could share your setup I could test a few 
things myself.

This is a list of things that could be discussed and checked for possible fixes. 
I will start with jScope. But I kept on going so it is more complete as intended.

Anyway, I am working on expanding the capabilities of jScope on the image 
display. However, since my java programming experience is still growing, I did 
not get the time yet to structure my modifications. You can check on the 
progress on my git (https://github.com/zack-vii/mdsplus/tree/jScope). Once W7X 
is running and i have more time I am happy to craft some pull requests for the 
notable mods.
So far the MdsDataprovider was the focus, but I expanded to the 
LocalDataProvider, which is where I discovered the byte swapping issue.

To fully integrate the modifications it would be nice if the native getinfo 
would return the dtype of the data as additional parameter.
This is in order to be able to differentiate betwee floating point and integer 
values of the same bitsize. (float vs. int32)  I suppose this is not necessarily 
an issue with the fixed javamds.dll as it will export float as double  (8bytes) 
but still the FrameData.Bitmap_... mode would be much more straight forward.

The check for the number of dimensions inside the library prevents reading 2D 
data d(t,x) which would be possible to display as image as well.

Further, I would like to read 4D int8 or int16 data (timer, width, height, 
(3=RGB tupel)) and display it as coloured images (scalable RGBA, 8bit per channel).
since getAllFrames check for dimensions it will return an error. I guess a 
native function that exports the raw byte plus a richer getinfo would do the trick.
getinfo will at least need to contain the dType, and the full shape plus 
whatever is needed for getAllFrames.
int32 [dtype, N=ndims, dim0, dim2, ...,dimN-1, <rest>] rest could contain 
pixelBitSize and the list of times of length dim0;
  better would be to define a proper structure with the respected fields. 
However this depends how easy this is to implement for the c-java-interface.

Segmented data seems to be messed up in general:
If you store e.g. 3D data as block data (not segmented) it will be stored 
(t,x,y) with dim_of($,i) (i=0:t,1:x,2:y).
however, if you store it segmented (makeSegment==buildSegment) it will be stored 
(x,y,t) but still with dim_of($,i) (i=0:t,1:x,2:y).
This get broken even more if you store it as one time instance per segment. the 
data read as segment will become a (Nx,Ny,1) array. The dimensions seem to be 
stored at first as dim_of($,i) (i=0:x,1:y). But since the time dimension is 
treated extra it overrides dim_of($,0) with the time data (shape=[1]).
- I think it is not good to have a discrepancy between the shape of the data 
  depending of its property being segmented or not.
  Independent of is segmentation the data should match the same format (t,x,y,...) 
  or (x,y,...,t), the prior works best with c as the serializing in c seems to 
  work that way. The same seems to be true for python's numpy and java. well, in 
  matlab the ordering is just the other way round, but the user will never notice 
  apart form the fact that he/she will have to use squeeze here and there if one 
  time instance is selected.
  > Anyway it seems to be tricky to decide to fix this  issue as users might have gotten user to this already. Also, if they used 
  workarounds, transposing the data before storing, their data is stored in a 
  wrong format. This could be detected by checking the data.shape against the 
  ndims and dim_of.lengths.

-some crashes on win64bit
python:        TreeNode.bool();  # ? whatever this is for
python:        MDSplus.TdiExecute('pytest(1)')

## %MDS_DIR%\pytest.py

def pytest(arg):
     return(arg)
-EOF-------------------------
it crashes on the TdiShr.TdiExecute call line 64 in _tdishr (execute = 
TdiShr.TdiExecute #ok\nstatus = execute(*argin) #crash in library)

wfevent crashes on an occurring event if the source of the event is the local 
machine (using mdsconnect to a server: and wfevent works fine)
wfevent seems to chrash on the library layer as any interface crashes (matlab, 
python, tditest,...)
tditest:     wfevent('myevent') on event occurring at 
MdsShr->MDSWfeventTimed(_event,_blen,ref(_data),ref(_dlen),val(_timeout));

As you can see, lots to do. No need to solve everything at once ;).
But maybe you have a good idea for one or the other and you can start up with 
some easier ones.

cheers
-Timo

GabrieleManduchi:
Hi Timo,

> Well, I still cannot compile the MDSplus stuff for windows yet (just 
> posix) as I
> have difficulties to get the right development environment. How is 
> yours set up?
> Are you using linux + mingw. if you could share your setup I could 
> test a few
> things myself.
> 
> We don't compile anymore MDSplus under windows, rather we use 
> linux+mingw integrated in Docker.  Tom may give you more details on how 
> setting this up (personally I don't compile the windows version since a 
> long time)
> 
> This is a list of things that could be discussed and checked for 
> possible fixes.
> I will start with jScope. But I kept on going so it is more complete 
> as intended.
> 
> Anyway, I am working on expanding the capabilities of jScope on the image
> display. However, since my java programming experience is still 
> growing, I did
> not get the time yet to structure my modifications. You can check on the
> progress on my git (https://github.com/zack-vii/mdsplus/tree/jScope). 
> Once W7X
> is running and i have more time I am happy to craft some pull requests 
> for the
> notable mods.
> So far the MdsDataprovider was the focus, but I expanded to the
> LocalDataProvider, which is where I discovered the byte swapping issue.
> 
> More in general, I found preferable avoiding JNI in jScope and using the 
> MdsDataProvider class instead. This holds especially for large data, 
> avoiding having the same process handle MDSplus data acquisition and 
> data  visualization. Using MdsDataProvider data access is confined in 
> the mdsip server, allowing for data size optimization for large signals. 
> Moreover, using MdsDataProvider jScope becomes a pure java application 
> and does not require the installation of other MDSplus components. 
> Anyway, LocalDataProvider must work, too.
> 
> To fully integrate the modifications it would be nice if the native 
> getinfo
> would return the dtype of the data as additional parameter.
> This is in order to be able to differentiate betwee floating point and 
> integer
> values of the same bitsize. (float vs. int32) I suppose this is not 
> necessarily
> an issue with the fixed javamds.dll as it will export float as double 
> (8bytes)
> but still the FrameData.Bitmap_... mode would be much more straight 
> forward.
> 
> The check for the number of dimensions inside the library prevents 
> reading 2D
> data d(t,x) which would be possible to display as image as well.
> 
> Further, I would like to read 4D int8 or int16 data (timer, width, 
> height,
> (3=RGB tupel)) and display it as coloured images (scalable RGBA, 8bit 
> per channel).
> since getAllFrames check for dimensions it will return an error. I 
> guess a
> native function that exports the raw byte plus a richer getinfo would 
> do the trick.
> getinfo will at least need to contain the dType, and the full shape plus
> whatever is needed for getAllFrames.
> int32 [dtype, N=ndims, dim0, dim2, ...,dimN-1, <rest>] rest could contain
> pixelBitSize and the list of times of length dim0;
> better would be to define a proper structure with the respected fields.
> However this depends how easy this is to implement for the 
> c-java-interface.

It is possible to modify the native getInfo() method so that it returns 
a Java object whose fields may be the following:

-dtype
-dimensions (as a java array, therefore the number of dimensions is 
implicit in it)
-pixelbitsize
-times array (if defined, otherwise 0..nTimes-1)

and then use this information in Java. In 3-4 weeks I think I can find 
the time to do this change which can be made  neutral in the alpha 
MDSplus release, but which would allow you extending imaging 
capabilities (and, of course, report them in MDSplus alpha when stable 
in your application)

> Segmented data seems to be messed up in general:
> If you store e.g. 3D data as block data (not segmented) it will be stored
> (t,x,y) with dim_of($,i) (i=0:t,1:x,2:y).
> however, if you store it segmented (makeSegment==buildSegment) it will 
> be stored
> (x,y,t) but still with dim_of($,i) (i=0:t,1:x,2:y).
> This get broken even more if you store it as one time instance per 
> segment. the
> data read as segment will become a (Nx,Ny,1) array. The dimensions 
> seem to be
> stored at first as dim_of($,i) (i=0:x,1:y). But since the time 
> dimension is
> treated extra it overrides dim_of($,0) with the time data (shape=[1]).
> - I think it is not good to have a discrepancy between the shape of 
>   the data
>   depending of its property being segmented or not.
>   Independent of is segmentation the data should match the same format 
>   (t,x,y,...)
>   or (x,y,...,t), the prior works best with c as the serializing in c 
>   seems to
>   work that way. The same seems to be true for python's numpy and java. 
>   well, in
>   matlab the ordering is just the other way round, but the user will 
>   never notice
>   apart form the fact that he/she will have to use squeeze here and 
>   there if one
>   time instance is selected.

Dimensions are always a tricky issue. MDSplus handles dimensions in the 
fortran way (that is, column first order). This explains why in 
segmented images time is the last dimension. In fact, segments are 
appended each other along time during continuous data acquisition. It it 
were C ordering, time would be the first dimension, but it becomes the 
last one in MDSplus TDI (i.e. dim_of). If I am not wrong, this is the 
ordering we have also used for our older non segmented frames, developed 
in advance than segments. If you write a non segmented 3D array in 
MDSplus you should define t as the last dimension and store the array as 
[frame at time 0, frame at time 1......]
I any case strongly advise using ALWAYS segments for images, and 
therefore time becomes always the last dimension (fortran style).

> > Anyway it seems to be tricky to decide to fix this issue as users 
> > might have gotten user to this already. Also, if they used
> > workarounds, transposing the data before storing, their data is stored 
> > in a
> > wrong format. This could be detected by checking the data.shape 
> > against the
> > ndims and dim_of.lengths.
> 
> -some crashes on win64bit
> python: TreeNode.bool(); # ? whatever this is for
> python: MDSplus.TdiExecute('pytest(1)')
> 
> ## %MDS_DIR%\pytest.py
> 
> def pytest(arg):
> return(arg)
> -EOF-------------------------
> it crashes on the TdiShr.TdiExecute call line 64 in _tdishr (execute =
> TdiShr.TdiExecute #ok\nstatus = execute(*argin) #crash in library)

Tom is the developer of the python part, so he may have a look at it

> wfevent crashes on an occurring event if the source of the event is 
> the local
> machine (using mdsconnect to a server: and wfevent works fine)
> wfevent seems to chrash on the library layer as any interface crashes 
> (matlab,
> python, tditest,...)
> tditest: wfevent('myevent') on event occurring at
> MdsShr->MDSWfeventTimed(_event,_blen,ref(_data),ref(_dlen),val(_timeout));
> 
> Does it happen also on Linux?   Are you using UDP events (i.e. 
> UDP_EVENTS env variable set)?

```
                                                     Gabriele
```

## 

Gabriele Manduchi

Istituto Gas Ionizzati del CNR
Consorzio RFX - Associazione EURATOM/ENEA sulla Fusione
Corso Stati Uniti 4, 35127 Padova - Italy
ph +39-049-829-5039/-5000 fax +39-049-8700718
mailto:gabriele.manduchi@igi.cnr.it, http://www.igi.cnr.it

GabrieleManduchi:
Hi Timo,

```
 I use docker images to build all but the Macosx distributions. To 
```

build the windows distribution (without the MdsObjectsCppShr-VS.dll) you 
can do the following on a linux system with an up-to-date docker-io 
installation (I use fc22 now):

# docker run --rm  -v /yoursource:/source -v /installerdir:/installer

mdsplus/docker:build_windows build

and this would use the MDSplus source in /yoursource and build a windows 
installer .exe in /installerdir.

If you want to build a particular branch you can do:

# docker run --rm  -v /installerdir:/installer

mdsplus/docker:build_windows build alpha

If you want to just build the libraries and executables you can do:

# docker run --rm -v /yoursource:/source -v /yourbuildroot:/buildroot

mdsplus/docker:build_windows build

Hope this helps.

Cheers,
Tom

zack-vii:
thanks for the quick response,

I totally agree on the DataProvider thing. I did not like the native solution 
either. well maybe there is some advantage on using "dll memory" instead of 
"java memory". Especially since java does all the swapping for you.

Thanks for having a look into the getInfo thing. its not urgent, so next month 
will be fine.

About the dimensions:

> If you write a non segmented 3D array inMDSplus
> you should define t as the last dimension and store the array as
> [frame at time 0, frame at time 1......]

after some testing the dim_of($,n) follows the python scheme
dim             0    1    2
dim_of        [time,row ,col ]
python(numpy):[time,row ,col ]
TDI array:    [col ,row ,time]
MATLAB array: [col ,row ,time]  -> that's good

data as it is stored in datafile:
"abcdefghigklmnopqrstuvwxyz."

data as 3x3 images:
t=0   t=1   t=2
abc   jkl stu
def   mno   vwx
ghi   pqr   yz.

## my test script:

import MDSplus,numpy
"""
#create tree
with MDSplus.Tree('TEST',2,'new') as tree:
     tree.addNode('MAT','SIGNAL').setCompressOnPut(False)
     tree.addNode('SEG','SIGNAL')
     tree.write()
"""

shape= [8,6,4]
data = numpy.array(range(shape[0]_shape[1]_shape[2])).astype('uint8')
data = data.reshape(shape).copy()
time = numpy.array(range(shape[0])).astype('uint8')
numpyInFull     = str(data.shape)
with MDSplus.Tree('TEST',2) as tree:
     node = tree.getNode('MAT')
     node.putData(MDSplus.Signal(data))
     tdiFull         = str(MDSplus.TdiExecute('SHAPE($)',(node,)))
     numpyOutFull    = str(node.data().shape)
     node = tree.getNode('SEG')
     for n in [8,4,2,1]:
         node.deleteData()
         N = int(shape[0]/n)
         seg= [[]]_N
         for i in range(N):
             start = i_n
             end   = i*n+n;
             seg[i] = data[start:end]
             dim = time[start:end]
             node.makeSegment(start,end,dim,seg[i])

```
     lendim          = str(dim.shape)
     numpyInSeg      = str(seg[0].shape)
     tdiSeg          = str(MDSplus.TdiExecute('SHAPE(GetSegment($,0))',(node,)))
     numpyOutSeg     = str(node.getSegment(0).data().shape)
     tdiSegFull      = str(MDSplus.TdiExecute('SHAPE($)',(node,)))
     numpyOutSegFull = str(node.data().shape)
     print("normal:= "+numpyInFull+"->"+tdiFull+"->"+numpyOutFull)
     print("segmented:= "+numpyInFull+"->"+tdiSegFull+"->"+numpyOutSegFull)
     print("segments:= {"+numpyInSeg+"->"+tdiSeg+"->"+numpyOutSeg+"}x"+str(i+1))
```

---

ouput:

normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[8,6,8]->(8, 6, 8)
segments:= {(8, 6, 4)->[8,6,4]->(4, 6, 8)}x1
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
segments:= {(4, 6, 4)->[4,6,4]->(4, 6, 4)}x2
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[2,6,8]->(8, 6, 2)
segments:= {(2, 6, 4)->[2,6,4]->(4, 6, 2)}x4
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[1,6,8]->(8, 6, 1)
segments:= {(1, 6, 4)->[1,6,4]->(4, 6, 1)}x8

In fact there is a mistake in the python code. If passed as descriptor the 
vector is transposed in _descriptor.py,106 a=descriptor_a(value.T)
However, if makeSegment is used, the array is passed directly to descriptor_a(). 
This call lacks the transposition as well as the contiguous check.(pull#45)

fixed output:
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
segments:= {(8, 6, 4)->[4,6,8]->(8, 6, 4)}x1
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
segments:= {(4, 6, 4)->[4,6,4]->(4, 6, 4)}x2
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
segments:= {(2, 6, 4)->[4,6,2]->(2, 6, 4)}x4
normal:=    (8, 6, 4)->[4,6,8]->(8, 6, 4)
segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
segments:= {(1, 6, 4)->[4,6,1]->(1, 6, 4)}x8

I guess that solves some problems with Segmented data.

Cheers
I call it a day for today.
-Timo

GabrieleManduchi:
Hi Timo,
    another reason for using MdsDataProvider is that the amount of 
required memory has been greatly optimized in the latest version of 
jScope. You may note that when displaying large signals and then 
zooming, the resolution of the signal is dynamically updated. This 
because MdsDataProvider agrees on the mdsip server on a resampled 
version of the signal (not exactly resampled in order to avoid aliasing 
problems, rather taking the minumum and the maximum value for a given 
range of x values), and dynamically retrieves portions of signals at 
higher resolution when required (e.g. after a zoom). Also for images 
MdsDataProvider has been optimized, keeping in java memory only a subset 
of the image frames and dynamically retrieving other frames when 
required.  All these advantages are useless using the LocalProvider 
because in any case the whole signal has to be read. Well, also the 
mdsip server has to read whole signals in order to compress them and 
sending to MdsDataProvider, but the assumption made is that mdsip is 
likely to be run by a server with enough memory, while jScope typically 
runs on a user laptop.

Ok, I will change LocalDataProvider.getInfo() in few weeks.

Tom, could you have a look at the problem reported by Timo about arrays 
in python?

```
     Gabriele
```

On 01/09/2015 04:47, zack-vii wrote:

> thanks for the quick response,
> 
> I totally agree on the DataProvider thing. I did not like the native 
> solution
> either. well maybe there is some advantage on using "dll memory" 
> instead of
> "java memory". Especially since java does all the swapping for you.
> 
> Thanks for having a look into the getInfo thing. its not urgent, so 
> next month
> will be fine.
> 
> About the dimensions:
> 
> > If you write a non segmented 3D array inMDSplus
> > you should define t as the last dimension and store the array as
> > [frame at time 0, frame at time 1......]
> 
> after some testing the dim_of($,n) follows the python scheme
> dim 0 1 2
> dim_of [time,row ,col ]
> python(numpy):[time,row ,col ]
> TDI array: [col ,row ,time]
> MATLAB array: [col ,row ,time] -> that's good
> 
> data as it is stored in datafile:
> "abcdefghigklmnopqrstuvwxyz."
> 
> data as 3x3 images:
> t=0 t=1 t=2
> abc jkl stu
> def mno vwx
> ghi pqr yz.
> 
> ## my test script:
> 
> import MDSplus,numpy
> """
> #create tree
> with MDSplus.Tree('TEST',2,'new') as tree:
> tree.addNode('MAT','SIGNAL').setCompressOnPut(False)
> tree.addNode('SEG','SIGNAL')
> tree.write()
> """
> 
> shape= [8,6,4]
> data = numpy.array(range(shape[0]_shape[1]_shape[2])).astype('uint8')
> data = data.reshape(shape).copy()
> time = numpy.array(range(shape[0])).astype('uint8')
> numpyInFull = str(data.shape)
> with MDSplus.Tree('TEST',2) as tree:
> node = tree.getNode('MAT')
> node.putData(MDSplus.Signal(data))
> tdiFull = str(MDSplus.TdiExecute('SHAPE($)',(node,)))
> numpyOutFull = str(node.data().shape)
> node = tree.getNode('SEG')
> for n in [8,4,2,1]:
> node.deleteData()
> N = int(shape[0]/n)
> seg= [[]]_N
> for i in range(N):
> start = i_n
> end = i*n+n;
> seg[i] = data[start:end]
> dim = time[start:end]
> node.makeSegment(start,end,dim,seg[i])
> 
> lendim = str(dim.shape)
> numpyInSeg = str(seg[0].shape)
> tdiSeg = str(MDSplus.TdiExecute('SHAPE(GetSegment($,0))',(node,)))
> numpyOutSeg = str(node.getSegment(0).data().shape)
> tdiSegFull = str(MDSplus.TdiExecute('SHAPE($)',(node,)))
> numpyOutSegFull = str(node.data().shape)
> print("normal:= "+numpyInFull+"->"+tdiFull+"->"+numpyOutFull)
> print("segmented:= "+numpyInFull+"->"+tdiSegFull+"->"+numpyOutSegFull)
> print("segments:= 
> 
> ## {"+numpyInSeg+"->"+tdiSeg+"->"+numpyOutSeg+"}x"+str(i+1))
> 
> ouput:
> 
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[8,6,8]->(8, 6, 8)
> segments:= {(8, 6, 4)->[8,6,4]->(4, 6, 8)}x1
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segments:= {(4, 6, 4)->[4,6,4]->(4, 6, 4)}x2
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[2,6,8]->(8, 6, 2)
> segments:= {(2, 6, 4)->[2,6,4]->(4, 6, 2)}x4
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[1,6,8]->(8, 6, 1)
> segments:= {(1, 6, 4)->[1,6,4]->(4, 6, 1)}x8
> 
> In fact there is a mistake in the python code. If passed as descriptor 
> the
> vector is transposed in _descriptor.py,106 a=descriptor_a(value.T)
> However, if makeSegment is used, the array is passed directly to 
> descriptor_a().
> This call lacks the transposition as well as the contiguous 
> check.(pull#45)
> 
> fixed output:
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segments:= {(8, 6, 4)->[4,6,8]->(8, 6, 4)}x1
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segments:= {(4, 6, 4)->[4,6,4]->(4, 6, 4)}x2
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segments:= {(2, 6, 4)->[4,6,2]->(2, 6, 4)}x4
> normal:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segmented:= (8, 6, 4)->[4,6,8]->(8, 6, 4)
> segments:= {(1, 6, 4)->[4,6,1]->(1, 6, 4)}x8
> 
> I guess that solves some problems with Segmented data.
> 
> Cheers
> I call it a day for today.
> -Timo

## 

Gabriele Manduchi

Istituto Gas Ionizzati del CNR
Consorzio RFX - Associazione EURATOM/ENEA sulla Fusione
Corso Stati Uniti 4, 35127 Padova - Italy
ph +39-049-829-5039/-5000 fax +39-049-8700718
mailto:gabriele.manduchi@igi.cnr.it, http://www.igi.cnr.it

PR #46, opened 2015-09-01
updated the help text

zack-vii:
explained how the error handling works


tfredian:
It looks like this pull request was made before the rebasing of alpha so desn't appear to be in a state where it should be pulled.

PR #51, opened 2015-09-03
pragma once

joshStillerman:
However
  Still define FILENAME_H for legacy references to these files

Remove ifdef VMS since we don't build for that anymore

### DIFF
--- a/include/dbidef.h
+++ b/include/dbidef.h
@@ -1,9 +1,9 @@
+#pragma once
 /*  VAX/DEC CMS REPLACEMENT HISTORY, Element DBIDEF.H */
 /*  *3     1-APR-1991 17:29:45 TWF "Add readonly open" */
 /*  *2    16-FEB-1990 10:28:18 TWF "Add itmlst struct" */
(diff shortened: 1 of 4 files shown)


joshStillerman:
I did remove them originally, but then thought, what if someone with an application (this is a public include) was relying on it.  I found stack overflow that said you should not only leave in the defines but also leave in the #ifdefs.  This seemed too extreme for me though.

tfredian:
MDSplus developers, what should we do regarding #pragma once? Here are some options:
1) Add #pragma once to all headers leaving the rest of the content the same (i.e. leave the #ifdef THIS_INCLUDE, #define THIS_INCLUDE, ... #endif)
2) Add #pragma once to all headers and remove the macro test and leave the #define.
3) Add #pragma once to all headers and remove the macro test and #42 
4) Only start using #pragma once on new headers leaving the old headers the same.

GabrieleManduchi:
I agree with Josh, I would use the pragma once only for new components, just to avoid side effects (highly unlikely, but....)

joshStillerman:
Actually, I would like to commit this change to these 4 public includes,  
a) using #pragma onece
b) defining FILENAME_H   (in case anyone is looking for this)

PR #52, opened 2015-09-06
Configure updates

kgerickson:
Updates to the configure files:
Regenerate the javadevices Makefile.in, which appears to be out of sync in a trivial way.

Update the build system to remove unused configure functions:
- choose a semaphore header
- test for shmat

Update the default autoconf to 2.69 since that's what was being used anyway, to avoid large scale configure changes

Rearrange some of the tests so that the java section is contiguous.  For some reasons, the tests for visual studio, python, and valgrind were all inserted into the middle of the java tests.


tfredian:
Looks good to me. I did a couple builds with it and they seemed to work.

PR #53, opened 2015-09-07
fix: long, basestring

zack-vii:
you need to redefine long and basestring conditional. as long and
basestring does not exist in python 3


joshStillerman:
Will test/merge tomorrow
-Josh

PR #55, opened 2015-09-08
Python minorfixes

zack-vii:
just small stuff


tfredian:
I'm pulling this request but currently the tdicompile.py as well as the lex and yacc modules are not used. They are left over from the beginning of a project investigating the port of tdishr to pure python which subsequently got moved over to a separate repository and will likely be disposed of as the latest thoughts are to rewrite tdishr in a compiled language such as c++.

PR #57, opened 2015-09-08
setevetn: cast to byte before reaching down to MDSSHR->MDS$EVENT (rebased)

zack-vii:
if not cast to Byte first setevent does a mix up length of original
vector but byte sequence of flattened vector.
e.g.: Int32[1,2,3,4,5] -> [1,0,0,0 , 2]

> wfevent crashed at that time if handled locally on Windows 64bit, it worked fine while mdsconnect-ed
> 
> now it works but throws error if no byte array is provided on setevent (on python 2.7 and 3.3,...; win and unix)
> 
> Traceback (most recent call last):
> File "<stdin>", line 1, in <module>
> File "python\Lib\site-packages\MDSplus\event.py", line 93, in wfevent
>   return MDSWfeventTimed(event,timeout).deserialize()
> File "python\Lib\site-packages\MDSplus\mdsarray.py", line 195, in deserialize
>   return _data.Data.deserialize(self)
> File "python\Lib\site-packages\MDSplus\mdsdata.py", line 806, in deserialize
>   return _mimport('_mdsshr',1).MdsSerializeDscIn(data)
> File "python\Lib\site-packages\MDSplus_mdsshr.py", line 120, in MdsSerializeDscIn
>   raise MdsException(MdsGetMsg(status))
> MDSplus._mdsshr.MdsException: %LIB-F-INVSTRDES, Invalid string descriptor
> 
> should return [] or None

### DIFF
--- a/tdi/setevent.fun
+++ b/tdi/setevent.fun
@@ -2,7 +2,7 @@ public fun SetEvent(in _eventname,optional in _eventdata)
 {
   if (vms())
   {
-    _edata = present(_eventdata) ? ((size(_eventdata) < 12) ? [_eventdata,zero(12,0b)] : _eventdata) : zero(12,0b);
+    _edata = present(_eventdata) ? ((size(_eventdata) < 12) ? [BYTE(_eventdata),zero(12,0b)] : BYTE(_eventdata)) : zero(12,0b);
     _status = MDSSHR->MDS$EVENT(_eventname,_edata);
   }
   else


tfredian:
Hi Timo,

It seems highly unlikely that your pull request would change any behavior. The vms() function should always return false (even if you ran it today on OpenVMS) since we no longer support OpenVMS with the current code base. That means the lines of code you changed should never even be execute. So if this seemed to fix something it is likely it did so by moving memory around a bit.

-tom

zack-vii:
No, I meant, when I was originally requesting this pull, I got a crash. So something that has been changed in the main trunk since then now either causes an error before the crash is triggered or partly fixed the reason for the crash.

zack-vii:
However, the typecast is still necessary as otherwise the behaviour is just messed up. 
I just open a new pull request explaining the issue #71

PR #58, opened 2015-09-08
Remove manual setting of executable extension, defer to EXEEXT

kgerickson:
Autoconf runs tests that determine the executable extension as part of running any compiler macro, such as AC_PROC_CC.  This gets stored in the output variable @EXEEXT@.  We want to use what autoconf provides instead of testing for MINGW platforms and manually setting .exe as the extension.


tfredian:
Cool| I'm sure there are lots of autoconf tricks that I am/was not aware of or were added since many of the Makefile.in's were first put together.

PR #60, opened 2015-09-09
alpha python3 version module

zack-vii:
version module: a better way of checking the python version.
The purpose of this helper module is to supply tools and type
substitutes that are used to generate version specific code.
Goal is to generate code that work on both python2x and python3x.

PS:  what is the range of python versions that are meant to be supported


tfredian:
This looks good. I tested it on our oldest supported python (2.4) on rhel5 and it worked fine there.

PR #68, opened 2015-09-09
Add testing to the build script used by the automated builds. Test na…

tfredian:
…mes can be listed as arguments to the docker run mdsplus/docker:build_xxx commands to specify which tests to do. A special stdtests environment variable can be used to specify a standard list of tests for the platform and then use the special stdtests test name to do the tests specified by the environment variable. Not all tests may be possible on some platforms so the stdtests environment variable is specifed in the docker image for each platform being built.

### DIFF
--- a/deploy/build_redhat_mdsplus
+++ b/deploy/build_redhat_mdsplus
@@ -1,8 +1,17 @@
 #!/bin/bash
 set -o verbose
 set -e 
+#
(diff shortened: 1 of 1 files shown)


tfredian:
I am merging the build related requests myself for now to try to get the automated builds working again. I plan to add more documentation on the build system and also encourage others to become familiar with docker and experiment with the build images.

PR #71, opened 2015-09-09
setevetn: cast to byte before reaching down to MDSSHR->MDS$EVENT

zack-vii:
If the array is not type-cast to Byte, the event system messes up:
Length of original vector but byte sequence of flattened vector.
 e.g.: 
an event is triggered using

> setevent('e',Int32[1,2,3,4,5])
> 
> wfevent('e')

returns

> [1,0,0,0 , 2]

but should return

> [1,2,3,4,5]

### DIFF
--- a/tdi/setevent.fun
+++ b/tdi/setevent.fun
@@ -2,7 +2,7 @@ public fun SetEvent(in _eventname,optional in _eventdata)
 {
   if (vms())
   {
-    _edata = present(_eventdata) ? ((size(_eventdata) < 12) ? [_eventdata,zero(12,0b)] : _eventdata) : zero(12,0b);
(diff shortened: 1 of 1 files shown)


tfredian:
I guess you didn't understand why I closed this in the first place. The fix you made never would have been executed because that section only would get executed if the vms() function returned non-zero which used to be the case when running on OpenVMS. I'm closing this again and I issued another pull request removing the vms test completely.

zack-vii:
OK I get it. sure. I guess when I tested it, I added the BYTE also in the else block, However I had to jungle a bit creating a proper branch for the pull-request and forgot it later.
FYI: it seems the crash is back. the error prevented it to happen

PR #74, opened 2015-09-10
Revert "Unsigned short data type management added"

joshStillerman:
This reverts commit ce08e861c806cf2fb92b84fc9ba7b7425de3b3ef.

This change introduced a compile error and made the indent of the code wrong

### DIFF
--- a/javascope/jScope/MdsConnection.java
+++ b/javascope/jScope/MdsConnection.java
@@ -268,45 +268,48 @@ public class MdsConnection
 
         MdsMessage message = receiveThread.GetMessage();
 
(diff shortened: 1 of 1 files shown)


tfredian:
This seems to be ok now but it is unclear how the problem happened in the first place. Perhaps Gabriele did a pull into a local alpha which had diverged from github alpha. Gabriele should verify that the contents of MdsConnection.java is what he wanted in the first place and pull this request. Since this prevents MDSplus from building this should be done ASAP.

GabrieleManduchi:
Merged now, it seems to be ok, really I cannot understand what went wrong before.....

PR #79, opened 2015-09-11
added bytes to basestring substitute

zack-vii:
bytes should be part of basestring as bytes was the same as str in py2


tfredian:
Tested

PR #80, opened 2015-09-11
Fix call to MDSWfeventTimed in wfevent.fun. The buffer lenth is an in…

tfredian:
…teger passed by value not address


tfredian:
This was my bad. I had fixed this problem in the stable branch a year ago for a bug fix but never applied it to the alpha branch. Surprised it took this long to show up!

PR #81, opened 2015-09-11
_tdishr: unite TdiFunctions

zack-vii:
since all tdi functions have the same structure the arguments can be
united in one shared helper function


tfredian:
Nice cleanup . Tested.

PR #82, opened 2015-09-12
Python tree treeshr.TreeFreeDbid ?

zack-vii:
not sure what exactly TreeFreeDbid does but this patch fixes an TdiException that occurred on 64bit python when attempting to compile treenodes
-TreeFreeDbid is defined in _treeshr and need its prefix
-ctx is private and needs its self. prefix
-replaced mimport('_treeshr',1) with on shared _treeshr


tfredian:
Tested

PR #83, opened 2015-09-12
return empty TreeNodeArray instead of None

zack-vii:
-Descendants,Members,Children return empty array if None
This allows for for-loops

> for desc in node.getDescendants:
>   do stuff with desc


tfredian:
Good idea!

PR #90, opened 2015-09-15
5. new version on mdsarray

zack-vii:
-numpy _np; ctypes _C
-numpy dtype '|S' vs. '<U' according to str by pyver

### DIFF
--- a/mdsobjects/python/mdsarray.py
+++ b/mdsobjects/python/mdsarray.py
@@ -1,4 +1,5 @@
-import numpy,ctypes#,copy
+import numpy as _np
+import ctypes as _C
 
 if '__package__' not in globals() or __package__ is None or len(__package__)==0:
   def _mimport(name,level):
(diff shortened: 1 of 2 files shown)


tfredian:
Hi Timo,

It seems like the new handling of strings at some point broke things with python3 that were working before. A simple example is:

python3 stable (correct):

> > > Data.decompile(StringArray(['XX','YY','ZZZ']))
> > > '["XX ","YY ","ZZZ"]'

python3 alpha (wrong):
Data.decompile(StringArray(['XX','YY','ZZz']))
'["X\000\000\000X\000\000\000\000\000\000\000","Y\000\000\000Y\000\000\000\000\000\000\000","Z\000\000\000Z\000\000\000z\000\000\000"]'

I'm not up to speed on the various changes you made using various numpy methods for dealing with unicode etc. 

StringArrays in MDSplus are not a list of zero terminated strings. They are blank padded strings all of the same length and contiguous in memory. For example: ['XX','YY','ZZZ'] would be layed out in memory as "XX YY ZZZ" and would be described by a descriptor with length 3 and arsize 9. So the first element in the array would be "XX " or "XX" padded with one blank.

I ran into this problem doing "python3 setup.py test" in the python source directory. The tests there are by no means exhaustive but the do test many of the methods.

Thanks,
Tom

zack-vii:
Hi Tom,
I ran in the same problem I guess I can fix it in version. just to confirm
StringArray(['X','YY','ZZZ']) being actually ["X  ","YY ","ZZZ"] is the correct 
representation (decompile is trimming the leading spaces)

I had a solution .. maybe I was too late pushing it online.. I fixed around the 
time your began to merge. I will sort it out, thanks.

tfredian:
Not sure I understand your question. Decompile should not trim leading 
spaces. StringArrays are right padded with spaces to make all the 
strings in the array the same length. 
Data.decompile(StringArray(['X','YY','ZZZ')) should be 
["X--","YY-","ZZZ"] where the - are spaces.

-tom

zack-vii:
actually my fix was in _descriptor however only applying to scalar strings
I fixed it in #100

PR #96, opened 2015-09-15
Indentation error in TreeFindNodeTags caused tags to only return one …

tfredian:
…tags when node actually has multiple tags


tfredian:
The TreeFindNodeTags does not work with python3. For example on python2 the following returns the expected answer:

> > > Tree('twf',-1).getNode('gub').tags
> > > ["FOO","BAR"]

but with python3 it produces:

> > > Tree('twf',-1).getNode('gub').tags
> > > ["F\000\000\000O\000\000\000O\000\000\000","B\000\000\000A\000\000\000R\000\000\000"]

PR #97, opened 2015-09-15
Fix TreeFindNodeTags for python3

tfredian:
Removed the astype call which fixed it for python3 and still works with python2.


tfredian:
I was confused. This fix did not work for python3.

PR #100, opened 2015-09-16
Fix for python3 StringArray

zack-vii:
typecast numpy array of string to bytes for descriptor
prevents strings to be passed to libraries as unicode


tfredian:
Great! Thanks. It tested ok with python2 and 3.

PR #103, opened 2015-09-16
Add the use of statusToException to raise appropriate MDSplus excepti…

tfredian:
…ons based on status values returned from MDSplus library routines


joshStillerman:
As per our discussion:
  How about making separate exception classes from each XML.  They could all subclass an MdsExecption that would know how to 'glue them together' ( if this is even needed ). 

So MdsExeption.getException(status)  would look up the code in a merged dictionary from all of the separate ones.  (or look it up in each of them), How would / could it get the list of them ?

To merge dicts see discussion:

http://stackoverflow.com/questions/38987/how-can-i-merge-two-python-dictionaries-in-a-single-expression

When something returns a status - could it just return an exception instead (probably not)

tfredian:
I changed the gen_messages.py script to generate a separate xxxxExceptions.py module for every xxxx_messages.xml status definition file. The statusToException(status) function will return the appropriate exception instance.

PR #104, opened 2015-09-17
Fix configure.h.in to be generated by autoheader

AndreaRigoni:
Hi All / Tom, 
I noted that the autoheader did not work anymore for the use of AC_DEFINE without comments that is no more accepted by autoheader (my version is 2.69). It never worked to me actually I was using just the committed template file configure.in.h ..
I saw that you edited it up several times manually .. so, how should I use it? via autoheader or manually?

I have some checks to add in there so I fixed issues in configure.ac that prevented autoheader to run.
I also added a specific section to configure.ac where we can add something that is written verbatim into the config.h.in.

Can you take a look of the changes?

Andrea


tfredian:
Hi Andrea,

I took a quick look and didn't see anything obviously wrong. I tested configure and a make and all worked so I went ahead and accepted the pull request. I did not know we were using autoheader now. I had removed many of the lines in config.h.in to try to clean it up a bit removing many of the macros that were no long being used. I probably didn't do a good job removing the tests from configure.in so autoheader probably put some of the macros back in. At some point we probably should see which tests are no longer needed by searching the source for usage of the defined macros in config.h and then remove the obsolete tests from configure.in.

CIao,
Tom

AndreaRigoni:
thanks Tom!

Andrea

PR #105, opened 2015-09-18
version tobytes on connection and tree

zack-vii:
version safer
-replace str.encode with _ver.tobytes
-move conversion to lower level function


tfredian:
Tested

PR #111, opened 2015-09-22
make PutMany and GetMany individual classes

zack-vii:
in order to avoid confusion between getMany and GetMany outsource
GetMany as a class of its own.
same for PutMany

!! check the **processGetMany** and **processPutMany** mechanism! I am not sure when it is used !!

### DIFF
--- a/mdsobjects/python/connection.py
+++ b/mdsobjects/python/connection.py
@@ -146,27 +146,6 @@ class Connection(object):
     def __del__(self):
         DisconnectFromMds(self.socket)
 
-    def __processGetMany__(cls):
-        try:
-            GetMany(value=Data.getTdiVar("__getManyIn__").deserialize()).execute().serialize().setTdiVar("__getManyOut__")
-        except Exception:
-            import sys
-            e=sys.exc_info()[1]
-            print(e)
-            raise
-    __processGetMany__=classmethod(__processGetMany__)
-
-    def __processPutMany__(cls):
-        try:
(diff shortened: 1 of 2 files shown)


tfredian:
I've done some testing and ran into a slight problem. In theory the GetMany.execute() should have worked as is but it turns out that the GetManyExecute function on the remote system is returning a Int8Array instead of a UInt8Array and the deserialize() call on this is falling through to the Data class method of deserialize which expects a argument. The original code worked because it did ans.deserialize(ans). One of the following fixes will make this work. 1) Fix the GetManyExecute function in the mdsobjects cpp shared library to return a Uin8Array. 2) Add code to the tdi fun to convert the array to Uint8 or 3) change the python code to do ans.deserialize(ans).

I believe those processGetMany and processPutMany methods could be removed. I think they predated the implementation of the GetMany and PutMany functions now in the cpp shared library.

I'll send an email to Gabriele to check on the data type being returned by the GetManyExecute function in the MdsObjectsCppShr library.

zack-vii:
That would be actually pretty smooth as java e.g. only know signed bytes. could 
save a lot of trouble. However only if the negative values a interpreted 
correctly via mod256. Should I update the pull request, accordingly? Or will you 
add the lines to Int8Array when the rest is checked for?

-timo

zack-vii:
I rebased it to ensure that it can be merged. However if you want to update you test branch you eventually need to use 'reset' instead of 'pull'.

tfredian:
Works! Thanks.

PR #130, opened 2015-10-06
add mdsException to package

zack-vii:
otherwise it will not be installed by
python setup.py install


tfredian:
Good find! Oops....

PR #135, opened 2015-10-08
Don't wait or close if kill of thread fails

joshStillerman:
If waiting, do wait WNOHANG


tfredian:
replaced by pull request 136

PR #137, opened 2015-10-09
allow python to load MDSplus libraries without DYLD_LIBRARY_PATH

dgarnier:
this is the first step in making mdsplus work in protected sessions on Mac OS X 10.11

### DIFF
--- a/mdsobjects/python/_mdsshr.py
+++ b/mdsobjects/python/_mdsshr.py
@@ -2,6 +2,7 @@ import ctypes as _C
 from ctypes.util import find_library as _find_library
 import numpy as _N
 import os as _os
+import platform as _platform
 
 if '__package__' not in globals() or __package__ is None or len(__package__)==0:
   def _mimport(name,level):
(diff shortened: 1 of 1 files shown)


dgarnier:
The other thing you need.. is to fix the linking.  The above change alone will still fail, but in a different spot. The solution is all there in the configure.ac file.. but for some reason, its been commented out... maybe whoever changed it didn't know why I had it.  Anyway, I don't have the whole build rig going so I can't make that change and test it trivially.

Now, you NEED to have -install_name set properly in the linker command.  Then once you link a file, it will know where to find it again.  We can chat about how you want to do that... how portable you want to make it.. (adding @rpath)...   

another way to fix it is with an install script that uses the install_name_tool.. but then you (may) also need to link with -headerpad_max_install_names

joshStillerman:
GREAT - Thanks for looking at this.  Maybe we can chat next week about LDX data, and this.  I would say burrito, but I am dietting

PR #138, opened 2015-10-13
MacOSX 10.11 support

dgarnier:
Combined two small changes that remove the dependence on LD_LIBRARY_PATH and / or DYLD_LIBRARY_PATH which are reset in protected processes going forward in MacOS X.  The python changes allow for the MDSplus libraries to be found under $MDSPLUS_DIR/lib  or /usr/local/mdsplus/lib.   

The second change, using @loader_path/../lib/ as the install name allows binaries and libraries in MDSplus to find libraries they've been linked to.  (This is just one way to solve the problem, but I like it since it remains portable... if you move MDSPLUS_DIR, it will still work.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -539,7 +539,7 @@ mips*) CFLAGS="$CFLAGS -signed -D_BSD_SIGNALS -KPIC";
        MOTIF_LD_LDARC="-multiply_defined suppress -L/usr/OpenMotif/lib"
        dnl LINKSHARED="$LDFLAGS -dynamiclib -install_name $libdir/\$(@F) -headerpad_max_install_names -prebind \
 	   dnl -seg_addr_table_filename \$(@F) -seg_addr_table ../macosx/bindtable -Wl,-single_module";
-	   LINKSHARED="$LDFLAGS -shared -arch i386 -arch x86_64";
+	   LINKSHARED="$LDFLAGS -shared -arch i386 -arch x86_64 -install_name @loader_path/../lib/\$(@F) -headerpad_max_install_names";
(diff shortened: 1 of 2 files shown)


joshStillerman:
Darren - 
  I had the same open motif problems.  I had the impression though, that we could just tell people they had to get brew or ports version of open motif, and then proceed. That we would not have to change our installation.

joshStillerman:
Also -
  I think we need to rip out the 32 bit stuff for mac

dgarnier:
The only problem with using "either ports or homebrew" is that they live in different places by default and the link is now fixed at compile time... (so the solution might be to have the installer search for one and then fix the installation using the "install_name_tool".)   It wouldn't be that tough a script to add actually and could be used for freetds as well.

I don't think you have to rip out the 32bit stuff.. you could fix it.. I have the technology...  (the issue isn't really are their 32 bit macs.. but is someone shipping 32bit apps that want to link.. so.. to support old versions of matlab, idl, python, etc.)  I know that some python distributions are only free for 32 bit installs.

joshStillerman:
I was wrong about the 32 bit stuff regardless.  I turns out I had not installed gfortran at all, and the configure script was not checking for it.  I am looking to see the best way to require the compiler be available.

We can talk more about motif ..

-Josh

dgarnier:
BTW.. I used this easy to install gfortran.

gfortran-5.2-bin.tar.gz http://prdownloads.sourceforge.net/hpc/gfortran-5.2-bin.tar.gz?download

It only exposes the gfortran binary in /usr/local/bin as a link and otherwise keeps itself in /usr/local/gfortran

PR #140, opened 2015-10-13
Fix memory overrun with quadword and octoword operations

tfredian:


tfredian:
This fix needs more work.

PR #150, opened 2015-10-14
rearraged import and _mimport

zack-vii:
-more straight forward way of importing modules.
-maybe not much faster but more consistent ; w/o level
-sys.exc_info()[1] <=> Exception as exc

python setup.py test -> OK  unix
python3 setup.py test -> OK  unix

### DIFF
--- a/mdsobjects/python/__init__.py
+++ b/mdsobjects/python/__init__.py
@@ -9,27 +9,22 @@ Information about the B{I{MDSplus Data System}} can be found at U{the MDSplus Ho
 @copyright: 2008
 @license: GNU GPL
 
-
-
 """
+import sys as _sys
 try:
+    if '__package__' not in globals() or __package__ is None or len(__package__)==0:
+        def _mimport(name):
+            return __import__(name,globals())
+    else:
+        def _mimport(name):
+            return __import__(name,globals(),{},[],1)
 
-  if '__package__' not in globals() or __package__ is None or len(__package__)==0:
-    def _mimport(name,level):
-      return __import__(name,globals())
-  else:
-    def _mimport(name,level):
-      return __import__(name,globals(),{},[],level)
-
-  _mimport('_loadglobals',1).load(globals())
-
-except:
-  import sys as _sys
-  print("Error importing MDSplus package: %s" % (_sys.exc_info()[1]))
-
+    _mimport('_loadglobals').load(globals())
+except Exception:
+    print('Error importing MDSplus package: %s' % (_sys.exc_info()[1],))
 
 def remove():
-  import sys,os
+  import os
   "Remove installed MDSplus package"
   _f=__file__.split(os.sep)
   while len(_f) > 1 and 'egg' not in _f[-1]:
@@ -39,5 +34,5 @@ def remove():
     try:
       import shutil
       shutil.rmtree(packagedir)
-    except:
-      print("Error removing %s: %s" % (packagedir,sys.error_sys()[1]))
+    except Exception:
+      print("Error removing %s: %s" % (packagedir,_sys.exc_info()[1]))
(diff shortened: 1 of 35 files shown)


tfredian:
Hi Timo, I think we've been down this road before... you can't use "except Exception as exc:". That syntax is not supported in python2.4 and we still support rhel5 which uses python2.4.

zack-vii:
yea my bad.. I will fix the exc part.

zack-vii:
I reverted the exc substitutions
tests on linux succeeded

tfredian:
I'm sorry that I didn't respond to your question about level ever being anything but 1 because the answer is yes. I think a better solution is to leave the level argument in but make it optional using:

def _mimport(name, level=1):

Unfortunately this is needed when trying to support python2.4, python2 > 4, and python3 all with the same code. Perhaps the definition of _mimport should use version_info tests instead of the kluge I came up with using "__package__".  I could work on this some more if you'd prefer or feel free to continue taking some of these suggestions into consideration.

zack-vii:
I see in tdibuildins are a few imports that require level 2. However since _mimport ist defined per module the changes would not interfere,  it is just nicer to have the identical code if it is the same function.
On the __package__  condition-monster: it seems **package** only exists since 2.6, and in consistently not None since 3.3.
For python 2.6 till 3.2 the value of __package is not set before the first regular 'import' inside the module. This is not always the case. Hence, apart from python >=3.3 the first function is defines and level has no meaning. I will check, when python requires additional information for the import. I have a feeling apart from the special cases where level!=1, the first function can be used. I will check it out

tfredian:
Hi Timo,

```
 I seem to recall needing to add the level stuff for version 3 
```

because the default level changed for the **import** builtin. In python2 
the level defaults to -1 which looks for the module name first as an 
absolute module name (i.e. found in sys.path) and then searches for the 
module relative to the module issuing the call. In python3 the default 
is 0 which means look just for an absolute module not relative modules. 
Specifying the level on python2 prevent erroneously picking up a stray 
module not in the same package. At least this is how I interpreted the 
help(**import**) info in both python2 and python3. I could be entirely 
wrong though.

-tom

zack-vii:
the level function has been added in version 2.5, considering dictating relative paths when possible a version check is a good way to decide which mimport to use.
The level seems to be relative to global(). since mimport is exported to version level is not required to address root modules. However when addressing siblings one needs to provide the personal globals() for it to work. I did some light testing executing "import MDSplus" on several versions of python. All versions from 2.4 to 3.4 imported MDSplus successfully. I will do a setup.py test tomorrow, once I am back in my office. Apart from the mdsException part, it looks clean now. It seem the Exceptions scripts are auto-generated. I could not find the generator, though.

...Nope does not work

tfredian:
This is much cleaner but still suffers from an import problem. You must have MDSplus installed on the system before: "from MDSplus import version as _ver" will work. I think you're going to just need to put the following in all the modules doing relative imports which is a whole lot cleaner than what I originally had:

from sys import version_info as pyver

if pyver<(2,5):
    def mimport(name, glbs=globals(), level=1):
        return **import**(name, glbs)
else:
    def mimport(name, glbs=globals(), level=1):
        return **import**(name, glbs, level=level)

tfredian:
The previous comment had underscores around the import which got turned into markdown styling.

zack-vii:
hmm...

```
ValueError: Attempted relative import in non-package
```

which is probably why you used that strange __package__ condition in the first place

I think trying the relative way first and falling back do default on error seems to be the best way to go.. Specially due to the strange way __package__ is filled with data in python 2.6 - 3.2

tfredian:
Nice work. Much cleaner.

PR #153, opened 2015-10-15
Use a weakref proxy for the Tree._activeTree property to enable prope…

tfredian:
…r cleanup of Tree instances


tfredian:
Hold off on this. Python3 problem.

PR #162, opened 2015-10-15
Add a python script that can be used to test for compatibility of the…

tfredian:
… python program and the MDSplus libraries used for testing


mdsobjects/python/tests/check_python_arch.py
```diff
@@ -0,0 +1,20 @@
...
+#
+# If ok the above command will return a status of 0.
+# If there is a architecture mismatch it will return 1.
+# Otherwise it will display an exception and return 1.
+#
+import ctypes,sys,os
+if sys.platform.startswith('win'):
+    lib='MdsShr.dll'
+else:
+    lib='libMdsShr.so'
```
dgarnier, line 16: This will fail on darwin.  (Which use .dylib as an ending).

PR #165, opened 2015-10-16
A python

zack-vii:


tfredian:
Fixes typo on use of mdsExceptions and removes an unused import of weakref.

Issue #169, opened 2015-10-20
Twf use weakref for active tree #155 raises errors when tree ref  is gone


zack-vii:
Since TreeNode.tree is a proxy instead of a ref checking its validity using

```
TreeNode.tree is None
```

fails and the following use of tree in the Context of a Tree type will reise an error:

> ReferenceError: weakly-referenced object no longer exists

In order to change tree to a weakref you need to either

use weakref.ref and replace each "read" appearance of TreeNode.tree with TreeNode.tree()

or change the condition "TreeNode.tree is None" to something that check if the proxy ist still valid

tfredian:
I went through all this when changing to weakref and thought I fixed all 
the problems that produced this error. Can you tell me under which 
circumstances you still get this error? I tested this with python27 and 
python34 and it just worked.

Thanks,
Tom

zack-vii:
The error occurs is a treenode is passed down by a user function. All references within the function are closed but the treenode keeps its .tree attribute which is called when decompiling for **repr**:
full error:

```
  File "E:\PortableProgs\Python\python-2.7.10\Lib\site-packages\MDSplus\mdsdata.py", line 507, in decompile
    return _tdishr.TdiDecompile(self)
  File "E:\PortableProgs\Python\python-2.7.10\Lib\site-packages\MDSplus\_tdishr.py", line 54, in TdiDecompile
    return _ver.tostr(_TdiShrFun(TdiShr.TdiDecompile,"Error decompiling",expression))
  File "E:\PortableProgs\Python\python-2.7.10\Lib\site-packages\MDSplus\_tdishr.py", line 29, in _TdiShrFun
    arguments = [_C.pointer(descriptor(expression))]+parseArguments(args)+[_C.pointer(xd),_C.c_void_p(-1)]
  File "E:\PortableProgs\Python\python-2.7.10\Lib\site-packages\MDSplus\_descriptor.py", line 214, in __init__
    value.restoreContext()
  File "E:\PortableProgs\Python\python-2.7.10\Lib\site-packages\MDSplus\treenode.py", line 1000, in restoreContext
    self.tree.restoreContext()
ReferenceError: weakly-referenced object no longer exists
```

I made a pull request for a suggestion. I am not sure if it conserves the desired behaviour introduced by weakref

zack-vii:
[...]

PR #171, opened 2015-10-20
changed _activetree to property

zack-vii:
this way you can use weakref ref without changing the behaviour of
_activetree

might be a more elegant way of using weakref (#155, #169)
but need testing if the desired behaviour is conserved


tfredian:
The handling of closing trees using the **del** method of Tree objects was handled using https://github.com/MDSplus/mdsplus/pull/173

PR #173, opened 2015-10-21
Different approach to tree deletion handling. Using weakref failed to…

tfredian:
… work for all cases.

### DIFF
--- a/mdsdcl/cmdExecute.c
+++ b/mdsdcl/cmdExecute.c
@@ -12,6 +12,7 @@
 #include <mdsshr.h>
 #include <dcl.h>
 #include <mdsdcl_messages.h>
+#include <pthread.h>
 #include "dcl_p.h"
 
 static dclDocListPtr dclDocs = 0;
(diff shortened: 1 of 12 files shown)


tfredian:
Instead of using the weakref module which didn't work if new tree instances were created from the "active tree" I decided to keep a list of the tree contexts (pointers to memory). The "active tree" was also saved as the tree context. When new trees are created they add the context to the list of open trees and set the active tree ctx value. When trees are destroyed they remove the first occurrence of the contexts from the list. If no more instances of that context are on the list then the tree is closed and the context freed.

tfredian:
Needs a bit more work.

tfredian:
At lot of rework and testing using threaded tests by doing python setup.py test with TEST_THREADS environment variable defined. Found and fixed reentrancy issues with the mdsdcl bison parse even though it is supposedly configured to be reentrant. Found nci locking issues in treeshr as well. Also discovered that the python garbage collector can execute the tree.__del__ method at any time and will pause any other execution at that time. If the __del__ method attempts to lock and another active piece of code has that lock the whole thing deadlocks since the other piece of code will not continue executing while the garbage collector is trying to execute the __del__ method. This was fixed by spinning up a thread in the __del__ method to finish the cleanup which avoids the locking issue with the garbage collector. I ran over 200 tests consisting 10 threads doing tree editing, tree writes, reads etc in parallel with 10 other threads doing many tdi operations all in parallel and had no deadlocks or failed tests. It probably would be a good idea to test these changes with python device implementations to ensure that there are no change in behavior after these changes.

PR #176, opened 2015-10-23
tree open Dialog improvement

zack-vii:
- combined common comands of if and else case
- removed if(curr_origin == null) which prevented Dialog to open occasionally
- turned open mode into radio buttons (no edit+readonly)
- edit -> edit/new hint on edit is new if shot does not exist


GabrieleManduchi:
You can drop the open realtime option as it is obsolete and no more used

PR #177, opened 2015-10-27
Simplify new frame detection, upload automatically

bstandley:
zach-vii and I made the frame detection loop simpler.  The only practical effect should be that post-pulse upload begins immediately after the last frame when using FRAME_SYNC = "FIXED FRAME RATE" and DURATION > 0.  The code has been tested but not yet with external timing.


zack-vii:
I tried to modify and contribute the changes but I did not know how.
you should consider deleting line 5

> import datetime  # it is  unused

line 72

> currTime = 0  # I suppose it is correctly initialised in line 71 with 'c_long(0)' 0 might not work with byref

and line 118

> currTime = 0.0  # it is also unused

PR #180, opened 2015-10-28
change the rename dialog ...

zack-vii:
... as it is buggy when changing the path
- node.rename is more like move as the full path can be changed
- node.renameLast is the proper rename method in node
- the error dialogs are safer on lower levels
- when requesting the node name the trimmed version is either requested
  or should fit as well
- removed "node" from title as it is too long anyway


zack-vii:
- fixed and implemented copy and move of node structures
- Ctrl+X cuts, Ctrl+C copies, and Ctrl+V pastes
  so far only tested on windows
- now the basis for drag and drop manipulation is set
- fixed the issue that no node is added if parent had no child

PR #183, opened 2015-10-29
fixed with structure exit

zack-vii:
- quit() in any case if edit. Write does not close edit session. If you want to write you should do it intentionally (i.e. manually). Just make sure the tree is released.


tfredian:
This is fine. I think I tested the "with" clause and was pretty sure the python garbage collected closed and release the tree so an exit method is probably not needed but explicitly calling the **del** should be fine.

PR #191, opened 2015-11-03
clean compress wont work if lock is set

zack-vii:
removed lock as it will prevent clean and compress from finishing

### DIFF
--- a/mdsobjects/python/tree.py
+++ b/mdsobjects/python/tree.py
@@ -645,12 +645,7 @@ class Tree(object):
         """Clean up data file.
         @rtype: None
         """
-        Tree.lock()
-        try:
-            #ctx = _treeshr.TreeGetContext()
-            status = _treeshr.TreeCleanDatafile(self.ctx, _ver.tobytes(self.tree), self.shot)
-        finally:
-            Tree.unlock()
+        status = _treeshr.TreeCleanDatafile(self.ctx, _ver.tobytes(self.tree), self.shot)
         if not (status & 1):
             raise _Exceptions.statusToException(status)
 
@@ -658,11 +653,6 @@ class Tree(object):
         """Compress data file.
         @rtype: None
         """
-        Tree.lock()
-        try:
-            #ctx = _treeshr.TreeGetContext()
-            status = _treeshr.TreeCompressDatafile(self.ctx, _ver.tobytes(self.tree), self.shot)
-        finally:
-            Tree.unlock()
+        status = _treeshr.TreeCompressDatafile(self.ctx, _ver.tobytes(self.tree), self.shot)
         if not (status & 1):
             raise _Exceptions.statusToException(status)


tfredian:
I'm not sure why this would be needed. The lock is only a python thread lock and will only prevent another python thread from doing a tree operation at the same time. It should not block the actual clean or compress operation. I just tried the current alpha and it seemed to compress and clean just fine with the locks in place. Can you do some more testing? Perhaps you ran into some other race condition.

zack-vii:
Hmm you are right.. it works now. Its was not working when I run my testtree script to rebuild the tree. 

> https://github.com/zack-vii/archive/blob/master/tdi/testtree.py

Now it completed without errors on python2.7 32bit, however, not always I am using the with structure to create the tree, then put data to the node in normal mode and finally the script tries to compress I cannot delete the old file:

```
File "MDSplus\tree.py", line 658, in compressDatafile
    raise _Exceptions.statusToException(status)
MDSplus.mdsExceptions.treeshrExceptions.TreeDELFAIL: %TREE-E-DELFAIL, Error deleting file.
```

It just work (once) when I removed the locks but now even with locks removed I receive the error.
Might have just been chance and is unrelated to the locks.

I tried with and without locks (both with the same results)

```
1a) open/new - addNodes - write - quit - open/normal - putData - compress   => errror
2a) open/new - addNodes - write - quit - open/normal - putData - reopen/normal - compress => error
3a) open/new - addNodes - write - putData - quit - compress => success
1b) with/new - addNodes - write - endwith - with/normal - putData - compress - endwith  => error
2b) with/new - addNodes - write - endwith - with/normal - putData - endwith - with/normal - compress - endwith  => error
3b) with/new - addNodes - write - putData - endwith - with - compress => success
```

I would have preferred version 1b, as it remains in edit mode for the shortest period.

zack-vii:
I also tested it on python3.3 64bit and python2.7 64bit, both times receiving

```
  File "MDSplus\treenode.py", line 919, in putData
    _treeshr.TreePutRecord(self,data)
  File "MDSplus\_treeshr.py", line 160, in TreePutRecord
    status=__TreePutRecord(n.tree.ctx,n.nid,_C.pointer(descriptor(value)),0)
WindowsError: exception: access violation reading 0xFFFFFFFE61582440
```

there still seems to be something wrong with the Float32 data type on 64 bit systems.

zack-vii:
hm maybe its not the Float32 alone. I tried to store a signal 2D array per timestamp (i.e. 64x64x64). I tired to store the very signal I read from the TestTree I previously build using the 32bit version. and got the same error again. its testtree node DATA:ARR2DF containing the following sample data:

> Build_Param(Build_Signal(Build_With_Units(FLOAT($VALUE), "V"), Build_With_Units(Set_Range(64,64,64,4. /**\* etc. ***/), "data"), Build_With_Units(0D0 : 1D0 : .01587301587301587D0, "time"), Build_With_Units(0D0 : 1D0 : .01587301587301587D0, "dim_of(1)"), Build_With_Units(0D0 : 1D0 : .01587301587301587D0, "dim_of(2)")), "this is the help text", *) 

I now changes the dimensions to type Dimension with no result. It turned out if I replace the raw data with numpy.ones(shape) before line 91 (before Float32Array(raw).setUnits('data')) no error occurs.
So I thought it has something to do with shuffled space in memory cause by my construction loops. However, an raw.copy('C') or raw.copy('F') did not help, either. I understood this function as a reallocation to get a contiguous slap in memory.

zack-vii:
I narrowed the error down to this data to write:

``` python
import MDSplus, numpy
tree = MDSplus.Tree('test',3)
node = tree.getNode('gub')
node.record=(numpy.array(range(33),'float64')*2/63-1).astype('float32')
```

So I calculated the vector with double precision and typecast the result to single precision.
If I omit the first or last element the storing works. But the whole 33 elements vector causes a problem.
Now it becomes interesting: The error occurs only if CompressOnPut is enabled. 
If it is disabled, and one would later invoke a compressDatafile from any 64bit application (e.g. mdstcl), the application crashes probably with the same error.
This seems to be a rare occasion. However, the error occurs before the datafile is released and apparently is not caught properly. In result the datafile will remain locked for any other process until the one causing the problem is closed. So if it happens, its a biggy. So one could either check the compressing routine, where it actually fails and fix it or work around the problem by catching it and write without compressing if compressing fails.

PR #198, opened 2015-11-06
Fix obscure bug when delta value = 0x8000000 during compression

tfredian:


zack-vii:
I merged the commit and compiled it with docker for win64 and debian64. I tested running the testtree.py. They all finished without error and wrote the correct data. That was a strange issue. Well how a the odds to hit 0x80000000 on the spot

zack-vii:
Thanks for the fix ;)

PR #199, opened 2015-11-06
Andrea external build

AndreaRigoni:
Some improvements about tests of this week:
- Build mdsplus off the source directory ( just cd into a build_dir and do path_to_source/configure )
- Makes it possible to build and run test using wine in mingw
- Do selection of some env for wine in testing.m4 macro (some more to be done...)


tfredian:
Wow! Anxious to try this out! Next we'll need tests for the tests since it would take weeks to examine all these changes!

AndreaRigoni:
hehehe.. I'm working on that too.. they are in testing/selftest
:-)

PR #200, opened 2015-11-06
jScope eventupdate block

zack-vii:
Prevents events from dispatching when busy.
It caused jScope to freeze if an event occurred during loading of configuration.


tfredian:

tfredian:

PR #204, opened 2015-11-09
fixed status bar on resizing

zack-vii:
- status bar remains a one-liner and the tree pane resizes


zack-vii:
As I was learning about the Layouts I optimized the resizing behaviour of the Editors.

GabrieleManduchi:
You will find many features which can be improved...jTraverser has been written in 2001 and never changed since then...go ahead!

PR #206, opened 2015-11-10
make MARTE Device family python3 ready

zack-vii:
- fix print
- for loop -> b=list(a) is like b=a.copy()
- less verbose
- import only what you need
  => faster more robust


GabrieleManduchi:
Thanks!

PR #210, opened 2015-11-11
all-uppercase attributes will be redirected to getNode

zack-vii:
all-uppercase attributes will be redirected to getNode
- allows a shortcut of node.getNode('DESC') by using node.DESC
- much faster way of getting the desired node


tfredian:
Great idea Timo!

quite powerful:

>>> Tree('main',-1).SUBTREE.A14.record
Build_Conglom("MIT$DEVICES", "A14", *, *)

way cool

PR #214, opened 2015-11-16
Use int64_t instead of long to always use 64-bits. Also fix some comp…

tfredian:
…ile warnings/errors


GabrieleManduchi:
Hi Tom,
   my fault, I should have used int64_t since the beginning....

PR #223, opened 2015-11-19
Fix indentation problem which prevents loading module in python3

tfredian:


tfredian:
Jenkins tested:FAILURE

PR #224, opened 2015-11-19
For some reason the use of python based funs causes a problem with th…

tfredian:
…e ctypes POINTER function used in creating descriptors. This fix uses the ctypes cast function to cast a POINTER(descriptor) to the type of an instance of a pointer to a descriptor.


tfredian:
Jenkins tested: SUCCESS

Issue #222, opened 2015-11-19
python TdiExecute a python script causes stange error when decompiling


zack-vii:
If you call a python tdi program from within python, Python becomes incapable of decompiling Signals.
The TDI script pytest.py:

``` python
import MDSplus
def pytest ( arg ):
    return str(type(arg))
```

The python input:

``` python
import MDSplus
sig=MDSplus.Signal([1,2,3],None,[1,2,3])
sig
MDSplus.TdiExecute("pytest(1)")
sig
```

The python output:

``` python
Python 2.7.3 (default, Mar 13 2014, 11:03:55)
[GCC 4.7.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import MDSplus 
>>> sig=MDSplus.Signal([1,2,3],None,[1,2,3])
>>> sig
Build_Signal([1,2,3], *, [1,2,3])
>>> MDSplus.TdiExecute("pytest(1)")
"<class 'MDSplus.mdsscalar.Int32'>"
>>> sig
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/dist-packages/MDSplus-alpha_7.0.182-py2.7.egg/MDSplus/mdsdata.py", line 512, in decompile
    return _tdishr.TdiDecompile(self)
  File "/usr/local/lib/python2.7/dist-packages/MDSplus-alpha_7.0.182-py2.7.egg/MDSplus/_tdishr.py", line 47, in TdiDecompile
    return _ver.tostr(_TdiShrFun(TdiShr.TdiDecompile,"Error decompiling",expression))
  File "/usr/local/lib/python2.7/dist-packages/MDSplus-alpha_7.0.182-py2.7.egg/MDSplus/_tdishr.py", line 30, in _TdiShrFun
    arguments = [_C.pointer(descriptor(expression))]+parseArguments(args)+[_C.pointer(xd),_C.c_void_p(-1)]
  File "/usr/local/lib/python2.7/dist-packages/MDSplus-alpha_7.0.182-py2.7.egg/MDSplus/_descriptor.py", line 196, in __init__
    c_d.dscptrs[i]=_C.pointer(descriptor(value.args[i]))
  File "/usr/local/lib/python2.7/dist-packages/MDSplus-alpha_7.0.182-py2.7.egg/MDSplus/_descriptor.py", line 246, in __init__
    apd=descriptor(value.toApd())
[...]

PR #226, opened 2015-11-20
make thread_data a static from Tree and detach private

zack-vii:
raised a "private is not an attribute of Thread._local" error with devices.

I thought, it might be related to issue #222 but maybe not entirely.

### DIFF
--- a/mdsobjects/python/tree.py
+++ b/mdsobjects/python/tree.py
@@ -18,10 +18,6 @@ _scalar=_mimport('mdsscalar')
 _treenode=_mimport('treenode')
 _ver=_mimport('version')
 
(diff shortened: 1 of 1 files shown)


tfredian:

tfredian:
Timo, I submitted another pull request (https://github.com/MDSplus/mdsplus/pull/228) that should fix this problem a little differently. I left the thread_data as a global but changed it to _thread_data so it would not be exported. I added global to the functions that reference it in the module. Could you let me know if this works with your device problems? If not I will test and merge this request instead.

zack-vii:
whatever works best. (use #228 with #229)

PR #227, opened 2015-11-20
No need to create new interpretter which causes problems with ctypes.…

tfredian:
… Also fixed handling of strings with python3 which uses unicode types instead of str types.


tfredian:
Fix for https://github.com/MDSplus/mdsplus/issues/222
I tested this on fc22 and Windows (64-bit) and python based funs work with both python2 and python3 and no strange side effects between fun execution and the python main application.

tfredian:
Jenkins tested: SUCCESS

PR #228, opened 2015-11-20
Make the thread_data symbol global but not exported in the tree module

tfredian:


tfredian:
Jenkins tested: SUCCESS

PR #229, opened 2015-11-21
Bstandley fix thread handling in tree py

bstandley:
I spent some time looking at this and believe I have a good solution.  To start off, I wrote some code (available at http://iceberg.brianstandley.com/derper.py) which demonstrates the problem.  I then found some documentation (https://pymotw.com/2/threading/#thread-specific-data) which indicates that adding attributes to an instance of thread.local _must_ be done by subclassing.   Evidently each additional thread gets a "fresh" copy of _thread_local which does not include any attributes added later in the main thread.  The official module docs completely fail to mention this subtlety . . .

I also removed the global statements, which are not necessary when interacting with only attributes of the global-level object.

Timo's pull request also makes the test code run, but involves more significant modification to the original code, and I am not familiar enough the problem space to say whether it is functionally equivalent, so I suggest merging mine instead.  Sorry Timo ;-)

### DIFF
--- a/mdsobjects/python/tree.py
+++ b/mdsobjects/python/tree.py
@@ -18,9 +18,9 @@ _scalar=_mimport('mdsscalar')
 _treenode=_mimport('treenode')
 _ver=_mimport('version')
 
-thread_data=_threading.local()
-thread_data._activeTree=0
-thread_data.private=False
+_thread_data=_threading.local()
+_thread_data._activeTree=0
+_thread_data.private=False
 
 _hard_lock=_threading.Lock()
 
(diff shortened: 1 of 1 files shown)


tfredian:
Jenkins tested: SUCCESS

zack-vii:
Whatever works best =)  (#228)

bstandley:
I just tested this with the CYGNET4K device and it worked.

tfredian:
Looks good to me. I just tested this using the thread private tree contexts by doing:

$ export TEST_THREADS=yes
$ cd mdsplus/mdsobjects/python
$ python setup.py test

and that worked as well.

PR #230, opened 2015-11-23
1) Only do locking of python GIL when doing setup methods (needed for …

tfredian:
…glade based setups)
2) Use MdsMisc->PyCall on windows too.
3)  Prevent closing dbid tree context when doing python doMethod.
4)  Fix py function on windows
5)  Fix tdi py functions for python3.


tfredian:
Jenkins tested: SUCCESS

PR #234, opened 2015-11-26
python added alternative for data like get for dict

zack-vii:


tfredian:
I'm not sure I understand the need for this enhancement. What is its purpose? To provide some kind of "default value"? Also is there some advantage of implementing it this way instead of using something like ,altvalue=None and checking whether altvalue is not None?

zack-vii:
Well I use it fro switches that are set by Nodes. If the Node is not set it should default to 0 (i.e. false). if you don't want to use a default you can omit the parameter and it will create a NoDataError as before. However, if you want to use a default, you do not need to create the try-except structure to catch the NoDataError. It is just for convenience. It acts in a similar way as the get method of the dict type.

zack-vii:
I needed to rebase though

zack-vii:
Oh about that second question: the way of using *altvalue instead of atlvalue=None is that the user might have wanted to have the default to be None. but this would generate the error. using *altvalue you can allow any default value but still keep the case where the function should throw the error

PR #236, opened 2015-11-26
temporal fix for the alpha branch #230

zack-vii:
since it is more important that devices run than Windows python to
execute python via TdiExecute:
- checkout Py.fun form prior #230
- added optional input to Py.fun for compatibility

### DIFF
--- a/tdi/Py.fun
+++ b/tdi/Py.fun
@@ -1,5 +1,23 @@
 public fun Py(in _cmd, optional in _varname, optional in _global_namespace, optional in _lock) {
-   _locked=present(_lock) ? _lock : 0;
+   fun PyCall(in _routine,_locked, optional in _arg) {
+      if (_locked) {
+        _GIL=build_call(51,getenv("PyLib"),"PyGILState_Ensure"); 
+      }
+      if (present(_arg)) {
+        _ans=build_call(8,getenv("PyLib"),_routine,_arg);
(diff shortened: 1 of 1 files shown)


zack-vii:
AS much as I want fix #230 to work. I think its best to remove it from alpha until it is fixed

tfredian:
I don't understand this "fix". Does it fix your device thread problems? As far as I can tell this now locks every py() call. Much of this patch should not be needed as it adds back in the Windows code but the MdsMisc libraries PyCall routine now works on Windows so both linux and Windows can use MdsMisc->PyCall(). If this does fix your device problems perhaps the recent modifications of pycall.c fixes the locking problem so that the GIL locking can always be done. I will attempt to decipher what exactly this change does and try to come up with another fix that does not require a separate python function for Windows and have you test that. I'm also confused now as to what does work and what does not after this change. I'll do some more exploring.

zack-vii:
basically this patch reverses all modifications to Py.fun done in #230, but since the other changes have been kept intact I needed to add a third optional input that has been introduced. with only this file reverted the device problems diapered and the windows problem returned. I am not sure how much of this was necessary. I just checked out the version prior to #230. I then checked out the changed files one by one until it broke again and Py.fun was the one.

zack-vii:
not needed since gil-locking is close to shipping

PR #242, opened 2015-11-30
setIncludedInPulse vs include_in_pulse

zack-vii:
should it not be setIncludeInPulse
but might be too late now

### DIFF
--- a/mdsobjects/python/treenode.py
+++ b/mdsobjects/python/treenode.py
@@ -1053,7 +1053,7 @@ class TreeNode(_data.Data):
         """
         return self.__setNode('essential',flag)
 
-    def setIncludedInPulse(self,flag):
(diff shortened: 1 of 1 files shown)


zack-vii:
I have to apologize.. this commit is not complete.
1. I forgot to fix the **setattr** for include_in_pulse to change accordingly
2. A search for further IncludedInPulse revealed that it is used in many placed libraries LabVIEW etc. 
I could sort out all appearances in alpha and change them as well. not sure If I can change the labview one. LabVIEW has a problem of its  own as the vi files should probably provided for a broader range of versions. This could be in a separate packages, one for each range of versions. one could use the compatibility table (http://www.ni.com/tutorial/8387/en/) to find reasonable versions to provide with the broadst range of compatible versions.

PR #243, opened 2015-11-30
systematically replaced IncludedInPulse with IncludeInPulse

zack-vii:
Not sure it this should be done, but at least the treenode.py should be considered. However, in this case also test1.py and treeUnitTest.py must
be changed (c.f. #242 and 5ab984a)
- I does compile and install without complains


joshStillerman:
The C++ consumers / developers should comment...

It seems to me at a minimum
IncludedInPulse should be a synonym for IncludeInPulse

otherwise any program using this API will be broken by this change.

-Josh

joshStillerman:
Great -
  I will leave it to a C++ person to merge this.

PR #246, opened 2015-12-01
Twf fix gil locking

tfredian:
Hopefully solves the GIL locking issues with executing python from within MDSplus. Adding a call to PyEval_SaveThread during first initialization of the python interface seems to enable the ability to lock and release the GIL for each python execution so no longer need to tweak the locking for various applications. The tutorial 2 demo now works without hangs even though the GIL is now always being locked and released.

The PyDoMethod.fun which is executed by TreeDoMethod.c has also been replaced with a PyDoMethod.py where it is much simpler to implement the importing of modules and calling the device methods. The python devices can now raise exceptions from mdsExceptions submodule if desired and the PyDoMethod function will return the status value of that exception which will be displayed on the action monitor instead of a generic failure status of 0.


joshStillerman:
Sounds great.

I rushed to the airport for an earlier flight, and it is delayed too … grrrr

zack-vii:
Nice I tested it on both, linux64bit running the cygnet4k driver and on windows with the python test script (#222). both worked flawlessly. good job.
before you commit you should consider to remove the now obsolete _lock in Py.fun and callers. I can prepare a device update for RfxDevices where I remove the line wher it set the _PyReleaseThreadLock

zack-vii:
Brian is reporting issues with his trigger box drivers and a digitizer. We are investigating the cause.

zack-vii:
for one there is a typo in PyDoMethod hassatr should be hasattr in line 26

tfredian:
Ok. I'll fix that and clean up the code to remove the obsolete lock 
options. Let me know if you discover the problem that Brian was running 
into. Hopefully it was this typo. This typo would have only have 
affected the handling of errors returned by the device method.

zack-vii:
I adjusted PyDoMethod.py to be more compatible, further, another mistake is on the method call args instead of *args. that solved a lot of problems.

zack-vii:
Hey Tom, I got to the bottom of it I think. object.**dict** does not include methods of the superclass. use **getattribute** instead with a try except structure. I have a working version now.
last thing is that device method used to require at least on argument we should remove this extra argument as it is no longer provided automatically. I will fill in a try except structure that provides a dummy one (1) if needed eg by customized devices. but I recommend to removed it from the ones provided with MDSplus.

tfredian:
yep. I just ran into this while testing the device setup here. I have a 
fix for that. I'll try to merge in your changes to PyDoMethod.py with 
mine and push an update.

thx!
-tom

zack-vii:
use this PyDoMethod.py

``` python
from MDSplus import StringArray, Int32
from MDSplus.mdsExceptions import TreeNOMETHOD
from sys import stderr

def PyDoMethod(n,method,*args):
    top = n.conglomerate_nids[0]
    c = top.record
    q = c.qualifiers
    model = str(c.model)
    method = str(method)
    if isinstance(q,StringArray):
        for i in range(len(q)):
            exec(str(q[i])) in globals()
    else:
        exec(str(q)) in globals()
    if not model in globals():
        return -1  # Device not found
    device = globals()[model](n)
    try:
        try:
            methodobj = device.__getattribute__(method)
        except AttributeError:
            return TreeNOMETHOD.status
        try:
            status = methodobj(*args)
        except TypeError:
            print('Your device method %s.%s requires at least one argument.' % (model,method))
            print('No argumnet has been provided as it is probably not required by the method.')
            print('MDSplus does not require device methods to accept an argument anymore.')
            status = methodobj(None)
        Int32(1).setTdiVar("_method_status")
    except Exception as exc:
        stderr.write("Python error in %s.%s:\n%s\n" % (model,method,repr(exc)))
        if hasattr(exc,'status'):
            status = exc.status
        else:
            status = 0
        Int32(status).setTdiVar("_method_status")
    return status
```

if model not found maybe a better error code is available

zack-vii:
Ah forgot about it again.. I think I messed up commit 8f27282. I was not able to write before and now suddenly I was. I guess I reverted it somehow.
edit: never mind, fixed it

zack-vii:
One issue remains: the Int32.setTdiVar("_method_status") how does it know that the return value is an error code or just a proper return value. e.g. the MitDevices ACQ return the original name of the nodes. i.e. a string which cannot be casted to Int32. it is strange enough that sub functions of device methods call PyDoMethod again instead of just running the python method.

tfredian:
Good question! Actually the _method_status is not used now. I'll have to 
work some on this. The TreeDoMethod.c code needs to deal with this issue 
and it is not currently doing this.

zack-vii:
The recall of PyDoMethod is triggered by TreeNode.getOriginalPartName which is invocing a getnci of 'Original_Part_Name'. This is not required as we are in python already and could call the Device.ORIGINAL_PART_NAME instead. also doMethod in TreeNode is probably not required anymore.

tfredian:
I think the original part name method being called by getnci is now 
working with python devices. Please test the latest version of this pull 
request branch and let me know if you find any issues. It seems like the 
locking issues are gone now. I've done some testing with the demo 
devices and they seem to work now without any lock conflicts. I've 
tested the glade based python device setup with some of the MitDevices 
and that seems to work again without lock issues.

The new implementation should have some benefits over the original 
python device support. If a device raises one of the standard 
MDSplusExceptions the return status of TreeDoMethod will be a standard 
status return so doing DO/METHOD in tcl will generate a useful error 
message as will the status messages in the action monitor. In theory the 
python device methods should be able to support multiple arguments (i.e. 
do/method device method arg1 arg2 arg3).

Python device methods should raise exceptions for failure modes. The 
return value will be treated as the "answer" of the method not the 
status. I believe this was always the case but I'm now not certain of 
this. If python device methods do not expect an argument they should not 
include an "arg" argument in the method definition. The current 
implementation will print out a message about this and attempt to 
execute the method with a dummy argument. If you want to have a device 
method with one or more optional arguments you could specify the method 
as: def mymethod(self,*args) and then examine the args as a tupple. This 
will not produce a message if the method is executed with no arguments.

Anyway, this turned into a much more complicated problem than I 
originally thought. Thanks Brian and Timo for doing a lot of testing and 
tracking down problems in the code. Hopefully we're starting to see some 
light at the end of a very long tunnel!

-tom

zack-vii:
It seems to work on the triggerbox and the acqs of MitDesvices. I could not test the cygnet4k yet since they where already turned off. but I will do so tomorrow. I assume the device methods need to throw an error now instead of returning the error.status. which is fine. We just need to modify the devices .py files once this one is merged.

zack-vii:
Good news. We tested the cygnet4k cameras (they use threading) and all 20 cameras run as expected. init, start_store, stop_store. So from what we can test here, this update really fixed all the bugs/issues related. As far as I am concerned It is ready to be merged. Updates build on it (error handling and clean up of the python devices) are on a different page. I could start going through the RfxDevices, which I am already familiar with. We are also interested in the ACQs of MitDevices. So those would be next on my list.

zack-vii:
If the device throws an error I would have assumed the error status/the related error would be thrown in e.g. in mdstcl.

```
Error doing init on node \TEST::TOP:C
  File "E:\PortableProgs\Python\python-2.7.10.amd64\Lib\site-packages\MDSplus\tree.py", line 360, in doMethod
    exec( str('status=_data.makeData('+model+'(n).'+method+'(_data.Data.getTdiVar("__do_method_arg__")))'))
  File "<string>", line 1, in <module>
  File "E:\PortableProgs\Python\python-2.7.10.amd64\Lib\site-packages\RfxDevices\CYGNET4K.py", line 137, in init
    tmpPath = self.genconf()
  File "E:\PortableProgs\Python\python-2.7.10.amd64\Lib\site-packages\RfxDevices\CYGNET4K.py", line 186, in genconf
    confPath = self.conf_file.data()
  File "E:\PortableProgs\Python\python-2.7.10.amd64\Lib\site-packages\MDSplus\mdsdata.py", line 528, in data
    raise _Exceptions.treeshrExceptions.TreeNODATA
TreeNODATA: %MDSPLUS-?-UNKNOWN, Unknown exception
Error executing device method
Error message was: %NONAME-E-NOMSG, Message number 0x00000002
mdsdcl: --> failed on line 'do/method C init'
```

instead of
NONAME-E-NOMSG, Message number 0x00000002
I expected the NoData error message

zack-vii:
seems like mdstcl at least is still using the return value to interpret the error. If no return value is given as the function terminated successful, it returns error 0x00000000

tfredian:
Ok, hopefully things are getting close on this pull request. Below is some examples of things working as I hoped they would. I am invoking the init method of Gabriele's demoadc module. The first time it ran successfully (no output or errors). Then I cleared a node in the device which the init tries to read. This produces an error and the error status is that of the exception raised in the python. Then I do a dir/full on a node in the module which invokes the ORIGINAL_PART_NAME method and that worked and the result of that method is the string. I had changed the demoadc python to 1) not return a value and 2) raise exceptions instead of returning a 0 status. Again, once we merge in these changes we should "fix" the python devices by:
1) Remove unused method args
2) Change clauses like:

try:
    xxxxx
except:
    print("something was done wrong here")
    return 0

To:

try:
     xxxxxxx
except:
     print("Something was done wrong here")
     raise
3) Optionally remove all return statements in the method if all errors are raised as exceptions.

TCL session with these changes in place:

TCL> set tree tutorial/shot=2
TCL> do/method .devices:adc_1 init
TCL> put .devices:adc_1:name ""
TCL> do/method .devices:adc_1 init
Missing Name in device
Python error in DEMOADC.init:
%TREE-E-NODATA, No data available for this node

Error executing device method
Error message was: %TREE-E-NODATA, No data available for this node
mdsdcl: --> failed on line 'do/method .devices:adc_1 init'
TCL> dir/full .devices:adc_1:name

\TUTORIAL::TOP.DEVICES:ADC_1

 :NAME  
      Status: on,parent is on, usage text
      compress on put
      Data inserted:  3-DEC-2015 13:27:52.02    Owner: gid=1000(twf),uid=1000(twf)
      There is no data stored for this node
      Model element: 2
      Original element name: \TUTORIAL::TOP.DEVICES:ADC_1:NAME

Total of 1 node.
TCL>

zack-vii:
I tested the new additions extensively, and it seem that the devices work, I will talk to Brian tomorrow about the details. However, in python I now get an error when toggling the state of an node with e.g. setOn.
```
  File "<stdin>", line 1, in <module>
  File "...\python-2.7.10.amd64\Lib\site-packages\MDSplus\treenode.py", line 1096, in setOn
    _treeshr.TreeTurnOn(self)
  File "...\python-2.7.10.amd64\Lib\site-packages\MDSplus\_treeshr.py", line 220, in TreeTurnOn
    raise _Exceptions.statusToException(status)
MDSplus.mdsExceptions.treeshrExceptions.TreeLOCK_FAILURE: %TREE-E-LOCK_FAILURE, Error locking file, perhaps NFSLOCKING not enabled on this system
```
is this NFSlocking related to GILlocking?

zack-vii:
I tested the new additions extensively, and it seem that the devices work.  Brian's TriggerBoxes still return some 0x0000000 error messages but this might be related to some settings. I will talk to Brian about the details tomorrow. It might be something trivial. The python error has been resolved as well. My libraries where not yet up to date, which caused an error in python itself. After doing a clean install all went back to normal.

tfredian:
Hmmm. This should not be related to the GIL lock. Is this repeatable? Do 
you have write access to the tree files?

PR #249, opened 2015-12-04
added getClass to Conglom

zack-vii:
- finds the python class of a device conglomerate
- helps to test devices from inside python using >>> dev = node.record.getClass(node)


zack-vii:
it only adds a feature that is very useful when developing devices in python

zack-vii:
it only add a feature that comes in pretty handy when developing devices in python

PR #259, opened 2015-12-16
Make python package namd and version compliant with naming convention…

tfredian:
…s and rules

### DIFF
--- a/mdsobjects/python/setup.py
+++ b/mdsobjects/python/setup.py
@@ -24,7 +24,11 @@ def getRelease():
         release=mdsplus_version
     except:
         pass
-    return (release,name)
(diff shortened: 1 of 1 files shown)


zack-vii:
Well, this does imply that the package name is now lower case "mdsplus" and cannot be imported using "MDSplus". This will cause errors in all third party programs (easy to fix however). Is this intended? I like the idea of having the package name lower case as I always struggle to type it correctly when manually testing things. (MDsplus, MDSPlus,...).

tfredian:
Hmmm... Josh questioned this as well and I was sure it didn't break it 
but it looks like I was wrong. I will fix this.

-tom

PR #260, opened 2015-12-16
Fix python module name to leave it as MDSplus while the egg is renamed to mdsplus[_branch]-version

tfredian:
As shown below the MDSplus module is still imported using the MDSplus name but the egg has the lowercase mdsplus[_branch]-version name. The module name being MDSplus may not be the preferred name but it would break a lot of applications if we renamed it to be lowercase mdsplus. Until the python developers decided to enforce lowercase module names this should be fine.

Python 2.7.10 (default, Sep 24 2015, 17:50:09) 
[GCC 5.1.1 20150618 (Red Hat 5.1.1-4)] on linux2
Type "help", "copyright", "credits" or "license" for more information.

> > > import MDSplus
> > > MDSplus.**file**
> > > '/usr/lib/python2.7/site-packages/mdsplus_alpha-1.2.3-py2.7.egg/MDSplus/**init**.pyc'


zack-vii:
would it make sence to do the same for RfxDevices, MitDevices, etc

PR #263, opened 2015-12-17
GIL workover

zack-vii:
- python can call tdi which can again call pyhton (unix,win)
- tdi can create a python threat (unix,win)
- tcl can do python methods (unix)
- devices like CYGNET4K do not lock themselves when worker is running (unix)

### DIFF
--- a/mdsmisc/pycall.c
+++ b/mdsmisc/pycall.c
@@ -20,7 +20,7 @@ static void (*PyRun_SimpleString) (char *) = 0;
 #define loadrtn(name,check) name=dlsym(handle,#name);	\
   if (check && !name) { \
   fprintf(stderr,"\n\nError finding python routine: %s\n\n",#name); \
-  PyGILState_Ensure=0;\
+  Py_Initialize=0;\
   return 0;\
 }
 
@@ -73,17 +73,17 @@ EXPORT int PyCall(char *cmd)
     loadrtn(PyRun_SimpleString, 1);
     loadrtn(PyGILState_Release, 1);
   }
-  (*Py_Initialize)();
-  if (!(*PyEval_ThreadsInitialized)())
-      (*PyEval_InitThreads)();
-  GIL = (*PyGILState_Ensure)();
+  Py_Initialize();
+  if (!PyEval_ThreadsInitialized())
+      PyEval_InitThreads();
+  GIL = PyGILState_Ensure();
 #ifndef _WIN32
   old_handler = signal(SIGCHLD, SIG_DFL);
 #endif
-  (*PyRun_SimpleString)(cmd);
+  PyRun_SimpleString(cmd);
 #ifndef _WIN32
   signal(SIGCHLD, old_handler);
 #endif
-  (*PyGILState_Release)(GIL);
+  PyGILState_Release(GIL);
   return 1;
 }
--- a/mdsmisc/pycall.c
+++ b/mdsmisc/pycall.c
@@ -6,14 +6,16 @@
 #ifndef _WIN32
 #include <signal.h>
 #endif
-static void *(*PyGILState_Ensure) () = 0;
-static void *(*PyGILState_GetThisThreadState) () = 0;
-static void (*Py_Initialize) () = 0;
-static void (*PyEval_InitThreads) () = 0;
-static void (*PyEval_ReleaseThread) (void *) = 0;
+typedef void* PyThreadState;
+static void (*Py_Initialize)() = 0;
+static int  (*PyEval_ThreadsInitialized)() = 0;
+static void (*PyEval_InitThreads)() = 0;
+static PyThreadState *(*PyGILState_Ensure)() = 0;
+static void (*PyGILState_Release)() = 0;
+static PyThreadState *(*PyEval_SaveThread)() = 0;
+static void (*PyEval_RestoreThread)() = 0;
+
 static void (*PyRun_SimpleString) (char *) = 0;
-static void (*PyGILState_Release) (void *) = 0;
-static void (*PyEval_SaveThread)() = 0;
 
 #define loadrtn(name,check) name=dlsym(handle,#name);	\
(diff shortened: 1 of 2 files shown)


zack-vii:
do not merge yet

tfredian:
This is all very complicated. I assume you did a lot of testing of this? Why is it necessary to muck with python calls inside TreeDoMethod since the GIL handling should be done when TdiShr invokes the PyDoMethod.py as a fun. I can't imagine why TreeShr has to deal with that.

zack-vii:
yea you are right the thing is you should take GIL when you open a python 
session and reseals the lock when done. on the other hand when coming from 
python you are supposed to invoke PyEval_SaveThread when leaving the Python 
environment to allow the called code to use the python interpreter if needed. 
and restore the lockstate once the external function is complete. My assumption 
was that DoMethod would be the right interface. and id did work.. however It was 
very late so I could only do partial testing with individually compiled 
libraries. It seemed to work on all fronts but today it turned out that the 
device still has a problem every so often. Tht is why I would not want it to be 
merged yet. In general it seems to be the right way. the SaveThread part however 
is at the wrong place. For Cygnet4K, not using the save thread will prevent the 
Tree opening in TreeOpen of cammdsutils. I assume the Saving of the thread 
should envelope the external call. the SaveThread in pycall/TdiExtPython just 
worked as it was saving the thread just after acquiring the lock but this will 
also release the lock allowing other threads to mess with the memory (probably 
causing the segfaults when calling python code from python vial tdi calls). 
Maybe it is worth putting the dynamic python lib load (replacing Python.h) into 
a shared lib/module and envelope any call of external code from python into a 
SaveThread/RestoreThread combo. I am just not sure yet how to do this the most 
elegant/robust way. I will try and test the implementation for cygnet4k's 
treeopen tomorrow. If this is succesful, this is probably the way to go.

zack-vii:
I think this is the flow of the lock

```
        GILEnsure  pyhton code  SaveThread  run ext. code  restoreTread  more python  GILRelease
haveLock       |___________________|                         |__________________________|
releasedLock__/                     \_____    _____    _____/                            \_____
externLock                                \__/     \__/
                                  python code by external process
```

If this does not corrupt any other method I suggest to merge this and fix the unclean drivers instead. Perhaps, the SaveThread and RestoreThread can also be called in python just before the function call to a CDDL.

zack-vii:
Hm they turned off the cameras over Christmas. I will look into it again and test on the cameras in January. Thing is, the current alpha segfaults in the PyEval_SaveThread with fstate is NULL error. i.e. it cannot save the thread state since it does not hold the GIL.

tfredian:
Under what condition does PyEval_SaveThread segfault??? The tests I have done with the current alpha all work fine including executing python fun's via tditest as well as testing them with python> Data.execute('call-python-fun()') (i.e. python->tdishr->python).

zack-vii:
Alright, I made a minimal example that generates the Fatal Python error: PyEval_SaveThread: NULL tstate.
The chain is: 
$ python is calling TDI function (funtest.fun via TdiExecute) -> the TDI function is calling a python function (mypack.pytest via pyfun) -> which is doing some dummy python operations.

## python commands

calling python code via pyfun

``` python
>>> import MDSplus
>>> MDSplus.TdiExecute("funtest('a')")
Fatal Python error: PyEval_SaveThread: NULL tstate
Aborted
```

calling python code from tdi dir

``` python
>>> import MDSplus
>>> MDSplus.TdiExecute("pytest('a')")
a
"<class 'MDSplus.mdsscalar.String'>"
>>>
```

## mypack package

/usr/local/lib/python2.7/dist-packages/mypack/pytest.py ; /usr/local/mdsplus/tdi/pytest.py

``` pyhton
def pytest(arg):
    print(str(arg))
    return(str(type(arg)))
```

/usr/local/lib/python2.7/dist-packages/mypack/__init__.py

``` pyhton
from .pytest import pytest
```

## TDI function

/usr/local/mdsplus/tdi/funtest.fun

```
fun public funtest ( in _var )
{
  return(pyfun('pytest', 'mypack', _var ));
}
```

zack-vii:
This is just a suggestion: Could it be that due to the second layer of c code. pycall  get a new instance. The functions have not yet been initialized and the code enters line 32 in pycall.c. Since the python is already initialized for threads PyEval_InitThreads is a NOP and since the process does not hold the GI-lock, PyEval_SaveThread in line 77 will fail to store the state which is NULL if GI-lock is not hold.
Strange enough. For each 'void\* state = PyEval_SaveThread()' call there is supposed to be a 'PyEval_RestroreThread(state)' which I cannot find.

tfredian:
I totally missed that you were doing a pyfun() call inside the 
funtest.fun! You are correct. The problem is doing the initial 
PyEval_SaveThread() which only works if done once after threads are 
initialized. I stuck in a check using |if (PyEval_ThreadsInitialized|() 
==0 ) before doing the initial PyEval_SaveThread() and it seems to have 
fixed this.

zack-vii:
I finally got the chance to test it again. There is some issue in pycall still. I made a branch where I attempt to make the python initialisation of TdiTxtPython and pycall as similar as possible. I have one question concerning the !_WIN32 specific calls before and after the python call.
pycall.c

``` c
#ifndef _WIN32
  old_handler = signal(SIGCHLD, SIG_DFL);
#endif
...
#ifndef _WIN32
  signal(SIGCHLD, old_handler);
#endif
```

TdiExtPython.c

``` c
#ifndef _WIN32
  struct sigaction offact = {SIG_DFL, 0, 0, 0, 0};
  struct sigaction oldact;
  stat=sigaction(SIGCHLD, &offact, &oldact);
#endif
...
#ifndef _WIN32
  stat=sigaction(SIGCHLD, &oldact, NULL);
#endif
```

What are their purposes and why is it different for both implementations?

zack-vii:
obsolete due to merge #272 and pull-request #279

PR #266, opened 2015-12-18
retry with dummy argument only if TypeError occured on method Type

zack-vii:
exc.message =
method + "() takes exactly 2 argument (1 given)"


tfredian:
Take out the "as exc".in line 24.

zack-vii:
it now fails on tests.treeUnitTest.treeTests but this seems to be something different. At least I don't see why the last edit should have changed anything.

tfredian:
I guess the tree instance garbage cleanup is still not correct. When I try python setup.py test on alpha and after merging this pull request on an fc22 vm it works fine doing 200+ iterations and yet in the pull-request-test jenkins job (done on a fc22 docker container) it fails. I am going to merge this request since it should be unrelated to this failure.

PR #276, opened 2015-12-30
Andrea testing

AndreaRigoni:
Hi all,
 I would make a pull to update feature-testing, getting closer to alpha and integrating your great recent changes..

These commits fix many of the python and windows issues in performing tests.
Valgrind output for forked process is also fixed.

To see a working example please take a look into dev_testing_2 jenkins build
The idea behind is to perform tests form a common source direcotry, leaving each possible build flavor in its own build directory.. The VPATH method was used.. The tests results remains there for now and must be picked up by jenkins manually. We could think a way to install them in datadir if it is necessary.

the step the user could do are:
1) clone the code in a source dir and make a project in the favourite ide.
2) make a set of builddirs from the parent of the source dir
3) from a builddir call sourcedir/configure --with-docker-image=(theimage)
4) simply call make from the build directory to build transparently into the docker image
5) call _make docker stop_ / _make docker start_ to control the container that is running underneath
6) call _make docker shell_ to go inside the container (or make docker shell USER=root for privileges)
7) _make tests tests-valgrind tests-valgrind-suppressions_ (to perform tests)
8) _make docker inspect_ (to see the ip address of the container session and set the gdb client to make remote debugging into the docker image)

to see how it was implemented you could look in the m4 directory for the following files:
m4_ax_mdsplus_dockerbuild.m4  (for the docker stuffs)
m4_ax_mdsplus_testing.m4 (for testing behaviours .. not really complete yet)
m4_ax_valgrind_check.m4 ( for valgrind targets generation )

A further script could be created to ease the build step of the user, an example is in deploy/mdsplus_build_tests, it would be great if the docker image could provide a configuration
for that instead of doing the entire build process.. 
I would eventually think of a python gui that compiles that configuration before the configure starts..
like cmake does..

I hope it could be a worthful mechanism for our development process.

Happy new Year !! :fireworks:


tfredian:
Hi Andrea,

Great!!! I looked at the dev_testing_2 jenkins build earlier today and 
saw it was really close to being complete!
  It will certainly help make MDSplus more robust if we do more 
extensive testing before issuing new releases. For now, I'm mostly 
interested in changing the release mechanism and and pull-testing to use 
the new testing scheme run in canned jenkins jobs. If we get that 
working and someone wants to run tests locally on their own system they 
could use those jenkins jobs as a template for doing it locally. For 
some of us, using python tests for much of the C code is the simplest 
method for extending our test coverage of the C based libraries so we 
will likely add more python tests. I just managed to make a fc22 docker 
image which contains running versions of both 64-bit and 32-bit pythons 
so we could work on extending the python testing to both 32-bit and 
64-bit. It is somewhat complicated to get python installed for both 
architectures. On fc22 I managed to do it by adding the following to the 
fc22 dockerfile. I will be committing this change soon after I do some 
more testing:

RUN dnf download python.x86_64 python.i686 numpy.i686; \
     rpm -e --nodeps python; \
     rpm -i python_.i686.rpm; \
     rpm -i --force numpy_.rpm; \
     mv /usr/bin/python2.7 /usr/bin/python-i686; \
     rpm -i python*.x86_64.rpm

If you just install python.i686 on a 64-bit system the main python 
program remains a 64-bit app. So I found the
above steps necessary to get it working. Then to run the tests as 32-bit 
you would need to add the mdsplus/lib32 directory to LD_LIBRARY_PATH and 
then run python-i686 setup.py test.

We will eventually also need to look into using the sanitizer options as 
yet another pass at the tests.

Thanks for the great work!

Ciao and Happy New Year,
Tom

AndreaRigoni:
2015-12-30 18:34 GMT+01:00 Tom Fredian notifications@github.com:

> also need to look into using the sanitizer options as
> yet another pass at the tests.

I did not see the reply ..

Great !! I will integrate new packages to see if tests are able to include
the new python path..

the followings are the commands I added for the tests .. I will do a merge
asap with yours.

# testing

RUN dnf install -y \
                   perl-Test-Harness.noarch \
                   valgrind valgrind.i686

# testing python

RUN pip2 install nose tap tap.py

# to run debugger in container

RUN dnf install -y \
                   gdb gdb-gdbserver \
                   openssh openssh-server

## 

(:2,
  Andrea Rigoni Garola
  mob: +39 347 0672536

" Il successo si cela nel metodo, la rovina nell'abitudine. "

PR #284, opened 2016-01-05
Initial support for D-tAcq ACQ425ELF and ACQ480FMC (untested)

bstandley:
This adds device support for the ACQ425 and ACQ480.  The former has been tested on the physical hardware with an older digitizer for comparison.  The latter will be tested in the near future.

The initial version supports pre/post trigger samples, trigger source (front panel only) and clock source (system internal only) and edge settings, internal clock frequency, input ranges (read only), and multiple ADC modules per chassis.  No additional software beyond the factory install is required.  For details on what is implemented see acqfmc.py.

The device structure (nodes and the code itself) is similar to that of the CPCI boards, but it was not possible to make a direct extension due to the significant differences between the two generations.  For example, in the case of multiple (possibly heterogenous) ADC modules in one chassis, some knobs are belong to the chassis and some to the modules.  I came up with an abstraction where the device associated with one designated module (indicated by the SYS_CTRL node) handles the chassis.  This aspect will surely require some additional work in the future.

I included a "super" class with the idea that the CPCI code could be updated and maintained alongside the FMC code (probably under slightly different names, i.e. ACQ196S vs. ACQ196).  Perhaps we can discuss this in the future.

### DIFF
--- a/tdi/W7xDevices/__init__.py
+++ b/tdi/W7xDevices/__init__.py
@@ -10,3 +10,5 @@ W7xDevices
 
 from bnc845 import BNC845
 from qc9200 import QC9200
+from acq425 import ACQ425
+from acq480 import ACQ480
--- a/tdi/W7xDevices/acq425.py
+++ b/tdi/W7xDevices/acq425.py
@@ -1,3 +1,5 @@
+import time
+
 import MDSplus
 import acqfmc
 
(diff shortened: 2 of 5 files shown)


tfredian:
There was a problem with docker on build system which cause the test build to fail. Will try to trigger a new build.

joshStillerman:
Brian -
  Great you are working on this.  I have not looked at the code yet, but a few things occur to me:

1 - For fast digitizers the pre-post mode is unreliable (check with Peter)
2 - I am reworking the acq devices to throw errors rather than return statuses.  If you wait a few days, you can model the new devices on acqxxx.  The changes to the device specific methods are minor, but the code cleans up nicely.

Ahh - now that I read your description more carefully I see that it is not a subclass of ACQ, which is fine.  You may want the newer error scheme regardless.  

I really like the idea of doing all of this without having to press files onto the board.  I will look through the code tomorrow when I get in.

-Josh

bstandley:
I made quite a few changes based on testing the ACQ425 and ACQ480 side-by-side.  You are right, pre-trigger samples are buggy.

tfredian:
Hi Brian,

I guess none of us had any idea of what to do with this pull request and since nothing was done with it since it was created back in January I'm going to close it. It is probably so far behind to be merged into the current alpha anyway. Feel free to update it and reopen if you really want some action done with it.

-tom

Issue #290, opened 2016-01-08
Add (Python-based) device hangs in 7.0.204 in traverser, mdstcl, and python


bstandley:
In MDSTCL, the following hangs in the latest alpha:
edit treename
add node nodename /model=ACQ196

I think it is related to the new Py.py from commit 5280c4b.  7.0.203 is OK.

zack-vii:
hm python directly fails as well:

``` python
import MDSplus
tree = MDSplus.Tree('test',-1,'new')
tree.addDevice('A','ACQ196')
```

zack-vii:
fixed by #293

PR #293, opened 2016-01-08
Fix hanging when adding python device by using a python tdi fun to ad…

tfredian:
…d it instead of using the py() tdi fun.


tfredian:
This should fix this issue: https://github.com/MDSplus/mdsplus/issues/290

zack-vii:
adding a python device does not hang anymore and successfully add the device

bstandley:
Tested to work with MDSTCL, Python, and jTraverser.  The motif-based traverser does not hang but does give an error

> Error adding device instance of ACQ196 from MitDevices: %TREE-E-PARSEERR, Invalid node search string.

but that is not critical for us . . .

tfredian:
The problem with adding a python device via traverser should be fixed now. The traverser use the full path specifier and the DevAddDevice function was putting in an extra backslash at the beginning of the path when using the execute(string). For the python devices it tries to add it without using execute() so the extra backslash was incorrect in that context.

Issue #292, opened 2016-01-08
TreeNodeArrays from children_nids and member_nids do not preserve tree ref


zack-vii:
``` python
import MDSplus
tree = MDSplus.Tree('test',-1,'new')
tree.addNode('A','STRUCTURE')
tree.write()
tree.quit()
tree = MDSplus.Tree('test',-1)
node = tree.top
nodes = node.getChildren()
print(nodes[0],node)
# tree holds the ref to tree -> ctx is valid
del(tree)
# node.tree still holds the ref to tree -> ctx is still valid
print(nodes[0],node)
del(node)
# nodes.tree only holds the pointer information in ctx -> ctx is invalid
print(nodes[0])
```

the del(tree) is similar to exiting a function returning 'nodes'. outside of the function the return value is broken.

zack-vii:
expected output:

```
(.A, \TEST::TOP)
(.A, \TEST::TOP)
.A
```

current output:

```
(.A, \TEST::TOP)
(.A, \TEST::TOP)
None
```

zack-vii:
fixed by #294

PR #294, opened 2016-01-08
Fix tree garbage collection by ensuring that TreeNodeArray instances …

tfredian:
…created from TreeNode methods include a reference to the tree used to create the TreeNode instance. In this way the tree context will not be garbage collected when the TreeNode instance is destroyed.


zack-vii:
fixes issue #292

PR #295, opened 2016-01-11
zck-cygnet4k-OSError-and-early-return-on-error

zack-vii:
- replaced WindowsError with more general OSError
- replaced as exc with sys.exc_info()
- do not use reserved DevPYDEVICE_NOT_FOUND

### DIFF
--- a/device_support/camera_utils/cammdsutils.cpp
+++ b/device_support/camera_utils/cammdsutils.cpp
@@ -47,9 +47,9 @@ class SaveFrame {
 		this->frameTime = frameTime;
 
         hasMetadata = true;
-		this->frameMetadata = frameMetadata; 
-		this->metaSize = metaSize; 
-		this->metaNid = metaNid;  
+		this->frameMetadata = frameMetadata;
+		this->metaSize = metaSize;
+		this->metaNid = metaNid;
 
 		this->treePtr = treePtr;
 		nxt = 0;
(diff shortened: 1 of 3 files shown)


zack-vii:
Cannot connect to the Docker daemon. Is 'docker -d' running on this host?

zack-vii:
are you adding this as well?

tfredian:
Hmmmm, I'm wondering if you went from one reserved exception to another. 
The TclFAILED_ESSENTIAL status is used to designate to the dispatcher 
that an "essential" error has happened and the state machine should not 
proceed to the next phase. Perhaps DevINV_SETUP? We could also add 
another status message if necessary. We're trying to add more generic 
messages in the future that could be used in new devices so if you need 
a new message let us know. (i.e. DevDEVICE_IO_ERROR "Error communicating 
with device. See log file for more details.")

zack-vii:
DevPYDEVICE_NOT_FOUND seems to be used if the class is not found. The desired error message should be more hardware related, since the actual device is not found, i.e.

> DevPYDEVICE_CONNECTION_FAILED "Could not connect to device. Maybe it is not connected or turned off."

 the TclFAILED_ESSENTIAL is currently used as the default error and I did not want to come up with a new one. sure, the use of a more generic error would be nice.

tfredian:
I added a pull request to add DevDEVICE_CONNECTION_FAILED exception. I'll merge that pull request if it passes the build tests. You can try using that exception instead of these two "reserved" exceptions.

PR #300, opened 2016-01-12
Fix typo

tfredian:
So close but a stupid typo prevented it from working. Guess we were lucky (unlucky) with the data we used for testing!


joshStillerman:
GREAT CATCH TOM !

bstandley:
Tested to work.

PR #301, opened 2016-01-12
Test compression

joshStillerman:
Add a testing tree to /mdsplus/trees.

Add tests reading from that tree and writing to nodes in the pyttest tree that are compressed

Use os.envrion  to deal with tree environment variable names instead of Data.execute('setenv...')


tfredian:
Looks good. May I suggest perhaps adding structure to the testing tree such as:

.compression
       WORD
       FS_FLOAT
       SIGNAL
       ...

Then:

```
  signode = self.pytree.SIG_CMPRS
  nodes = t.getNodeWild(".compression:*")
  for node in nodes:
      signode.record=node.record
      assertEqual((signode.record == node.record).all())
```

This would enable us to add more compression tests by simply adding more testing tree compression nodes. The extra layer of .compression would enable us to add other nodes for other test purposes in the future if desired.

I don't think you need both num_cmprs and sig_cmprs since I believe you can write arrays into signal nodes.

joshStillerman:
Good suggestions, the only thing is that then this would be 1 test, and it would not tell you which one failed.  There is probably an idiom for that though

From:  Tom Fredian
Reply-To:  MDSplus/mdsplus
Date:  Tuesday, January 12, 2016 at 3:14 PM
To:  MDSplus/mdsplus
Cc:  Josh Stillerman
Subject:  Re: [mdsplus] Test compression (#301)

Looks good. May I suggest perhaps adding structure to the testing tree such as:

.compression
 WORD
 FS_FLOAT
 SIGNAL
 ...

Then:
  signode = self.pytree.SIG_CMPRS
  nodes = t.getNodeWild(".compression:*")
  for node in nodes:
      signode.record=node.record
      assertEqual((signode.record == node.record).all())
This would enable us to add more compression tests by simply adding more testing tree compression nodes. The extra layer of .compression would enable us to add other nodes for other test purposes in the future if desired.

I don't think you need both num_cmprs and sig_cmprs since I believe you can write arrays into signal nodes.

PR #321, opened 2016-01-19
Twf add compress segments

tfredian:
This pull request should add the ability to compressed segmented records when using: TCL> COMPRESS treename/SHOT=shot.


tfredian:
Testing

PR #324, opened 2016-01-20
Feature testing

AndreaRigoni:
This is to finally join back feature-testing with alpha.

some features was added:
1) the possibility to detach build directory from source directory
2) the possible use of a docker build .. a container running underneath used as a make shell
3) testing and valgrind testing with tap output for c cpp and python


tfredian:
This is HUGE! Quite a major piece of work. I'm going to merge it in and we can see what (if anything) breaks. This should pave the way to a much better build, test and release scheme! Thanks Andrea!

PR #337, opened 2016-01-27
Remove test of uninitialized status variable when doing compress/clea…

tfredian:
…n of segmented data


zack-vii:
I tested it on one of our unix systems debian(weezle x64) and on win7 x64. both times no errors. seems like it fixed the problem. I will inject it into out server tomorrow and see if anything else pops up. Thanks for the fast fix.

PR #341, opened 2016-01-28
the ,1 after the ipaddress in the connect is breaking things

joshStillerman:


tfredian:
Should this actually be checking the return value of the mdsconnect? I suspect the original intent of the ,1 was to put it at the end of the expression so it always evaluated to 1. I'm not sure of the ramifications but perhaps it should be testing the return status of the connect and if < 0 putting out an error message and abort().

PR #350, opened 2016-02-08
change exceptions from class references to instances.

joshStillerman:
Change exceptions that are raised from classes to instances.

This allows added arguments to be specified, which in turn means that we can remove some of the prints for error conditions in the code.


tfredian:
Eventually you should look into making these devices python3 compatible. I noticed a few things that won't work in py3.
1) print "xxxxx" must be changed to print("xxxxx")
2) except Exception,e do something with e must be changed to:
    except Exception:
        e = sys.exc_info()[1]

PR #357, opened 2016-02-17
fix for .uid clean issue

AndreaRigoni:
now it purges the .uid files on clean .. this should fix the problem with 32/64 until the build is done separately .. maybe macos will have issues as the versions are mixed up there?


tfredian:
Slightly different pull request https://github.com/MDSplus/mdsplus/pull/359

Issue #358, opened 2016-02-17
treeshr's TreeSetUsage is broken


zack-vii:
``` python
import MDSplus
t = MDSplus.Tree('test',-1,'new')
n=t.addNode('S','TEXT')
t.write()
t.quit()
t = MDSplus.Tree('test',-1,'edit')
s=t.getNode('S')
s.setUsage('SIGNAL')
```

throws 

> MDSplus.mdsExceptions.treeshrExceptions.TreeILLEGAL_ITEM: %TREE-E-ILLEGAL_ITEM, Invalid item code or part number specified

zack-vii:
Solved, Thanks a lot ;)

PR #360, opened 2016-02-17
Fix set usage by breaking from switch statement

tfredian:


tfredian:
Fix for issue: https://github.com/MDSplus/mdsplus/issues/358

PR #372, opened 2016-02-24
Fix saved reversed flag

GabrieleManduchi:


GabrieleManduchi:
This fix should resolve the problem about reversed jScope reported by Timo

PR #393, opened 2016-03-01
cygnet4k update on store timeout workaround

zack-vii:
Dispatcher closes tree when process exits, upload thread may be able to reopen the tree


zack-vii:
is there a way of telling jerkins to resend the testresults

PR #394, opened 2016-03-02
fix Event object race condition.

AndreaRigoni:
fixes Event object and  MdsEventTests


tfredian:
Oops.... this broke labview. It has a routine called mdsplus_event_abort 
which calls the abort method of the event and this method has been 
removed. Not exactly sure what it should do instead. Perhaps delete the 
event object?

Ciao,
Tom

PR #405, opened 2016-03-07
Dev gm

GabrieleManduchi:
This time I hope not to break anything.......


tfredian:
Gabriele, Unless I'm mistaken this does not look correct either. I'm not sure why your pull request would contain a bunch of unrelated commits.

PR #409, opened 2016-03-10
Fix MARTE_DEVICE deviceButtons

GabrieleManduchi:
More info on device apply error


tfredian:
Gabriele, there was an error in the java if you didn't notice the error from the pull-request-test:
DeviceComponent.java:140: error: fullpath has private access in NodeInfo
                    System.out.println("at node: " + subtree.getInfo(nidData, Tree.context).fullpath);

GabrieleManduchi:
Now it should compile...

PR #413, opened 2016-03-15
adds check for JAVA_HOME eventually set JDK_DIR to JAVA_HOME

AndreaRigoni:
This adds a m4 macro that looks for JAVA_HOME in envs .. JAVA_HOME is the standard variable pointing to the JDK directory installed .. if it doesn't find JAVA_HOME set it test the path of the javac executable.

the following are the default java variable that are usually defined:
- JAVA_HOME: must point to installation directory of JDK.
- JRE_HOME: must point to installation directory of JRE.
- CLASSPATH: contains libraries path which JVM will look for.


AndreaRigoni:
Not working everywhere ..

PR #425, opened 2016-03-22
Add with statement support to Connection

tfredian:


tfredian:
Provides:

with MDSplus.Connection as c:
   ans=c.get('42')

to close connection after with block.

PR #432, opened 2016-03-31
Cleanup 428

zack-vii:

PR #433, opened 2016-03-31
fixedY replaceAll is not in-place

zack-vii:
and regexpr requires another level of \escape


zack-vii:
this is an essential fix for #432

PR #435, opened 2016-04-01
cleaner __getattr__ for TreeNode

zack-vii:
- removed unreachable code
- replaced try-pass with proper check
- handle 'nid' in dict by reordering

### DIFF
--- a/mdsobjects/python/treenode.py
+++ b/mdsobjects/python/treenode.py
@@ -139,19 +139,13 @@ class TreeNode(_data.Data):
             except:
                 pass
         name = name.lower()
+        if name in self.__dict__.keys():
+            return self.__dict__[name]
         if name == 'nid':
             try:
-                return self.__dict__['nid']
-            except KeyError:
-                try:
-                    return self.tree.getNode(self.tree_path)
-                except:
-                    return None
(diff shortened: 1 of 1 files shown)


tfredian:
Hi Timo. You might want to rebase this pull request. I was planning to do a major rework of the tree.py and treenode.py after I completed work on the new MDSplus build system Andrea and I have been working on. There are lots of improvements that can be made some of which I had experimented with when I was attempting a python based tdi implementation. Some include huge performance improvements when looping over TreeNodeArrays as well as calling _TreeGetNci directly instead of going through tdi for node properties. Much of the original implementation was done using the easiest path to completion but unfortunately did not end up with very efficient code performance wise. Once this request is rebased I'll take a look and merge but just wanted to give you a heads up on a potential complete rewrite on treenode.py and tree.py (and perhaps others) so you don't spend a lot of time and effort working on this.

zack-vii:
This is probably what happened to gab as well. It all looks fine and then you do a pull request from you local repo. instead of pushing git syncs and pulls the old un-rebased version from the origin - puff you have the mess. I will not actively ty to improve python then, I just stumbled upon this while writing a subclass of TreeNode.

PR #448, opened 2016-04-14
Add connection tests to python test suite

tfredian:


tfredian:
The connection test uses the LOCAL:// protocol which requires the buildroot bin directory to be in the PATH. This appears not to be set correctly when executing the tests in the build and pull-request test jobs so it is not finding the protocol correctly. I'm closing this until the test environment is corrected.

PR #460, opened 2016-04-25
prevent infinite loop (stack)

zack-vii:
If _value is not defined an infinite loop is generated. This may occur
in __new__ executing line:    return super(Scalar,cls).**new**(cls).
Not sure if it should throw and AttributeError or just return None

### DIFF
--- a/mdsobjects/python/mdsscalar.py
+++ b/mdsobjects/python/mdsscalar.py
@@ -40,6 +40,7 @@ def makeScalar(value):
     raise TypeError('Cannot make Scalar out of '+str(type(value)))
 
 class Scalar(_data.Data):
+    _value = None
     def __new__(cls,value=0):
         try:
             if (isinstance(value,_array.Array)) or isinstance(value,list) or isinstance(value, _N.ndarray):


tfredian:
Hi Timo, Can you give an example of how this might occur? I assume it will happen if you have a Scalar instance that has no _value but not sure how you create one of these short of doing del instance._value.

zack-vii:
If I would know a simple example. It is happening in a multithreaded program. I could debug up to the super.**new**() line which is probably calling the new of object. breakpoints in the first line of **init** or **getattr** are not caught. However if it happens on python 2.7 on both Windows and unix. I will try to generate the issue in a light version

zack-vii:
hmm its as simple as that the multiprocessing pool cases the error somehow.

``` python
if __name__ == '__main__':
  from multiprocessing import pool
  from MDSplus import String
  p=pool.Pool(1)
  # run w/o pool
  print(map(String,['abc','def']))  # results: ["abc", "def"]
  # run in pool
  print(p.map(String,['abc','def']))  # stack-overflows on alpha
  # results in ["abc", "def"] with the patch applied
```

tfredian:
Strange indeed. Can you try doing a slightly different patch? Might be less likely to get lost when I quickly forget why that code in **getattr**() was added ;-)

class Scalar(_data.Data):
    _value=None

zack-vii:
hm however would this not throw the error 

```
AttributeError: 'NoneType' object has no attribute '_value'
```

in the last line of **getattr**

zack-vii:
ah no it would not even call **getattr** right as it is invoked only if the field has not been found

tfredian:
yep. I tested it with your example and seemed to work. Can you try it with your real application to verify that is fixed as well?

zack-vii:
so far it is running normally.

PR #463, opened 2016-04-26
Fix pending lock in TreeSetXNci

GabrieleManduchi:
This change is required to leave NCI unlocked. It did not show using a single thread because that is a recursive lock. However a separate thread of the same process is blocked after the first call to TreeSetXNci

### DIFF
--- a/treeshr/TreeSegments.c
+++ b/treeshr/TreeSegments.c
@@ -1579,6 +1579,8 @@ int _TreeSetXNci(void *dbid, int nid, char *xnciname, struct descriptor *value)
       SeekToRfa(attributes_offset, local_nci.DATA_INFO.DATA_LOCATION.rfa);
       local_nci.flags2 |= NciM_EXTENDED_NCI;
       TreePutNci(info_ptr, nidx, &local_nci, 0);
+//Required because TreeGetNciLw is holding the lock
+      TreeUnLockNci(info_ptr, 0, nidx);
     } else {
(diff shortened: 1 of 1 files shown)


tfredian:
Hi Gabriele,

```
Could you resubmit this pull request with a slight change? Instead of adding the unlock inside the if block it seems that simply removing the else block and putting the unlock after the if block would be a cleaner fix for the problem. Agree?
```

-tom

GabrieleManduchi:
Hi Tom,
    right, I reissued the pull request

```
                             Gabriele
```

On 26/04/2016 14:11, Tom Fredian wrote:

> Hi Gabriele,
> 
> |Could you resubmit this pull request with a slight change? Instead of adding the unlock inside the if block it seems that simply removing the else block and putting the unlock after the if block would be a cleaner fix for the problem. Agree?
> |
> 
> -tom
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub 
> https://github.com/MDSplus/mdsplus/pull/463#issuecomment-214719645

## 

Gabriele Manduchi

Istituto Gas Ionizzati del CNR
Consorzio RFX - Associazione EURATOM/ENEA sulla Fusione
Corso Stati Uniti 4, 35127 Padova - Italy
ph +39-049-829-5039/-5000 fax +39-049-8700718
mailto:gabriele.manduchi@igi.cnr.it, http://www.igi.cnr.it

PR #465, opened 2016-04-27
Update dev-kge branch from alpha

kgerickson:


tfredian:
Ummm, not sure what to do with this. I guess your dev-kge was a bit out of date? Perhaps you might want to make a new dev-kge branch out of alpha and push that to origin instead of merging 1,614 commits from alpha in to your current dev-kge branch?

PR #473, opened 2016-04-28
javamds/JavaTrav.c: Pass pointers to free without cast

kgerickson:
The free() function takes in a void\* as its argument, so casting from any pointer type to void\* is wrong, and this code casted from int\* to char*, which is more wrong.  Remove casting entirely.

### DIFF
--- a/javamds/JavaTrav.c
+++ b/javamds/JavaTrav.c
@@ -809,7 +809,7 @@ JNIEXPORT jobjectArray JNICALL Java_Database_getMembers
     (*env)->SetObjectArrayElement(env, jnids, i, jnid);
   }
   if (num_nids > 0)
-    free((char *)nids);
+    free(nids);
 
 /* //printf("\nEnd getMembers");*/
   return jnids;


kgerickson:
This build didn't seem to fail because of the file I changed:

In file included from /var/lib/jenkins/workspace/pull-request-test/mdsobjects/labview/mdsdataobjectswrp.cpp:1:0:
/var/lib/jenkins/workspace/pull-request-test/mdsobjects/labview/mdsobjectswrp.h:27:26: fatal error: platdefines.h: No such file or directory
 #include  <platdefines.h>
                          ^
compilation terminated.
<builtin>: recipe for target 'mdsdataobjectswrp.o' failed

Any idea what's going on?

tfredian:
Working on it. The pull-request-test jenkins job is still having some 
problems. Testing the pull request again now. I must have removed the 
git submodule init option so the 3rd-party-apis submodule was not pull 
in so it couldn't find the matlab headers.

PR #483, opened 2016-05-03
added optional return value to createPulse

zack-vii:
only if mode is specified, otherwise behave as before

### DIFF
--- a/mdsobjects/python/tree.py
+++ b/mdsobjects/python/tree.py
@@ -250,11 +250,13 @@ class Tree(object):
             Tree.unlock()
         return _treenode.TreeNode(nid,self)
 
-    def createPulse(self,shot):
+    def createPulse(self,shot,mode=None):
         """Create pulse.
         @param shot: Shot number to create
         @type shot: int
-        @rtype: None
+        @param mode: Optional mode, one of 'Normal','Edit','New','Readonly'
+        @type mode: str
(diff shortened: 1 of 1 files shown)


tfredian:
I don't quite understand the need for this. It seems to be a direct contradiction to what createPulse is supposed to do, namely copy an existing tree instance into a new one with a different shot number. Why would one want or need to add a mode argument to turn createPulse() into a Tree() call?

zack-vii:
well, usually you would create a new pulse from the model. and then you are using the new pulse to do actions write data etc. So this would open that new pulse right after creation. if you do not want it to open the pulse, you just omit the mode parameter and nothing has changed compared to the old behavior.
I personally use it quite often when testing cod on a tree. createPulse creates a fresh copy of my sample tree and the following code does something with that copy.

tfredian:
Hmmm, maybe just me but I'm still not convinced that saving one tree=Tree(treename,shot[,mode]) line in the application (which seems to me would make the application more readable) does not warrant adding optional behavior to the createPulse method. Any other opinions out there??? Sorry to be so stubborn.

kgerickson:
This would create a fundamentally different createPulse just for python.  While it would be backwards compatible with the concept of createPulse elsewhere, it would introduce an additional paradigm that isn't present in any other language.  This change should not be taken lightly.

PR #488, opened 2016-05-10
Kge config cleanup

kgerickson:
Moderate cleanup of some old configure issues:
- Remove support for old platforms and tools
- Update automake
- General cleanup


tfredian:
CamShr cannot build on windows (not sure about macosx) so the test for the scsi/sg.h header is still needed. Eventually I would like to change camshr to build on windows to provide the remote camac functionality and generate errors if "camac_server" env is not set and get rid of remcamshr entirely.

kgerickson:
I modified the branch to remove that commit.  Do I have to resend a new pull request?

kgerickson:
Ok, I just opened a new one

tfredian:
Actually if this is the one I closed then you could have reopened it 
after pushing changes.

PR #489, opened 2016-05-10
mdsshr: Remove double xmllib

kgerickson:
For some reason, xml libraries were being linked twice in mdsshr.  Remove one.


tfredian:
I now vaguely remember why there were two XML_LIB's in the link. On the mingw64 build the XML_LIB  has to be placed before the SOCKETLIB because it is using the status lib for xml. I recall I had to try a lot of different combinations to get this to work on windows. What I don't remember is whether the trailing XML_LIB was needed or not. I can't imagine why it would be. Pushing new commit to move XML_LIB to before SOCKETLIB.

tfredian:
The builds seemed to work after reordering the libs so merging it now.

PR #501, opened 2016-05-17
python treeshr underscores

zack-vii:
- added missing underscore to TreeFindTagEnd
- corrected underscores of function wrappers that do not follow schema.

### DIFF
--- a/mdsobjects/python/_treeshr.py
+++ b/mdsobjects/python/_treeshr.py
@@ -350,16 +350,16 @@ def TreeDeletePulse(tree,shot):
 
 def TreeRestoreContext(ctx):
     try:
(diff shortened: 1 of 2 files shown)


tfredian:
This does not work unfortunately because of the special effect of double underscores when found in a class method definition. See: http://stackoverflow.com/questions/1301346/the-meaning-of-a-single-and-a-double-underscore-before-an-object-name-in-python

If you try to run the python tests you will see the following:

```
status = _treeshr.__TreeCreatePulseFile(self.ctx,shot,0,_C.c_void_p(0))
```

AttributeError: 'module' object has no attribute '_Tree__TreeCreatePulseFile'

PR #521, opened 2016-06-07
record attribute should return GetNci's "record"

zack-vii:
otherwise it is misleading

### DIFF
--- a/mdsobjects/python/treenode.py
+++ b/mdsobjects/python/treenode.py
@@ -178,8 +178,6 @@ class TreeNode(_data.Data):
                 return self.tree.getNode(self.tree_path)
             except:
                 return None
-        if name == 'record':
-            return self.getData()
         if name == 'tags':
             return self.getTags()
         if name == 'usage':


tfredian:
Unless you discovered some case where the current implementation behaves differently than using getnci(node,"record") I'm not sure we necessary want to change this. In fact when I get the opportunity to refactor some of the python code I had intended to avoid the use of TDI's getnci entirely. The implementation of TDI's getnci(node,"record"), I believe, is to just call TreeShr->TreeGetRecord to return the data stored in the node so the current python implementation avoids the extra tdi layer of the code.

zack-vii:
Oh getData() is actually not the same as data() , so the confusion was only on my part. thinking of it would it make sense to turn the attributes into properties since the getter and setter functions are already there?

tfredian:
Unfortunately I probably used the wrong name for the getData method which is easily confused with data(). The data() method was intended to return the data as a native numpy scalar or array. My fault but probably too late to change as it may break existing user code. I agree turning many of these "attributes" into properties would be much better. Perhaps it makes sense to make note of this and incorporate it into the refactoring of the python code. Unfortunately much of this was developed when I was just started learning python so there are many places where it could be improved. If you have the time when I begin the rewrite I'll make a new branch and do a first pass and then ask you to look over it and we could work on it together.

zack-vii:
If I have the time, I am happy to do so.

PR #529, opened 2016-06-13
conversion for Uint64 date to/from python's time

zack-vii:
convert between seconds since 01/01/1970 and the MDSplus time base


zack-vii:
requires verification:
- independent of time zone -> _utc1
- correct mapping  -> _utc0

PR #533, opened 2016-06-13
add 'tags' to Device.Add

zack-vii:
allows to auto add tags to device nodes


tfredian:
I don't think it is a good idea to define tags in a device implementation. This would make it impossible to add more than one instance of a device with tags defined in a single tree and would also make the tags ambiguous if multiple subtrees contain one of these devices. Even if you made the tag name some expression of the head node of the device it could cause tag duplication problems if the same node name was used when adding multiple instances of the same device type to a tree.

zack-vii:
hm.. true

PR #543, opened 2016-06-22
added dir function to tree and treenode

zack-vii:
allows easier browsing


joshStillerman:
Cool -

   Should we add more of these?  Like

showData (TCL> show data)

-josh

PR #544, opened 2016-06-23
Twf fix mdsshr get set env

joshStillerman:
Added one more fix to this module, could not figure out how to squash the commit.  Probably had to do it before I pushed.


tfredian:
Closing and opening to trigger a retest. Request test failed on a udp events test which happens intermittently.

PR #546, opened 2016-06-29
typo in constant definitions

zack-vii:


zack-vii:
failed on

```
No 32-bit support for sanitize=thread. Skipping.

Failure: 64-bit santize with address failed

```

should not be related to the changes.

tfredian:
yes. We're investigating.

zack-vii:
I rebased the branch, just in case

zack-vii:
I assume the test itself or the test environment is buggy.

tfredian:
Turning off the sanitize tests since they seem to be failing when they should not.

PR #559, opened 2016-07-04
use length field for length for all other dtypes

zack-vii:
allows to send more dtypes correctly e.g. pointers

could not test it since the docker engine does not work


zack-vii:
compiled it for unix 64 and it works.

PR #560, opened 2016-07-04
ensure consistent behavior on different platforms

zack-vii:
%p results in different outputs on unix and windows systems

``` tdi
Windows64: Pointer(0000000000000000)
Unix:      Pointer(nil)

Windows64: Pointer(0001234567890ABC)
Unix:      Pointer(0x1234567890abc)
```


zack-vii:
I compiled it on unix64 and it works

PR #568, opened 2016-07-08
added tdi function that attempts to collect available shot numbers

zack-vii:


tfredian:
Good idea! Need to retest because of intermittent udpevent test failure.

tfredian:
I tried this on C-Mod but we use the shot number digit substitution feature in our archived shots to reduce the number of shots per directory as well as using ~t for the tree names. For example:

cmod_path=alcdata-test::/cmod/trees/test/~t/;alcdata-new::/cmod/trees/new/~t/;alcdata-models::/cmod/trees/models/~t/;/cmod/trees/archives/~i~h/~g~f/~e~d/~t;alcdata-saved::/cmod/trees/saved/~t/

spectroscopy_path=alcdata-test::/cmod/trees/test/~t/;alcdata-new::/cmod/trees/new/~t/;alcdata-models::/cmod/trees/models/~t/;/cmod/trees/archives/~i~h/~g~f/~e~d/~t;alcdata-saved::/cmod/trees/saved/~t/

This might be tricky to deal with though not impossible.

zack-vii:
hm well this could of course be improved I am not too familiar with the ~replacements. We could merge this and eventually add the support for ~ihgfedcba later. it could probably been dealt with recursively. is it the digits of the shotnumber?

tfredian:
Yes these are the digits of the shot number. Will merge this version though and we can see about adding the digit substitution in an enhanced version.

PR #573, opened 2016-07-26
Added support for conenction put signal

GabrieleManduchi:
A new directory MDSplus_LV2015 has been added to mdsobjects/labview in order to avoid breaking the previous set of LabVIEW VIs.  This new directory contains the required VIs for writing a signal via Connection object.


tfredian:
I assume this pull request will add quite a few new files to the installation kits which makes it a good time to inform you about one feature of the new build scripts. When a new release is created, the contents of the rpm's and deb's are checked against the expected contents listed in files found in deploy/packaging/debian|redhat. If you add or remove files from the installation kits you will need to update these files to match what you expect to see in the installers. This can be done manually but it is best to let the build script do it instead. You will need docker running on your system and then issue the following commands:

$ srcdir/deploy/build.sh --os=fc22 --release --updatepkg                   # chose a supported redhat os
$ srcdir/deploy/build.sh --os=ubuntu14 --release --updatepkg          # chose a supported debian os

These commands will build release kits and update the packaging verification files. You can use git diff on these files to make sure the changes are what you anticipated. The changes to these verification files will need to included in the pull request or the builds of new release installer kits will fail since the installer content check will produce an error.

PR #576, opened 2016-07-29
Decompile Q and QU as decimal strings human readable

zack-vii:
using formula "length \* 8 \* .29 + 2" works for 8,16,32,64, bit integers
size:  4, 6, 11, 20, 39
expanding for 128bit integer would require a special treatment of O and OU i.e. +1

### DIFF
--- a/mdsobjects/cpp/testing/MdsScalarTest_NumericConversion.cpp
+++ b/mdsobjects/cpp/testing/MdsScalarTest_NumericConversion.cpp
@@ -95,7 +95,7 @@ void test_Int64() {
 
     Data * data = new Int64(value);
     MdsDataTest::test_data_numerics(data,value);
-    MdsDataTest::test_data_string(data,"0X7bQ");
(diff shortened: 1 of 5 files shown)


zack-vii:
Is the rest due to the fact, that the output has changed. if this commit does not break other stuff I would like to have it sorted out with the test. human readable form of uint64 timestamps is so much better than the hex form.
any comments?

tfredian:
If you change the behavior of MDSplus then you also need to change the tests that depend on the current behavior. While this change is unlikely to cause problems for existing user applications we try to not change existing behavior unless the existing behavior is a bug or there is a very compelling reason to do so. If you click on the details button of the tests you'll get to the jenkins pull request test that failed which contains links to the triggered builds from that trigger job. If you look at the console output of those jobs you'll see which tests failed. Alternatively you can make; make -k tests to test locally.

zack-vii:
replaced by #579

PR #607, opened 2016-08-23
Twf fix compiler warnings step1

tfredian:


tfredian:
After modifying 172 modules and 20k lines of code the compiler warning throughout MDSplus are now gone! Any new builds will now be using -Werror so if changes to the MDSplus code causes compiler errors the builds will not succeed. When doing development work please use --enable-werror with configure. This will ensure that you are not introducing new compiler warnings.

PR #624, opened 2016-09-06
Twf refact python local

zack-vii:
last changes so far I did a rebase attempt to see what has changed. I guess best would be to checkout an fresh offspring of alpha and there checkout mdsobjects/python from the refact-python branch. I did it in
https://github.com/zack-vii/mdsplus/tree/zck_rebased_twf-refact-python
and performed the test the device adding and some other scripts so far without errors that I have not fixed yet. There is one thing with the findPyDevices. the appended \0 is not safely transferred through the numpy lib (it trims the \0 if it ist the last character in the string). This will result in an space instead of the \0 when converting it back to tdi. Not sure if this would cause a problem. one could think of omitting the \0 and handle the trailing spaces on the client side. (e.g. via trim)


tfredian:
I think the null terminator in the device name list are currently needed by the traverser. I can try to determine what else, if any, are depending is depending on that and change the code to just use strings constructed by trimming the values in the string array. The traverser is just passing a pointer to the offset in the string array as a char \* depending on that being a null terminated string. Ugly but cheap in the days when malloc'ing and copying was an expensive operation. ;-)

zack-vii:
well if one could fix traverser to trim the leading spaces the use of numpy string arrays is very nice. However, the special treatment of the trailing \0 might be tricky.

Also Conglom Dispatch and maybe others do not compile/decompile correctly, atm.

zack-vii:
fixed the conglom

tfredian:
I inadvertently merged this pull request assuming that it had been tested. With these changes if I try:

cd mdsobjects/python
python setup.py test

it segfaults!

If I revert this branch back to it's state before https://github.com/MDSplus/mdsplus/commit/772a529c70fa9c9ccf316694c4b32fc5d7c119bb
it works fine.

tfredian:
Ok. I pushed a fix for the python segfault. No idea why it just showed up but it was calling TreeFree(ptr) but the argument of TreeFree was not defined to be a c_void_p() so it must have been converting it to an integer. The tests seem to be working now.

PR #626, opened 2016-09-06
Remove all compiler warnings with using -g -O2 -Wall and -Wextra

tfredian:


tfredian:
Oops... missed ~500 more warnings in the math fortran code.... not done yet.

PR #632, opened 2016-09-12
%p has a different representation on unix and windows

zack-vii:


tfredian:
This failed the pull request tests with compiler warnings which is why it was changed back to %p in the first place. It also seems to be invalid for 32-bit platforms.

zack-vii:
assume it will cast 32 bit integer properly to 64bit integer by appending 0x00000000. I guess I could fix the compiler warning by properly casting the pointer.

tfredian:
This still does not seem to be valid for 32-bit. You are casting the in_ptr->pointer as a pointer to a 64-bit integer. On a 32-bit system this pointer would point to a 32-bit pointer so if you treat it as a 64-bit integer you would get the 32-bit pointer plus whatever else is in the next 32-bit's of random memory or possibly a segfault. Unless I'm mistaken to do this correctly you would need to figure out whether you are running on 32-bit or 64-bit and if on 32-bit get the pointer as a 32-bit integer and then cast the 32-bit integer as a 64-bit integer.

zack-vii:
Hm same holds for the current implementation
e.g. you could deserialize a 32 bit pointer on a 64 bit machine and vice versa.

``` tdi
serializein(Byte_Unsigned([4,0,51,1,8,0,0,0,1,2,3,4]))
Pointer(0000135504030201)
```

this could become an issue when using mdsip, I will adjust the implementation to cast based on the in_ptr.length.

zack-vii:
this one does the trick ?!

``` tdi
serializein(Byte_Unsigned([4,0,51,1,8,0,0,0,1,2,3,4,5,6,7,8]))
Pointer(0x4030201)
serializein(Byte_Unsigned([8,0,51,1,8,0,0,0,1,2,3,4,5,6,7,8]))
Pointer(0x807060504030201)
```

zack-vii:
This should be a safe solution as the POINTER descriptor is not bound to the architecture. 
It in a Long or QuadWord with a different dtype.
The decision whether it is 32bit or 64bit should be made based on length anyway.
Could you merge this once reviewed.

PR #648, opened 2016-10-04
Fix memory misaglinment in labview interface for 32 bits

GabrieleManduchi:


tfredian (changes requested):
Enclose #pragma pack(1) in
 #ifndef **MINGW64**
#pragma pack(1)
#endif

GabrieleManduchi:
Surreounded #pragma pack(1) directive in #ifndef **MIGNW64**  (with underscores)

GabrieleManduchi:
"**MINGW64**"

drjdpowell:
Hello, I’m trying to use the objects LabVIEW library for Mdsplus on a 32-but system.  The library works, but any misuse or error seems to crash LabVIEW.   Could the crashes be a result of this issue?

PR #652, opened 2016-10-06
Jscope fix

GabrieleManduchi:
A set of fixes for handling inconsistent expression evaluation in jScope and 2D management

### DIFF
--- a/javamds/javamds.c
+++ b/javamds/javamds.c
@@ -188,12 +188,12 @@ static void *MdsGetArray(char *in, int *out_dim, int type)
   *out_dim = 0;
   switch (type) {
   case FLOAT:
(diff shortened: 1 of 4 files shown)


zack-vii:
Actually my experience is to better have the semicolon in the expression that holds the parenthesis, i.e. 
you could revert the changes in MdsDataProvider and add the ; in javamds.c instead:

``` c
sprintf(expanded_in, "_xxx = (%s;);fs_float(_xxx)", in);
```

this way it would work regardless of the presents of terminating semicolon in the input.
note: - an additional ; does not matter and is ignored by tdi -

``` tdi
(_a=1;_b=2;_a+_b;;)==(_a=1;_b=2;_a+_b;)==3
```

PR #660, opened 2016-10-12
Zck tdi cleanup

zack-vii:
retry since to rule out something


zack-vii:
Hm I tempered the kernel.<arch> files but the issue according to the "Console Output" is that /workspace does not seem to exist and it is trying to create it without permission. is it not?

zack-vii:
temporarily closed to test out a few things

PR #669, opened 2016-10-13
Twf refact python local

zack-vii:


tfredian:
This will not fly. First of all the python/mdsExceptions/xxxExceptions.py files are all generated from xml files as noted at the top of these files. Second I'm not sure about the default status value of the exceptions being -1 or -2. I don't know if the -1 or the -2 value is checked anywhere but we should probably define an exception in an xml to use for this and use it's status value. Haven't checked out any of the rest though.

PR #670, opened 2016-10-13
Twf refact python local

zack-vii:


tfredian (changes requested):
I guess you didn't read my last comments about this. You CANNOT modify the python/mdsExceptions/*Exceptions.py files manually. As noted at the top of these files they are automatically generated. If you want add new exceptions you have to modify the xml file used to generate them and then cd to mdsshr and invoke python gen_messages.py

tfredian (commented):
Why change DevPYDEVICE_NOT_FOUND to DevPYDEVICE_NOT_FOUND2????

tfredian (commented):
Why change self.status in mdsExceptions to -2? Where is this used btw?

tfredian (changes requested):
Remove all changes to mdsExceptions/xxxExceptions.py files which were not generated by a change to the corresponding xxx_messages.xml file.

tfredian:
There is a problem with the getClass method in Conglom. It now raises a plain Exception if the device is not a python device. Changing that to raise DevPYDEVICE_NOT_FOUND fixes the problem as that is caught. This however uncovers another problem which also causes the tests to fail. The test to see if two TreeNodeArray's use to return the equiv of the comparison of two numpy arrays which returns an array of bools. It now returns a bool False even if the two TreeNodeArrays are identical.

zack-vii:
Ah sorry I did not respond to your comments, but they did not seem to be notified by email so I just tumbled upon them.

to the part with the auto-generated exceptions:
is **init** also autogenerated? any changes to other files in mdsExceptions are unintended and possibly a residual of checkout mdsobjects/python between my rebased and non rebased branch.  I only really altered **init**.py

about the error codes: 
I choose -2 as the default value for exceptions as it is not a success (status&1) and it has an unknown severity? If I should have been a success one would use MDSplusSuccess and for an Error you should use MDSplusError. which will have the code -8|2 = -6 given by the severity E for Errors. This way default errors have a non-zero status that specifies the correct severity.
You need the static status if you want to refer to the error by status code. instead of instantiating the object and grabbing is status you can access the static status field of the class directly. it is used in PyDoMethod.py

I will fix the getClass part and make the test work before preparing the rebased version with fixed package check

zack-vii:
I see die I eventually change the PYDEVICE_NOT_FOUND to PYDEVICE_NOT_FOUND2. this is because it was a duplicate and had been renamed in alpha  like this. so the alpha xmls will create the right xxxExceptions.

one thing about this 
PYDEVICE_NOT_FOUND and PYDEVICE_NOT_FOUND2 does not seem to be very useful.
Would it not make more sense to have something like 
PYDEVICE_NOT_FOUND: "Python device class not found"
for when the import package.device fails to get the desvice class
 and 
PYDEVICE_INVALID_IMAGE: "Device is not a python device."
for when conglom.image is not '__python__'
What do you think?

zack-vii:
I made a test pull-request or the fully rebased refactor #704

tfredian:
Yep, I finally tracked this down. Josh had inadvertently added what he 
thought was a new message to use in his picam device to indicate that 
the device support module/library was not installed. He is going to 
change this to a different error name. This was done at the beginning of 
the year and didn't cause any problems surprisingly. I renamed it 
PYDEVICE_NOT_FOUND2 when I was trying to remove the compiler warnings 
which the duplicated names caused. I didn't check to see where it was 
being referenced like I should instead of just renaming it. I like the 
idea of the PYDEVICE_INVALID_TYPE but I think more appropriate name 
might be NOT_PYTHON_DEVICE. Don't rename NOT_FOUND2 as Josh will 
probably use that status code.

-tom

zack-vii:
I see, I will add an Error instead of replacing PYDEV_NOT_FOUND2.
How about PYDEVICE_NOT_FOUND, PYDEVICE_NOT_SUPPORTED, NOT_A_PYDEVICE?
Is the name of PYDEVICE_NOT_FOUND2 does not seem to make problems as it was shadowed by PYDEVICE_NOT_FOUND a few weeks ago.

zack-vii:
finally the rebased version  passes the python unit test again. 
I will check in tomorrow if the build test succeeded.. ist late here I got to go home ;)

tfredian:
Getting closer. Looks like the sanitize address tests uncovered a bug in 
MdsShr. I'll do some investigation. It only showed up on 4 of the 
platforms. -tom

zack-vii:
continued in #704

PR #674, opened 2016-10-17
added uniform debug methods to device group initializer

zack-vii:
this allows to test the core shot cycle (dispatcher) independant from
without the devices
importer is wrapped arround static method of the MDSplus.Device

in MitDevices some changes to device dt100 and Acq200 where made to fit
with general schema for MitDevices (upper-case device names)

### DIFF
--- a/mdsobjects/python/mdsdevice.py
+++ b/mdsobjects/python/mdsdevice.py
@@ -91,6 +91,56 @@ class Device(_treenode.TreeNode):
     debug = _getenv('DEBUG_DEVICES')
     gtkThread = None
 
+    """ debug safe import """
+    if debug:
+        if int(debug)<0:
+            @staticmethod
+            def _debugDevice(d):
+                from MDSplus import Device
+                import types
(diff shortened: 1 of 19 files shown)


zack-vii:
cloud you verify that you do not use Acq200 and dt100 
dt100 was broken a few weeks ago and acq200 depended on it. so it is likely no one ever used them

joshStillerman:
I can verify that we are not using either of these.  I think it very 
likely that no one else is either. Eventually, I would like to remove 
not only these but all of the old d-tacq device types:

dt100
acq200

dt132
dt196a
dt196b
dt196  - we still use this
acq_ftp
dt216
dt216a
dt216b
dt216_ftp

dtao32  - this is in use, but should be redoneless
dtdo32 - this is in use, but should be redone

You can remove those 2, and I will post something asking about the rest 
of the list.

zack-vii:
well I could not remove any device due to interdependencies. I think the removal should become a separate pull together with the eventual reimplementations

PR #678, opened 2016-10-24
updated doMethod

zack-vii:


tfredian:
I'm not sure about having the print("doing %s(%s).%s(%s)"%(device,model,method,','.join(map(str,args)))) line in PyDoMethod.py. Current when you do TCL> dir/full on nodes in a device it will call do method with the original_path method to get the original name of the node when the device conglomerate was added and mdsdcl commands expect output to be returned by the command implementation so you can execute mdsdcl commands in tdi fun's and return the output. If I'm not mistaken adding this print output would make dir/full produces some messy output. It's conceivable we could add an attribute to a device method to disable the print or even easier check the method against "ORIGINAL_PART_NAME" and not do the print.

zack-vii:
final version ?

zack-vii:
Sorry about that this was not a proper pull request, I just wants have ut checked on all oses easily. The debug messages should go of course. But I ran out of time today. Will temper it tomorrow

Sent from my Windows Phone

PR #682, opened 2016-10-25
fixed key None type and key in apdadd and apdrm

zack-vii:


zack-vii:
Jenkins error does not seem to be related to changes made

PR #704, opened 2016-11-04
zck rebased twf refact python DO NOT MERGE

zack-vii:
This is only a test pull-request to trigger a jankins test build.
DO NOT MERGE BEFORE CONSULTING TOM OR ME


zack-vii:
Well it seems that fc22 does not pass the sanitizer due to a memory leak in the tdi-python-interface I tried adding a more general test in the current alpha with the same sanitizer FAIL:

``` python
self.assertEqual(Data.execute("Py('import MDSplus;a=MDSplus.Uint8(8)','a')"),Uint8(8))
```

Maybe there is a way to fix this as it only appeared on fc22 so far. I will temporarily remove the apd check. since this is not a problem introduced by the refactor of the python interface

zack-vii:
the descriptor make up works but on some oses the status cannot be mapped to an exception. I used the __new__ approach to generate the error classes based on status using MDSplusException.
265388258 (NoData) does not translate well .. possibly due some stuff related to (status & -8) and the dict lookup

PR #714, opened 2016-11-11
added Python-Tdi-Interface check and fixes discovered sanitizer issue on fc22

zack-vii:


zack-vii:
@tfredian: If it does not crash any other os this should fix the sanitizer issue on fc22.
lib in Initialize and ans_d.pointer in findModule have not been freed.

PR #723, opened 2016-11-16
Zck windows compiler warninges

zack-vii:
fixes all current code related warnings so -Werror can be used.
This might fix some Windows problems that are pthread related


zack-vii:
Testet it on windows10 and indeed it fixed the pthread problem on dispatching from mdstcl.
wfevent still takes two events to return this is the same behavior as for e.g. alpha-7.1-55

PR #725, opened 2016-11-16
Zck windows compiler warninges

zack-vii:
performs unit tests on wine when building for windows.
python tests are skipped though ?!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ bin32/
 bin64/
 bin_x86/
 bin_x86_64/
+build/
 etc/
 lib/
 lib32/
--- a/configure
+++ b/configure
@@ -12379,8 +12379,8 @@ fi
   :
 else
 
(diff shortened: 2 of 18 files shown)


zack-vii:
@tfredian: I assume one need to update the dockerimage to enable valgrind and sanitizer support, also it might be useful to use a more recent version of wine64 ::edit:: with python 2.7 installed

tfredian:
I don't know what the current state of sanitize support is for mingw64. 
Here is a link to some mention of supporting undefined: 
https://sourceforge.net/p/mingw-w64/mailman/message/32314249/

The following link indicates we should be able to run wine under 
valgrind: https://wiki.winehq.org/Wine_and_Valgrind

The dockerfile used to build the windows docker can be found at: 
https://github.com/MDSplus/Docker/tree/master/windows. If you want to 
play with getting valgrind, sanitize feel free to build a new docker 
image starting with the original dockerfile. In theory you should be 
able to override the docker image used by specifying the 
--dockerimage=image option to build.sh. If necessary we could use a 
different base os for the windows dockerimage if we need newer versions 
of any of the tools. If you come up with a new improved dockerfile issue 
a pull request and I can merge it and get dockerhub to build a new 
docker image.

PR #734, opened 2016-11-18
Add an MDSplus symlink which can be used for tests

tfredian:
The python tests now add the srcdir/mdsobjects/python/tests
to the beginning of the python sys.path. This causes the import
of the MDSplus module to refer to the srcdir/mdsobjects/python
directory ensuring that the srcdir version of the MDSplus module
is what is imported. The python tests also define PYTHONPATH
so that the import of MDSplus during tdi's dynamic invocation of
python will use the srcdir version.

modified:   deploy/platform/debian/debian_docker_build.sh
modified:   deploy/platform/redhat/redhat_docker_build.sh

Removed creation of MDSplus symlink and define the PyLib environment
variable using the version information of the python used in tests.

new file:   mdsobjects/python/tests/MDSplus

Add symlink for MDSplus module which can be used for python testing or
debugging the MDSplus python module. If the MDSplus module is not currently
installed you can tell python to load the MDSplus modul from your srcdir
by creating an environment variable as follows:

export PYTHONPATH='srcdir'/mdsobjects/python/tests

where: 'srcdir' is replaced by the directory of your mdsplus sources.

modified:   mdsobjects/python/tests/connectionUnitTest.py
modified:   mdsobjects/python/tests/dataUnitTest.py
modified:   mdsobjects/python/tests/exceptionUnitTest.py
modified:   mdsobjects/python/tests/segmentsUnitTest.py
modified:   mdsobjects/python/tests/threadsUnitTest.py
modified:   mdsobjects/python/tests/treeUnitTest.py

Use the MDSplus symlink to find the MDSplus module. The use of the names of
individual python modules in the python directory is no longer needed since
with the MDSplus module location defined you can import MDSplus directly.


tfredian:
The change to define PyLib in the tests script uncovered a memory leak due to not freeing the string passed to putenv. This will probably need to be ignored since some earlier versions of glibc do not copy the  string and therefore require the memory used when calling putenv to not be freed. I'll fix this next week.

zack-vii:
HI Tom,
I am planning to do some major changes to the deploy scripts, once this pull is ready I will craft a rebased version on this commit. The Idea is to make use of common code and functions to make it easier to evolve them.

PR #736, opened 2016-11-19
Zck deploy

zack-vii:
makes it easier to introduce more tests.
python unittests in progress are relying on it.
however they discovered some errors that cause valgrind to fail on memcheck.


zack-vii:
works. but that a tricky one. It gets stuck in an endless lo on fc23

zack-vii:
not sure if my mail was actually send:
Good Morning Tom  @tfredian
there seems to be a problem on jenkins with my last PR, only on fc23 it got stuck in a test. not sure why as I cannot reproduce it locally. But this makes me think if a time out might be maybe useful or even better a possibility to interrupt running tests that are stuck or obsolete as an other os raised an error already.

update: seems like it is the "MDSplus link"-merge that causes the "loop?" as alpha got stuck as well. I realised that there is an interactiveTerminal and an ssh access. I was wondering if I could get an account on jenkins.

cheers
Timo

PR #756, opened 2016-12-01
Fix limits error introduced by check for date

GabrieleManduchi:
Urgent fix for a bug recently introduced when checking for X limits expressed ad date


GabrieleManduchi:
In addition to the fix, I have also added the changes made by Keith to remove compiler warnings that apparently did not break anything

PR #780, opened 2016-12-29
Fix fc25

AndreaRigoni:
This should fix issues in fc25.

[ref1]
It appears that the various sanitizers no longer ignore suppression rules that are not applicable to the type of sanitize being run. Instead it issues an parse error when reading the suppression file. I guess we will need to split up the suppression into separate files based on the type of suppression being used. Also as far as I can tell the leak sanitize is now separated from the address sanitize (now lsan and asan) so the address sanitize can't even parse leak:xxxx suppression statements. This might add some more complexity to the configure/make setup since some platforms only have asan which does both address and leak.

[ref2]
getting compiler errors in MdsScalarTest_NumericConversion.cpp


tfredian:
Builds succeeded after invoking them separately. Perhaps still a multi-build race condition exists.

PR #787, opened 2017-01-05
Prevent thread conflicts in mdsdcl

tfredian:
The routines that set and get global settings of the command prompt
and default file extensions were not thread safe. This change adds
thread locking to prevent race conditions in threads.


tfredian:
retest

PR #796, opened 2017-01-09
migration to alpha_release-7-1-83

zack-vii:


tfredian:
We will need to decide if we want to the promotion this way or base the promoted stable on the alpha branch thus inheriting the commit history of the alpha branch. Closing for now until we decide how best to do the promotion.

PR #797, opened 2017-01-10
Doxygen fix

AndreaRigoni:
- fix configure if python is missing (docker build_docs image does not have python installed)
- fix docs to be compiled if builddir is different from srcdir
- add docker selfhelp tags in "make help"


zack-vii:
@Andrea: try reinserting the python part. And pr your changes in two steps. So we can narrow down the problem. Alpha works at least. It passed all tests before

Andrea:
you mean @AndreaRigoni ???

zack-vii:
ah, yes.. oh, did github bind this to your account? sorry.

PR #807, opened 2017-01-19
Build: Remove santize and valgrind test options from os/*.opts

tfredian:
Instead of putting the test options in the source code they will
be added to the jenkins build jobs. This will enable us to change
how many and which build jobs will perform these tests without
making modifications to the source files.


tfredian:
This has been replaced by the addition of --sanitize --valgrind (without test lists) options.

PR #809, opened 2017-01-25
Build: Update configure to define new MDSPLUS_SEARCH_LIBS function

tfredian:
The first attempt to ensure the TARGET_ARCH compiler options are
applied when checking for libraries was implemented by redefining
AC_CHECK_LIB and AC_SEARCH_LIBS. This change performs the same action
but does so without modifying the existing functions. It also replaces
the use of AC_CHECK_LIB with AC_SEARCH_LIBS which is the recommended
approach. The overall effect is to ensure that the checks used to
check for the existance of libraries do so using the architecture (32-bit
or 64-bit) the build is being performed for.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -70,24 +70,12 @@ AC_MSG_RESULT([**********VISIBILITY*********$CFLAG_VISIBILITY*******************
 #if test -n "$CFLAG_VISIBILITY" && test "$is_w32" != yes; then
 #	CFLAGS="$CFLAGS $CFLAG_VISIBILITY"
 #fi
-m4_pattern_allow([AC_CHECK_LIB])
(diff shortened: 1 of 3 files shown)


AndreaRigoni:
Hi Tom,
I wonder if there is an example where the changes were not functioning .. My concern is that this way probably also the AC_LINK_IFELSE and AC_COMPILE_IFELSE (and maybe other functions too) should be changed as well .. and if an imported m4 macro uses instead the standard autoconf functions it may fail to work again.

Another approach would be to replace in the way you did the last two, that seem to be more basic functions as they appear in the implementation of AC_CHECK_LIB and AC_SEARCH_LIB  (see: /usr/share/autoconf/autoconf/libs.m4). for example the search for readline macro is also using the AC_TRY_RUN function that could be left without the proper TARGET_ARCH then.

PR #811, opened 2017-01-25
Build: Handle commit revert feature of github

tfredian:
If you revert the merge of a pull request, github adds a commit
with a title that begins with Revert "original title".
This change will parse that Revert title and treat it as the
same type of commit as the original which was reverted.


tfredian:
retest this please - this phrase should trigger a new pr test build

Issue #680, opened 2016-10-25
missing 'make distclean'


zack-vii:
When running deploy/build.sh after executing ./configure . The build terminates suggesting to run 'make distclean'. However, there is no setup for 'make distclean' and 'make clean' does not fulfill the job.
It not urgent or anything but a nice to have.

```
configure: error: source directory already configured; run "make distclean" there first
```

tfredian:
I agree the clean options in the top level Makefile needs a lot of work. There is a full_clean which kind of works but not completely and the diskclean does not exist while the aclocal.m4 suggests doing make distclean.

I normally don't run the build.sh script inside the source tree which makes things a little easier to deal with (or even configure for that matter). I use a different directory outside of the sources and do ${srcdir}/deploy/build.sh in that build directly (or ${srcdir}/configure). At this point the srcdir remains pristine and you can just rm all the files in the build directory tree when done.

tfredian:
make distclean added

PR #813, opened 2017-01-26
Build: Add distclean target to top Makefile

tfredian:
This commit adds the distclean target to the top directory Makefile.
The command "make distclean" should clean the directory tree of any
files or directory not in the original source tree. If the sources
where obtained via git it will use the "git clean" command to remove
the files otherwise it will use "make fullclean" if the "make distclean"
command was issued from the top source directory. It is now recommended
that configure and make be performed from a build directory outside
of the source tree which should leave the source directory tree in
a clean state.

This commit is to address git issue:

https://github.com/MDSplus/mdsplus/issues/680

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -31,7 +31,7 @@ HDF5_APS = hdf5
 D3D_PACKAGE = d3dshr
 
 PARTS = \
-        mdsshr \
+mdsshr \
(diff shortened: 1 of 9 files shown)


tfredian:
retest this please

PR #820, opened 2017-02-03
Build: Support for alpine linux support

tfredian:
This contains a few patches needed to be able to compile MDSplus on the alpine linux
platform.


tfredian:
retest this please

PR #824, opened 2017-02-07
Fix: Use different signing keys for ubuntu14 than ubuntu16

tfredian:
For some reason Ubuntu14 is unable to sign the debian packages with the
new secure RSA keys now used to sign Ubuntu16 packages. This commit enables the ability
to create signing keys in a ${OS} subdirectory of the signing keys. If the directory is
found it will set HOME to that subdirectory for signing packages. If the directory is not
found it will set HOME to the signing keys directory to use the default keys. A ubuntu16
subdirectory with the secure RSA keys was created and the other repositories use the origina
DSA keys.


tfredian:
retest this please

PR #833, opened 2017-02-16
Fix: improved Exception handling of Python code in Tdi

zack-vii:
* If python code excepts with an MDSplusException, the status is used as STATUS code.
* fixed use after free issue in TdiExtPython (filename)
* PyDoMethod.py does not need to return a List object anymore and can simply raise Exceptions.
* cleaned up code using status.h
* included related tests to treeUnitTest.py
* fixed ErrorCode on dcl_dispatch if server is down

### DIFF
--- a/actions/actmon.c
+++ b/actions/actmon.c
@@ -389,55 +389,43 @@ static void ParseTime(LinkedEvent * event)
     event->time = strtok(0, " ");
 }
 
-static int ParseMsg(char *msg, LinkedEvent * event)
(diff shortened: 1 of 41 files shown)


zack-vii:
wait: I want do add some tests to python/tests first

zack-vii:
I tested it manually but due to issue #834 the Jenkins result cannot be trusted.

zack-vii:
While I was testing the changes some other issues popped up, that have been fixed.

PR #836, opened 2017-02-17
Tests: Fix python testing

tfredian:
The earlier fix to deal with the use of sanitize with python caused a problem
of environment variables not being properly set preventing the tests from being executed. This should fix that.


tfredian:
Did not work. Andrea will fix it on Monday.

PR #839, opened 2017-02-21
Fix: addressed some performance issues in python

zack-vii:
* minor speedup of TreeNode __new__ and __init__ 
* major speedup of getNodeWild (tree with 18000 nodes in 0.18 seconds compared to 19 second)


zack-vii:
rhel6 failure related to  bug in the trigger/main build script

tfredian:
retest this please

zack-vii:
This commit now includes the changes by @AndreaRigoni #837.
Further, it fixes the memory leak caused by tcl calls from python.

PR #860, opened 2017-03-24
Fix: cleaned out code for TCP und TCPv6

zack-vii:
*some differences might not be necessary
*SIGINT will stop process on first attempt


zack-vii:
alpine failed at first because the java compiler crashed again. second test succeeded.

PR #864, opened 2017-03-27
Fix: Use find tag instead of find node wild in create pulse for perfo…

tfredian:
…rmance

For some reason that I cannot recall or find I had changed TreeCreatePule to
use a wildcard call to TreeFindNodeWild to identify all the subtree nodes but under
certain conditions this can be very inefficient. (The TreeFindNodeWild code probably
needs revisited to improve performance!) This change reverts the code back to use
TreeFindTag using TOP to find subtrees.


tfredian:
retest this please

PR #869, opened 2017-03-30
Zck action monitor

zack-vii:


zack-vii:
TODO: actmon still does not display actions

PR #875, opened 2017-04-04
Zck gethost udt.h

zack-vii:
* protest datafiel and ncifile pointer 
* unified mdsip udt in one header file  (TCPv6 and UDTv6 need testing)


zack-vii:
tcp, udt, tcpv6, and udtv6 are working (udtv6 was not for a long time)
ssh seems unaffected, the others should be too.

port separator for v6 is '#' as it was for udtv6 before

PR #880, opened 2017-04-05
Fix: Better handling of text arrays of zero length srings

tfredian:
When an array of strings is created using, for example, ['',''],
an array descriptor is created with length=0 and arsize=0 but with aflags.coeff set and
dimct and m[0:dimct-1] filled in. This enables the determination of number of elements and
shape without using length and arsize. Many tdishr function still rely on length and arsize
so not all functions will work on these empty string arrays yet. Since the code prior to this
commit didn't work with empty string arrays this commit shouldn't break anything that was
working in ealier versions.


tfredian:
retest this please

PR #884, opened 2017-04-10
Fix: segfault on mdsconnect

zack-vii:


zack-vii:
ubuntu16 fail not related to pr. I manually restarted the test and it succeeded

PR #891, opened 2017-04-11
FIX: add library to jScope and jTraverser - fixes mac SIP problem

joshStillerman:
Novimir Pablant found that adding -Djava.library.path=$MDSPLUS_DIR/${lib}
to the jTraverser command allowed him to fire it up even with macintosh
SIP (System Integrity Protection) enabled.

tested on linux, works there also.


tfredian:
retest this please

PR #895, opened 2017-04-13
Feature: Add a magic submodule to the MDSplus python module

tfredian:
This enhancement enables the use of %tcl and %tdi magic lines
in an iPython or jupyter notebook. You will be able to intersperse
mdstcl commands and tdi expressions in an ipython notebook along
with python.


tfredian:
oops. forgot packaging changes. Will retry tomorrow.

joshStillerman:
Tom -
  Check that you can still import MDSplus from a python that is not IPython  and on a system that does not have IPython installed.

PR #916, opened 2017-05-02
Build: Add --promote_to=<branch> option on trigger script

tfredian:
This should enable us to add an option promote option on release builds
such as the jenkins stable release build job to optionally perform a promotion
of stable to the alpha head and creating a stable release with the same version
and content as the current alpha branch.


tfredian:
Retest this please

PR #919, opened 2017-05-17
Zck python ctx

zack-vii:
restores the tdi/tcl context properly
hierarchy:
1. ctx argument (e.g. to work on specific Tree() object)
2. local tree context _TreeCtx.gettctx() (e.g. to handle trees using the static tdi and dcl methods)
3. current database =switchDbid() (e.g. when called as pyfun from external program)
4. use empty db and init it as local if required (opened=True)


zack-vii:
retest this please

PR #920, opened 2017-05-17
Feature: added LIST, DICT, adn TUPLE to built in functions

zack-vii:
this is much faster than the python workaround and it breaks the 250 element limit.
_l=List(*,_a,_b,_c)   =>  [_a, _b, _c]
_ll=List(_l,_d,_e,_f)  => [_a, _b, _c, _d, _e, _f]

if required a tdi function could wrap around the old apdadd but I don't think apdadd have been used.


zack-vii:
retest this please

zack-vii:
merged as part of #927

PR #924, opened 2017-05-23
Fix: ensure TreeOpenEdit uses a null pointer for ctx if no ctx argume…

tfredian:
…nt supplied

The existing code passed the default context by reference to TreeOpenEdit which
ended up changing the actual default from a pointer to null to the ctx set by the
TreeOpenEdit call so the next time the python method was called it did not behave as
expected and passed a pointer to memory that had been freed. This fixes this problem.


tfredian:
retest this please

PR #931, opened 2017-05-26
Build: Perform tests only if open --test=yes

tfredian:
The test section of the build scripts was performing tests if either the
--test option or the --valgrind_tools option was specified.
The --valgrind_tools option is meant to be specified in either os.opts file
or in the build job for a particular os to indicate what valgrind tools should
be used if the build was to perform tests. This change will correct the logic
to only test if the --test option is specified.


tfredian:
retest this please

PR #934, opened 2017-05-29
Fix: added missing finalize() to java MDSplus.Tree

zack-vii:


zack-vii:
Hey @GabrieleManduchi, could you verify this against any interference with JTraverser or jScope. I think this should be the desired behavior (close tree of object gets destroyed)

zack-vii:
retest this please

tfredian:
retest this please

GabrieleManduchi:
Thanks for anticipating me....

PR #940, opened 2017-06-01
Fix: tcl> abort server would break locks if Worker holds any

zack-vii:


zack-vii:
merged in lite version in #941

PR #941, opened 2017-06-02
Zck abort server locks lite

zack-vii:
we should agree on a dockerimage to do the bootstrap
I used fc23 and fc25 (they produce the same output)
```bash
deploy/build.sh --os=fc23 --test -i
 cd /source
 ./bootstrap
```


zack-vii:
retest this please

PR #943, opened 2017-06-07
Feature: added javatraverser2

zack-vii:
need some car .. name eventual different script to call etc


zack-vii:
possibly need new java env for debian as Entry should be public but is not etc

zack-vii:
retest this please

zack-vii:
retest this please

Issue #949, opened 2017-06-15
[python] list returned from Tree.getNodeWild() is broken


amicitas:
There is a problem with the list returned from Tree.getNodeWild().  Using `len()` on the returned list results in:

    TypeError: Arrays of type DTYPE_L are unsupported. Error message was: 'float' object cannot be interpreted as an integer

This list is also not iterable, and any attempt to iterate produced the same error.  

Current testing with alpha-7.9.5 on Debian Linux using python 3.6.
As far as I can tell this problem does not exist in stable-7.1.14.

Example code on w7x:

    import MDSplus
    tree =  MDSplus.Tree('QSW_EVAL', 160309010)
    node_name = 'XICS:ANALYSIS:100MS_3CM:AR16:*:TI'
    node_list = tree.getNodeWild(node_name)
    num = len(node_list)

This will produce the error given above.

zack-vii:
hm I have problems reproducing this on 2.7 and 3.4 (I don't have 3.6 atm) and the result is:
node_list: [129] and len(node_list): 1
 without raising any Exception. So it might either be a python 3.6 problem or it has been fixed in the latest release The latter however is unlikely since we did not change anything python related to my knowledge. Could you run "MDSplus.\_\_version\_\_" to verify the version of the MDSplus python package.

amicitas:
Here is the version information:

    In [3]: MDSplus.__version__
    Out[3]: '7.9.5'

I did a little further testing and I found that there problem does not exist in Python 3.4.4, but does exist in python 3.6.1.

amicitas:
I have tried using the python source from the `alpha` branch of github.  Same issue on Python 3.6.1.

amicitas:
I found the problem in:

    _descriptor.py Line:466

Integer division (//) or explicit conversion to int is required:

    shape=[int(descr.arsize/descr.length) if descr.length != 0 else 0,] 

I'll to see if I can figure out how to make a merge request.

amicitas:
Pull request created: https://github.com/MDSplus/mdsplus/pull/951

PR #951, opened 2017-06-16
Fixes issue #949 caused by floating point division for int value

amicitas:
Fixes [Issue #949](https://github.com/MDSplus/mdsplus/issues/949).  This is the only place in _descriptor.py that does not already have the required explicit conversion to int.


tfredian:

tfredian:
test this please

tfredian:
Thanks for discovering this problem and providing a fix!

amicitas:
Happy to help.

PR #956, opened 2017-06-23
Revert "Twf fix python deadlock"

tfredian:
Reverts MDSplus/mdsplus#955


tfredian:
New python api has problems with certain data types. Investigating now but this will revert the alpha branch to before new python api until these problems are solved.

PR #959, opened 2017-06-28
Tests: Added writeSegments test

zack-vii:


zack-vii:
seems like the old api did not accept lists as input

tfredian:
No it does not accept lists and neither should the new api. Segments are arrays. Lists are implemented with apd's and cannot be stored in a segment.

zack-vii:
i meant python lists like [[1,2],[3,4]] the new api simply does an input = Data(input) and lets Data handle the the conversation. this way python types are only converted at one place. Data turns the python list into via numpy.array() into either int32 or float64/32 so it can be accepted by MDSplus Arrays

Issue #978, opened 2017-07-22
remote signal put fails python 3.6 and windows 7


cameronsamuell:
Hi, 

I had a problem whereby putting a signal to a local tree was successful but putting it to a remote tree using the python object orientated routines fails. I was able to put a numeric to both local and remote trees just fine and the trees are near-identical which points towards it being signal-specific.

Here is what i would be trying to do:
node = tree.getNode('PROCESSED:ACQUIRING')
acq_sig = mds.Signal(mds.makeArray(acq_volts), 'V', mds.makeArray(acq_times), 'ms')
node.putData(acq_sig)

It would fail with the following: 
OSError: exception: access violation reading oxFFFFFFFFFFFFFFFF

We were able to do a work around by building the signal and a thin client (i.e. a mds.Connection() object)

I don't know if this is relevant or not, but it mirrors a similar problem I had with Labview: local signals were fine but those to a remote server were not.

tfredian:
Can you tell us which release version of MDSplus is installed on your Windows system? If it is a relatively recent version you can use the "show version" command in mdstcl. If you have not updated MDSplus recently you might try doing so to see if the problem still persists. If you continue to have the problem on recent releases of MDSplus we will try to replicated the problem and hopefully provide a fix.

It shouldn't be relevant to the access violation problem but assuming that you are trying to associate units to the value and timebase of the signal you should do the following:

y=mds.makeArray(acq_volts)
y.units='V'
x=mds.makeArray(acq_times)
x.units='ms'
acq_sig=mds.Signal(y,None,x)
The Signal constructor arguments are (value, raw, dim0[,dim1,...dimn). In python units are assigned to MDSplus data items using item.units=units-string.

tfredian:
I tested this with the latest alpha release of MDSplus and the latest anaconda3 and it seems to work. However I ran into a problem with the latest stable release of MDSplus. That release at least on Windows is having trouble returning floating point arrays. It does not get an access violation but complains about the datatype being returned when retrieving double or single floating point arrays. I will start another issue for this new problem and will try to track it down. In the mean time I would recommend using the latest MDSplus alpha release. To install the MDSplus module in anaconda, the best approach is to do:

pip install [--user] "C:\Program Files\MDSplus\mdsobjects\python"

This ensures that the MDSplus python module version matches that of the MDSplus libraries installed on your system.

cameronsamuell:
Thanks for taking a look. 

[...]

PR #983, opened 2017-07-24
Fix: MdsValue termination argument must be Null pointer

tfredian:
MdsValue was called in several places with a terminating 32-bit integer 0
instead of a NULL pointer causing unpredictable behavior.

This should fix issue https://github.com/MDSplus/mdsplus/issues/978


tfredian:
retest this please

tfredian:
retest this please

tfredian:
retest this please

PR #987, opened 2017-07-28
BUILD: Redo mac packging using the Packages utility

joshStillerman:
As of 10.12 Apple no longer supports the packgaing utility we were using. This change uses packages to build a pkg for MDSplus  Packages is another requirement for building MDSplus:  http://s.sudre.free.fr/Software/Packages/about.html


tfredian:
This needs more work. I'm not sure about the name of the postinstall script file. The proj file used rtfd but the filename in the PR ended in rtf. The mdsplus/macosx/Makefile.am will need to be changed since you are installing different scripts for this new installer and after changing that you will need to:

 cd mdsplus
 ./bootstrap

and commit the new Makefile.am and Makefile.in files in macosx.

tfredian:
retest this please

PR #989, opened 2017-08-02
Zck python ctx DO_NOT_MERGE_YET

zack-vii:
not deprecation warning yet.
e.g.:
Instead of 
Data.execute(expression_with_node_refs)
one should use
tree_instance.tdi(expression_with_node_refs)

### DIFF
--- a/deploy/commit_type_check.sh
+++ b/deploy/commit_type_check.sh
@@ -4,7 +4,7 @@
 #
 # The original commit is passed in as an argument ("origin" for pull requests and
 # the last release tag for release builds.
-# 
+#
 # The second argument to this script should be the filespec of the contents of an
 # email to be sent to the person who created a commit with an invalid title.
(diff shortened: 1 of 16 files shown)


zack-vii:
TODO: TreeCtx does not properly clean up once the Tree is closed. this caused Windows test to fail as the trees are not closed when they should have been.
TODO2: fishy value: -1for size in _TreeGetRecord->_TreeGetSegmentedRecord->getOpaque_List->TreeGetDsc->malloc(length)
possibly since #1068 in treeshr/TreeSegments.c,359

zack-vii:
retest this please

tfredian (changes requested):
compound.py has a try: arg.setTree() except print(arg). Was this for debugging? If so, it should probably be changed.

tfredian (approved):

Issue #995, opened 2017-08-04
jTraverser "no JavaMds in java.library.path" problem


wbb12:
#I have installed MDSplu in the Ubuntu 16.04 LTS , but when I open the jTraverser in the terminal and try to open and edit a tree I have just created on this computer, the  jTraverser has no response and the terminal shows a info ‘no JavaMds in java.library.path’. The detailed info from the terminal is listed bellow. 

I have install the openjdk-8 successfully.  I want to know how to fix the problem and edit the tree by  jTraverser.

can any one has ever  encountered this problem and know how to fix it? Thank you!
Exception in thread "AWT-EventQueue-0" java.lang.UnsatisfiedLinkError: no JavaMds in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1122)
	at Database.<clinit>(Database.java:16)
	at Tree.open(Tree.java:357)
	at Tree.open_ok(Tree.java:326)
	at Tree.actionPerformed(Tree.java:1063)
	at javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:2022)
	at javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2348)
	at javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:402)
	at javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:259)
	at javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:252)
	at java.awt.Component.processMouseEvent(Component.java:6533)
	at javax.swing.JComponent.processMouseEvent(JComponent.java:3324)
	at java.awt.Component.processEvent(Component.java:6298)
	at java.awt.Container.processEvent(Container.java:2236)
	at java.awt.Component.dispatchEventImpl(Component.java:4889)
	at java.awt.Container.dispatchEventImpl(Container.java:2294)
	at java.awt.Component.dispatchEvent(Component.java:4711)
	at java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4888)
	at java.awt.LightweightDispatcher.processMouseEvent(Container.java:4525)
	at java.awt.LightweightDispatcher.dispatchEvent(Container.java:4466)
	at java.awt.Container.dispatchEventImpl(Container.java:2280)
[...]

PR #999, opened 2017-08-07
Zck j traverser2.bat

### LINKED ISSUE #997: traverser.bat problem in Alpha 7.15-5 Win10 64-bit
Dear all,

after setup the traverser.bat file contains the following single line
```
start "" javaw -jar "%MDSPLUS_DIR%\Java\Classes\jTraverser.jar" %*
```
which doesn't work.

I solve the problem by modifying it to read:
```
start javaw -cp "%MDSPLUS_DIR%\Java\Classes\jTraverser.jar" jTraverser
```
although I am not sure if the `%*` is something meant to be passed as command line argument by some applications. In my application (making a shortcut to start jTraverser) the `%*` is annoying, because I can't input the tree name with the mouse.

Apart of this, the right string should be ok from old commit  [fd39554] (sept. 2015), but it is not installed correctly at the moment.

Sincerely,

zack-vii:
this should fix #997


zack-vii:
retest this please

zack-vii:
could eventually also fix #995

PR #1002, opened 2017-08-11
Build: Add --test_timeunit option to modify test timeouts

tfredian:
The --test_timeunit option for the build.sh script can be used to adjust
the normal timeout duration used for tests. This option specifies a floating point
multiplier for the normal timeout. For example: --test_timeunit=.5 would reduce
the timeout on tests to half their normal time.


tfredian:
retest this please

PR #1004, opened 2017-08-11
Zck device help

zack-vii:
added help and fullhelp_str and fullhelp() to devicenode which will print/return the help string or print some useful information of the device node. Will be useful to display in traverser


zack-vii:
also fixes the .head field propagation in the device

zack-vii:
retest this please

PR #1010, opened 2017-08-16
Zck dimless segments

zack-vii:
does not fix the issue yet


zack-vii:
its a major clean up of xtreeshr but it enables the read of segments that have been stored with NULL dimensions.
It only uses start, end, and delta to select teh segments of interest but does not trim the first or last segment lacking the dimension. Instead it will return the concatenated array of all segments that cover the range start:end.

zack-vii:
@GabrieleManduchi : could you please check this PR. Thanks.

GabrieleManduchi (approved):

PR #1015, opened 2017-08-25
Fix: Repair handling of Opaque objects in segments

tfredian:
Some of the efficiency improvements of selecting segments based on
start, end and interval which were made quite a while ago had broken
the ability to store and retrieve Opaque objects as segments. Essentially
it is possible to use segments as a List allowing the list to grow by
adding more segments.

Tests were added to the segmentUnitTest for storing and retrieving
images stored as Opaque objects.


tfredian:
retest this please

zack-vii:
merged with #1017

PR #1021, opened 2017-09-01
Fix jscope streaming

GabrieleManduchi:
Added a feature requested by several users, i.e. showing a sequence of images stored as gif or jpg where every image is stored as a single segment in a segmented node


tfredian:
retest this please

PR #1026, opened 2017-09-13
Feature: Add username, date, and connection info to mdsip-ssh.log

joshStillerman:
When a user connects to a server using the ssh protocol, the server
runs mdsip-ssh-server and directs stderr to ~/mdsip-ssh.log.

This change adds information about the connection to the file. There
does not appear to be a sample server script for windows, only a client
script, so this only applies to linux like platforms.

Incomming ssh connections to windows servers have not been tested or
implemented.


tfredian:
retest this please

PR #1031, opened 2017-09-18
Feature: Reduce library exports to improve image activation performance

tfredian:
The MDSplus libraries contain numerous routines which are only used locally in that library. Currently all entry points are being exported by default and a large percentage of these symbols are not used externally and can be removed from the export list. This PR will only export symbols that are needed by other libraries, executables or those referenced by languages such as idl, python, java, matlab or tdi. Reducing the number of exports in the libraries should make program startup and dynamic lookup of routines much more efficient.


tfredian:
retest this please

tfredian:
retest this please

tfredian:
retest this please

joshStillerman (approved):

PR #1034, opened 2017-09-19
FIX: removed repetition in signal name for legend

GabrieleManduchi:
Removed repetition in jScope legend


tfredian:
This is an automated comment because the author of the PR is not a member of the MDSplus core development team. Can one of the core team members verify this patch and add a comment containing "ok-to-test" (without the hyphens) to the PR if it appears legitimate.

tfredian:
ok to test

PR #1045, opened 2017-09-26
Fix: Add MIT license to source files

joshStillerman:
As discussed in developer's meeting, adding the MIT license to
most source files.

We should look into adding License / copywrite information to
documentation files also.


tfredian (approved):
I looked in detail at all 859 files and did not find anything amiss ;-) What's the worse that can happen? Approved.

PR #1047, opened 2017-09-27
Gm fix connection timeout

GabrieleManduchi:
Fix timeout parameter in GetAnswerInfoTS(). The argument was added but not reported outside mdstcpip. Therefore a random value was passed as timeout in java and C++ Connection objects. This is likely the cause for random failures in Connection tests during build.

A timeout in server side streaming management (used by jScope MdsStreamingDataProvider) has been also added to  avoid ghost mdsip processes hanging on a condition that will never be set.


tfredian (changes requested):
There are many more calls to GetAnswerInfoTS which also need repaired. I will try to track them down and fix them and push another commit to this branch.

AndreaRigoni (approved):
Ok for me !

tfredian (approved):

PR #1051, opened 2017-09-28
FIX: Unfreeze waveform in all same x scale

GabrieleManduchi:
When selecting all sema X scale Auto Y the waveforms were wrongly freezed


tfredian (approved):
Seems ok to me.

PR #1053, opened 2017-09-29
Gm epics new archiver

GabrieleManduchi:
Improved version of EPICS MDSplus channel archiver. 
Does not affect ant other file except epics/archiver/ChannelArchiver.java


AndreaRigoni (approved):
Ok

PR #1054, opened 2017-09-29
Zck fix mdsip timeout

zack-vii:
* Added recv_to which uses poll instead of the io_settimeout implementation.
* Attempt to fix the issue addressed in #1047 by adding GetAnswerInfoTO instead of replacing GetAnswerInfoTS to avoid conflicts. therefore reverts some changes in #1047


zack-vii:
@GabrieleManduchi : could you check if i reverted all changes you made to meet the extra argument of GetAnswerInfoTS .. is there any in java/LV that i missed. it should be the usual GetAnswerInfoTS for without timeout and GetAnswerInfoTO if timeout is desired. timeout is now in milliseconds.

tfredian (changes requested):
I think this should be closed because https://github.com/MDSplus/mdsplus/pull/1056 supposedly includes these changes and was merged.

PR #1056, opened 2017-09-30
Zck fix connection

zack-vii:
when connecting to an action_monitor mdsip using actmon on an bash for windows session the mdsip process hosted by a debian system locked up due to an circular relation (c == c->next) of the ConnectionList. This could only happen during new and disconnect. Could not check yet if this solves the problem but the adjustments help to further protects the list. eventually the io->disconnect need to be protected via reentrant mutex but not sure. I have a reproducible setup in my office. i will test it once i am back.

includes #1054


zack-vii:
actually the field next is accessible through Connection so it could be any method actually

tfredian (approved):
This seems ok. Should 1054 be closed?

tfredian:
retest this please

PR #1058, opened 2017-10-01
FIX: Organize imports

kgerickson:
Remove all unused variables and organize imports based on default eclipse settings


GabrieleManduchi (approved):
Ok for me
I remind your pending review on TreeNodeThonClient

Issue #1032, opened 2017-09-18
dim_of() after settimecontext() on segmented data returns strange offset


astechow:
steps to reproduce in TDI:
```
> opentree() for valid shot with segmented data
> _d = \VALID.NODE
> dim_of(_d)
Set_Range(2399997,0Q ...
> settimecontext(0QU,10000QU) my data is stored in nanoseconds
> _d = \VALID.NODE
> dim_of(_d)
[18446657673.70955D0,18446657673.70955D0, ...
```

the data seems to be returned in seconds now instead, with all the intervals still correct but this huge offset on top of everything.

GabrieleManduchi:
I tried to reproduce the issue but everything appears to work as expected. Are you sure you are running the latest stable version? There was a similar issue about one year ago, but it had been fixed.
In case, could you provide me a pulse file reproducing the issue?

zack-vii:
I am working on this issue part time. the issue occurs if the recorded timestamps start with negative values. i created a branch including a test the investigates this behavior but i could not fix it yet.
 zack-vii:zck_timecontext pr #1060

zack-vii:
should be fixed in version with PR #1060

zack-vii:
it still seems to have an offset of 86400 (i.e. second in one day) due to an incorrect setting of isFloat. This will be/is fixed though in PR #1083

PR #1060, opened 2017-10-02
Bug: TimeContext with negative timestamps (issue #1032)

zack-vii:
* this changes all Q into c types conversions into int_64 rather than uint_64 
* fixes TdiEq to properly compare Q and O types and should fix issue  #1032
* SetTImeContext did not work with nodes that contained negative timestamps. (c.f, segmentsUnitTest)


kgerickson:
1. Correct me if I'm wrong, but isn't this an ABI break across pretty much every library?
1. There are a lot of changes here that are unrelated to your very sparse commit message
1. Your commit message says nothing about what you are changing and why. It's tagged as a Bug, but doesn't mention the issue in the commit message.  Mentioning it in the conversation attached to the pull request doesn't help with looking through git log. 

EDIT: Didn't mean to close the pull request.. it's strange that the default submit action on a comment is to close the request instead of just adding a comment....

zack-vii:
if someone has some spare time, could you revise this. would be great if we could use it next week.

tfredian (changes requested):
I somewhat agree with Keith. There are a lot of seemingly unrelated changes which could be better described in the commit messages. Regarding ABI changes, I think the routines changed are unlikely to be called by user code.  It seems like they are only called in a couple modules, one in XTreeShr and one in MdsMiscShr. Perhaps these routines could be moved to a single library and not be exported at all. That could probably be done in another PR later if it seems worth doing.

zack-vii:
HM, well these changes in mdsshr etc. are all meant to make timestamps uniformly treated as int64_t rather than uin64_t. these would be Wconversion warnings otherwise if we would have turned it on. I was not judging if or if not these methods are used at all. but if they are used they should at least be consistent with the way timestamps are treated elsewhere. e.g. you would want to convert -1Q to -1D0 rather than this awful large number. the changes are not that big of a break actually. since the data size (64bit) remains the same and library call would work either way.

tfredian (dismissed):

kgerickson:
Whether or not it's correct is a separate issue from it being a clear ABI
break.  This doesn't have anything to do with how often the functions are
called, this has to do with claiming binary library compatibility when
you've clearly broken it.  This is aside from the obvious API change.  You
can't do this change without addressing the compatibility breakages.  This
is EXACTLY the kind of backwards-incompatibilities that we must avoid.

kgerickson:
Whether or not it's correct is a separate issue from it being a clear ABI break.  This doesn't have anything to do with how often the functions are called, this has to do with claiming binary library compatibility when you've clearly broken it.  This is aside from the obvious API change.  You can't do this change without addressing the compatibility breakages.  This is EXACTLY the kind of backwards-incompatibilities that we must avoid.

kgerickson:
Please revert this merge until there is proper handling of the API and ABI breakages that this introduced.

zack-vii:
@kgerickson: i resorted the changes and made more detailed commit messages. Again, I doubt that the changes in the ABI would break any existing code as it is unlikely that anyone would have used uint64 _t's full scale. However this introduces the possibility to store signed timestamps and use them with full support of mdsplus capabilities. treating timestamps uniformly throughout all mdsplus libraries makes perfect sense, don't you agree. at some point if we activate Wconversion warnings, we will have to address the ABI break anyway.

tfredian (approved):
looks ok to me

PR #1062, opened 2017-10-03
F4E-OFC-280

fmolon:


tfredian:
This is an automated comment because the author of the PR is not a member of the MDSplus core development team. Can one of the core team members verify this patch and add a comment containing "ok-to-test" (without the hyphens) to the PR if it appears legitimate.

GabrieleManduchi:
ok-to-test
The changes in the code address only files in device_support. They will be eventually merged in alpha once the tests on the involved hardware have been carried out

GabrieleManduchi:
ok to test

PR #1063, opened 2017-10-03
FIX: GetManyExecute

GabrieleManduchi:
Routime GetManyExecute of MdsObjectsCppShr must be globally visible because it is accessed from the Java MdsObjects interface, i.e. from library JavaMds.


tfredian (approved):
Looks good

PR #1066, opened 2017-10-04
Fix: fixed get and added tests; TODO: fix GetMany and PutMany

zack-vii:
* Get/PutMany if not tested and lost functionality at some point.. seems like GetManyExecute does not work or is not interfaced correctly   (serialized List/Dict/?)
* get was not tested and is now
* put should be tested at some point


zack-vii:
retest this please

tfredian (approved):
Looks good to me

PR #1068, opened 2017-10-05
Feature: Provide WRITE_ONCE enforcement with segmented signals

tfredian:
If a node is set to be WRITE_ONCE (ie. TCL> SET NODE sig/WRITE_ONCE)
you will not be permitted to add or update any segments unless the
node is empty at which point only one segment can be written before
writes are blocked.

This also fixes the length characteristic of segmented nodes. The
length characteristic is a 32-bit value so the length value will
peak around 4 billion even if the total length of all the segments exceeds
this amount.


tfredian:
It turns out I broke the length value five years ago when I added the brilliant feature of having the length peak at 2^31. The not so brilliant part was I coded it in C as 2^31 and since the carat is an XOR operator the length max'ed out at a whopping 29!

zack-vii:
also you might need to change the datatype of add_length and length to uint32_t as well, since they are only int atm. or you need to stick with int. 
is the max value not rather (1<<32)-1 for uint32 and  (1<<31)-1 for int (0xffffffff or 0x7fffffff, respectively)

tfredian:
The length/rlength values seem to work better now.

>TCL> dir/full sig
>
>\TWF::TOP
>
> :SIG         
>      Status: on,parent is on, usage signal
>      compress on put
>      Data inserted:  6-OCT-2017 09:29:33.98    Owner: gid=1000(twf),uid=1000(twf)
>      Dtype: DTYPE_FS              Class: CLASS_R             Length: 4294967295 bytes
>
>Total of 1 node.

zack-vii (approved):
looks good

PR #1069, opened 2017-10-06
Fix: use default recv if recv_to is not available

zack-vii:
silly did not take other protocols into account

this should also add timeout for the tunnel but it is currently not used. maybe it can be used via python Connection. i will see if this can be included in tests somewhere


tfredian (approved):
Seems good to me.

PR #1071, opened 2017-10-09
Fix: Avoid deepcopy when fixing tree references in Compound objects

tfredian:
The deepcopy function can nolonger be used to copy Compound instances as
it now may contain ctypes objects. This change avoids the use of deepcopy.


tfredian:
Closing this PR for now. Hopefully will come up with a better solution described in: https://github.com/MDSplus/mdsplus/issues/1072

PR #1079, opened 2017-10-10
Gm treenode thin client

GabrieleManduchi:
New Thin client TreeNode - removed unised variable warnings/errors


mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,360 @@ StringArray *TreeNodeArray::getUsage()
...
+      case NciCONGLOMERATE_NIDS:	return "\'CONGLOMERATE_NIDS\'";
+      case NciORIGINAL_PART_NAME:	return "\'ORIGINAL_PART_NAME\'";
+      case NciNUMBER_OF_MEMBERS:	return "\'NUMBER_OF_MEMBERS\'";
+      case NciNUMBER_OF_CHILDREN :	return "\'NUMBER_OF_CHILDREN\'";
+      case NciMEMBER_NIDS:		return "\'MEMBER_NIDS\'";
+      case NciCHILDREN_NIDS:		return "\'CHILDREN_NIDS\'";
+      case NciFULLPATH:			return "\'FULLPATH\'";
+      case NciMINPATH:			return "\'MINPATH\'";
+      case NciUSAGE:			return "\'USAGE\'";
+      default:				std::cout << "Unecpected NCI item in TreeNodeThinClient" << std::endl; return "\'\'";
```
kgerickson: typo: unexpected. Also, should the library be doing std::cout, or throwing an exception?

GabrieleManduchi:
Reported Keith's observations. 
The only exception is  getnci stuff originally defined as template and used directly, and now used through four added methods getNciString(), getNciString(), getNciChar(), getNciInt(), getNciInt64(). 

I have been forced doing this way because the derived TreeNodeThinClient redefines getNcixxx() so that it can reuse a lot of TreeNode code. Unfortunately it is not possible having a virtual method that is also a template, i.e. making the original template getnci virtual so that it could be redefined by the subclass. So the only way was to make it a bit less elegant in TreeNode and providing four  virtual methods for getNci.

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+std::string  TreeNodeThinClient::getNciString(int itm)
+{
+    char expr[64];
+    sprintf(expr, "GETNCI(%d,%s)", nid, convertNciItm(itm));
+    Data *retStringData = connection->get(expr);
+    if(!retStringData)
+	throw MdsException("Error in Remote evaluation of getnci");
+    
+    AutoString as(retStringData->getString());
+//    char *retStringPtr = retStringData->getString();
```
kgerickson: Remove commented lines

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+
+std::string  TreeNodeThinClient::getNciString(int itm)
+{
+    char expr[64];
+    sprintf(expr, "GETNCI(%d,%s)", nid, convertNciItm(itm));
+    Data *retStringData = connection->get(expr);
+    if(!retStringData)
+	throw MdsException("Error in Remote evaluation of getnci");
+    
+    AutoString as(retStringData->getString());
```
kgerickson: The point was to use this in all cases, not just one

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+    clazz = CLASS_S;
+    dtype = DTYPE_NID;
+    setAccessory(units, error, help, validation);
+}
+
+std::string  TreeNodeThinClient::getNciString(int itm)
+{
+    char expr[64];
+    sprintf(expr, "GETNCI(%d,%s)", nid, convertNciItm(itm));
+    Data *retStringData = connection->get(expr);
```
kgerickson: You can use AutoData for Data pointers, too...

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+    
+    AutoString as(retStringData->getString());
+//    char *retStringPtr = retStringData->getString();
+//    std::string retString(retStringPtr);
+//    delete [] retStringPtr;
+    MDSplus::deleteData(retStringData);
+    return as.string;
+}
+
+char TreeNodeThinClient::getNciChar(int itm)
```
kgerickson: Complete example:

char TreeNodeThinClient::getNciChar(int itm)
{
    char expr[64];
    sprintf(expr, "GETNCI(%d,%s)", nid, convertNciItm(itm));
    AutoData<Data> d(connection->get(expr));
    if (!d.get())
        throw MdsException("Error in Remote evaluation of getnci");
    return d->getByte();
}

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+EXPORT void TreeNodeThinClient::deleteData()
+{
+    char *path = getFullPath();
+    Data *args[1];
+    args[0] = 0;
+    connection->put(path, (char *)"*", args, 0);
+    delete [] path;
+}
+
+EXPORT bool TreeNodeThinClient::isOn()
```
kgerickson: Even in this case:

    AutoData<Data> d(connection->get(expr));
....etc.

    return d->getByte() == 0;  // This is really inverted logic?
}

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+
+EXPORT void TreeNodeThinClient::setOn(bool on)
+{
+    char expr[64];
+    if(on)
+      sprintf(expr, "TreeTurnOn(%d)", nid);
+    else
+      sprintf(expr, "TreeTurnOff(%d)", nid);
+    Data *retData = connection->get(expr);
+    if(retData)
```
kgerickson: and here, with AutoData, you can remove the if-check entirely

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,337 @@ StringArray *TreeNodeArray::getUsage()
...
+	MDSplus::deleteData(retData);
+}
+
+EXPORT void TreeNodeThinClient::beginSegment(Data *start, Data *end, Data *time, Array *initialData)
+{
+    char expr[256];
+    Data *args[4];
```
kgerickson: I believe I had a comment to do this inline:

Data * args[4] = {
start->data(),
 end->data(),
 time->data(),
 initialData->data()
};

This gets you basic compile checking of array bounds

And of course, making it be an array of AutoData's means that you can remove the for loop at the end (please do this)

kgerickson (changes requested):
Basically, you only changed one instance of using an automatically deleted pointer. Change them all.  I highlighted a few, but do all of them, and be sure to use AutoData where possible for Data pointers.

GabrieleManduchi:
Ok, used AutoData whenever possible now.
As soon as I can proceed I will report this to LabVIEW as requested by Greg

joshStillerman:
retest this please

mdsobjects/cpp/mdsipobjects.cpp
```diff
@@ -486,6 +490,18 @@ void Connection::setDefault(char *path)
 		throw MdsException(status);
 }
 
+TreeNodeThinClient *Connection::getNode(char *path)
+{
+	char expr[256];
+	sprintf(expr, "GETNCI(%s, \'NID_NUMBER\')", path);
+	Data *nidData = get(expr);
```
kgerickson: Still missing AutoData

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1681,307 @@ StringArray *TreeNodeArray::getUsage()
...
+    Data *retData = connection->get(expr);
+    if(!retData)
+	throw MdsException("Error in Remote evaluation of getnci(record)");
+    return retData;
+}
+  
+EXPORT void TreeNodeThinClient::putData(Data *data)
+{
+    const char *path = (const char *)getPath();
+    Data *args[1];
```
kgerickson: Still missing AutoData on these arrays of args (and something on the path on the line above).  Since get and put can throw, this will leak.  If passing the array of args is problematic, then you can use aggregation instead of composition to separate ownership:

AutoData<Data> d[] = { data->data() };
Data * args[] = { d[0].get() };

Otherwise, you need lots of try/catches, deletes, the delete in the for loop, etc.  Use C++ here to simplify.

kgerickson:
I added two more comments.  You missed two Auto's in general, and you avoided all the ones that involved an array of Data*. I explained in the comment how to handle the array using two parallel arrays.

GabrieleManduchi:
Now I think that all deleteData() stuff has gone away.

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1674,288 @@ StringArray *TreeNodeArray::getUsage()
...
+}
+  
+EXPORT void TreeNodeThinClient::putData(Data *data)
+{
+    const char *path = (const char *)getPath();
+    AutoData<Data> argsD[] = {data->data()};
+    Data *args[] = {argsD[0].get()};
+    connection->put(path, (char *)"$", args, 1);
+    delete [] path;
+
```
kgerickson: Remove blank line

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1674,288 @@ StringArray *TreeNodeArray::getUsage()
...
+    sprintf(expr, "GETNCI(%d,\'RECORD\')", nid);
+    Data *retData = connection->get(expr);
+    if(!retData)
+	throw MdsException("Error in Remote evaluation of getnci(record)");
+    return retData;
+}
+  
+EXPORT void TreeNodeThinClient::putData(Data *data)
+{
+    const char *path = (const char *)getPath();
```
kgerickson: Use "AutoArray<char> path = getPath()" here, and remove the delete on line 1792

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1674,288 @@ StringArray *TreeNodeArray::getUsage()
...
+    AutoData<Data> argsD[] = {data->data()};
+    Data *args[] = {argsD[0].get()};
+    connection->put(path, (char *)"$", args, 1);
+    delete [] path;
+  
+}
+
+EXPORT void TreeNodeThinClient::deleteData()
+{
+    char *path = getFullPath();
```
kgerickson: Another AutoArray<char>

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -1661,7 +1674,288 @@ StringArray *TreeNodeArray::getUsage()
...
+    Data *args[1];
```
kgerickson: Replace with: Data * args[1] = {0}, remove following line of "args[0] = 0".  In C++, we always prefer initialization over assignment, unlike in Java.

kgerickson:
You still have a lot of " if(!retData)" instead of  "if(!retData.get())".  The former will always be false.

GabrieleManduchi:
Used AutoArray.

Not sure  that if(!retData) will always return false, because it must be converted to a number/boolean and therefore I believe get() is implicitly called.  Anyway, changed as suggested, it makes the code clearer.

GabrieleManduchi:
Keith, 
   if you approve this, unless you discover other issues, I can proceed in making the development asked by Greg

kgerickson (approved):

PR #1083, opened 2017-10-10
Zck setTimeContext does return double even if the original dim was an expression generating Q types

zack-vii:
* resolved inconsistent data conversions (uint <-> int) in XTreeDefaultResample
* if dim is an expression (e.g. "-10Q : 10Q : 1Q") the original code would not identify the datatype as int64 but default to float instead.


tfredian (commented):
looks good

tfredian (approved):

PR #1085, opened 2017-10-11
Zck random py fixes

zack-vii:
fixes python compare and Float{F,G,H,D}Arrays


tfredian (approved):
looks good

kgerickson:
We need much better titles than "random py fixes".  Having descriptive pull requests is the same as having descriptive log messages.  Without them, archeology becomes useless.

PR #1088, opened 2017-10-12
Fix: Opaque attempt to get dim for opaque segments

tfredian:
When storing opaque segments (list of jpegs etc) no dimension
information is stored for each entry. The ReadSegment code was
trying to obtain a dimension which resulted in a malloc(-1) call.
This fix prevents attempting to read the dimension if their is none.


tfredian:
retest this please

joshStillerman (approved):

PR #1089, opened 2017-10-12
Fix: Fix internal clock check in store

joshStillerman:
The check of the status register for external clock was after a shift
of the register in the wrong direction.


tfredian (approved):
looks good to me.

PR #1101, opened 2017-10-24
Fix: matlab uses only single quotes for strings

joshStillerman:
Replaced the double quoted strings with single quoted strings


tfredian (approved):

tfredian:
retest this please

Issue #1106, opened 2017-10-27
Tree context problems in XtreeShr evaluating expressions in segmented records


tfredian:
The global tree context is not properly handled when referencing segments in XTreeShr. When _TreeGetRecord is called in TreeShr on a segmented record the code calls into XTreeShr which determines which segments to combine in the answer. If the dimensions in the segments refer to other treenodes there is a tree context problem when calling from XTreeShr into TdiShr. Since TdiShr uses a global tree context any callers of TdiShr routines which have a tree context (dbid) passed to it must do the following:

lock the global context (dbid)
save the current global context
set the global context to the dbid passed to it
call tdishr routine
restore the original global context
unlock the global context

PR #1107, opened 2017-10-27
Fix: Workaround for issue #1106: no ctx in XTreeShr

zack-vii:


tfredian (approved):
This should mostly work but this does present a race condition since we don't currently have any locking on the global tree ctx. We'll look into a mechanism in the c code for locking the global ctx, push ctx, call tdishr routine, pop ctx, unlock global ctx and see if this needs to be done elsewhere in all the code.

zack-vii:
well considering the python code we do actually. not? pushTree acquires a lock and popTree releases it.
________________________________

PR #1113, opened 2017-11-07
Build: User real source directory inside docker images

tfredian:
The original docker build scripts used /source for the source volume.
Various development tools can be used better if the docker builds use
the same source directory spec as that outside of docker. For example
if the source lives in /home/twf/mdsplus-alpha and that same directory
is used as the internal volume inside the docker container instead of
mapping it to /source, all debug information etc will contain the correct
source file location if used outside of docker.

### DIFF
--- a/deploy/packaging/windows/mdsplus.nsi
+++ b/deploy/packaging/windows/mdsplus.nsi
@@ -78,8 +78,8 @@ Section
 SetOutPath "$INSTDIR"
 SetShellVarContext all
 File "/oname=ChangeLog.rtf" ChangeLog
-File /source/mdsplus.ico
-File /source/MDSplus-License.rtf
+File ${srcdir}/mdsplus.ico
+File ${srcdir}/MDSplus-License.rtf
 writeUninstaller "$INSTDIR\uninstall.exe"
 WriteRegStr HKLM "${ENVREG}" MDS_PATH "$INSTDIR\tdi"
 WriteRegStr HKLM "${ENVREG}" MDSPLUS_DIR "$INSTDIR"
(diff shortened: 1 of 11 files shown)


tfredian:
retest this please

AndreaRigoni (approved):
great !

zack-vii:
actually i would rather recomment an option to the build.sh that may overrider the /source default. this is more transparent an alows for mire freedom.

AndreaRigoni:
Hi Timo,
 it could be an option but this approach fits well with the same logic of the ./configure script for me.
You wont fire the build.sh outside the source directory for now so the source directory is an implicit information likewise the configure does.

zack-vii:
No i meant what if you are compiling for a different machine that does not have docker e.g. the source files might be lying elsewhere. i e.g have a link to the sources on each computer so the debugging works fine.
"deploy/build.sh --os=fc25 --test --source-dir=." could do want you want to do but with the same effort you would allow for much more flexibility.
"deploy/build.sh --os=fc25 --test --source-dir=/my/folder/on/target/machine"

deploy/packaging/windows/mdsplus.nsi
```diff
@@ -165,7 +165,7 @@ WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\MDSplus" "
 WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\MDSplus" "DisplayIcon" "INSTDIR\mdsplus.ico"
 WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\MDSplus" "Publisher" "MDSplus Collaboratory"
 WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\MDSplus" "HelpLink" "${HELPURL}"
-WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\MDSplus" "InstallSource" "http://www.mdsplus.org/dist/SOURCES/mdsplus${FLAVOR}-${MAJOR}.${MINOR}-${RELEASE}.tgz"
+WriteRegStr HKLM "Software\Microsoft\Windows\CurrentVersion\Uninstall\MDSplus" "InstallSource" "http://www.mdsplus.org/dist${SRCDIR}S/mdsplus${FLAVOR}-${MAJOR}.${MINOR}-${RELEASE}.tgz"
```
zack-vii, line 168: this cant be right

deploy/platform/platform_docker_build.sh
```diff
@@ -7,9 +7,10 @@
 # Build script from within a docker image
 #
 export HOME=/tmp/home
+srcdir=$(realpath $(dirname ${0})/../..)
```
zack-vii: could be an option passed by build.sh
if [ -z  $srcdir ]
then srcdir=/source
fi

zack-vii (changes requested):

AndreaRigoni:
Ah now I got your point .. that is because I'm biased by the fact that I don't use the build script very often as I prefer the direct compilation with --with-docker-image parameter

tfredian:
This pull request will be replaced by a new one.

PR #1122, opened 2017-11-09
Fix: Fix mdstcpip exit hander

tfredian:
The mdstcpip exit handler in connections seems to be getting into
a race condition. This should prevent that.


zack-vii (changes requested):
Since Windows seems to like this now it should not hurt. Strange though that the atexit is also called by exiting threads they should not close ALL connections. If they don't call it why the race I think you should use another mutex, i.e. ConnectionList

PR #1123, opened 2017-11-10
Gm fix xtreedefaultresample

GabrieleManduchi:
Fix possible negative array index in XTreeDefaultResample


zack-vii:
I rebased it to current alpha. as it seems in the end you just added these two lines.
It might be best to always start your branch form the newest alpha.

zack-vii (approved):
should be fine.

PR #1127, opened 2017-11-16
Fix: Restore MDSplus python exception superclasses

tfredian:
The MDSplus python exception superclasses were changed to begin
with an underscore to prevent them from being exported but it is
sometimes useful to catch a class of exceptions by using the superclass
name. Some existing applications were doing this and the change to
hide these superclasses broke these applications. This change will
restore their visibility making them useable in user applications.

### DIFF
--- a/mdsobjects/python/mdsExceptions.py
+++ b/mdsobjects/python/mdsExceptions.py
@@ -33,6 +33,7 @@
 ########################################################
 
 class MDSplusException(Exception):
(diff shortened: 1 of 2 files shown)


tfredian:
retest this please

tfredian:
retest this please

tfredian:
retest this please

joshStillerman (approved):

PR #1128, opened 2017-11-18
Zck fix localtime r

zack-vii:
fixes the race condition on tmval by using localtime_r

### DIFF
--- a/mdsshr/librtl.c
+++ b/mdsshr/librtl.c
@@ -86,7 +86,15 @@ typedef struct node {
 } LibTreeNode;
 
 #include <libroutines.h>
-
+#ifdef _WIN32
+static void localtime_r(time_t* time, struct tm* tval){
+  static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
+  pthread_mutex_lock(&lock);
+  struct tm* ttval = localtime(time);
+  tval = memcpy(ttval,tval,sizeof(struct tm));
+  pthread_mutex_unlock(&lock);
+}
+#endif
 STATIC_CONSTANT int64_t VMS_TIME_OFFSET = LONG_LONG_CONSTANT(0x7c95674beb4000);
 
 ///
--- a/mdsshr/librtl.c
+++ b/mdsshr/librtl.c
@@ -87,14 +87,32 @@ typedef struct node {
 
 #include <libroutines.h>
 #ifdef _WIN32
-static void localtime_r(time_t* time, struct tm* tval){
-  static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
-  pthread_mutex_lock(&lock);
-  struct tm* ttval = localtime(time);
-  tval = memcpy(ttval,tval,sizeof(struct tm));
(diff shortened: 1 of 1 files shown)


zack-vii:
retest this please

zack-vii:
i am out of battery, if someone can pick it up. windows is not blessed with localtime_r, we should just move the lines closer together and ifdef _WIN32 it between locks.

tfredian:
Couple questions....
1) With the new code if you store a value in the node does dir/full on that node show the correct time? There are some differences in localtime and localtime_r related to the daylight and timezone values which are used in the vmstime calculation and it is difficult to tell from the documentation whether the fact that localtime_r doesn't update daylight and timezone values if it will work as before.
2) I got some hits on localtime_r and mingw64 and it seems that localtime_r should be available when doing the windows build. Was the compile getting undefined refs before adding the #ifdef _WIN32 block of code?
3) Some of the recent build failures were caused by a problem with jenkins bringing all the slave executors online after jenkins was restarted. I suspect an update to jenkins may have changed how it verifies the host certificates and for some reason it was getting errors when ssh'ing to the slaves. They are all back online now.
4) This PR did seem to fix the race condition with the time struct so unless this behaves differently with the switch from localtime to localtime_r it can be merged.
5) There are still two race conditions left to fix which seem to be related to TdiDoTask.

zack-vii:
Well:
1. There is no indication why localtime_r should behave differently from localtime apart from the fact that it is storing the 'struct tm' output in a caller allocated memory instead of a static thread shared variable. it is a posix reentrant replacement for localtime.
2. it might be included in some other/newer version of the mingw64 but yes the build on windows failed because localtime_r was not defined.
3. i noticed them being down. thanks for getting them back online.
4. :)
5. I searched for other methods of that kind xyz_r and we are using them a few of them in any places. i will see to replace them with the threadsafe versions where necessary on monday. also i can have a look on the tdidotask. do you have the valgrind results.

tfredian:
I am concerned about the following:
'''
The  localtime() function converts the calendar time timep to broken-down time repâ€
       resentation, expressed relative to the user's  specified  timezone.   The  function
       acts  as if it called tzset(3) and sets the external variables tzname with informaâ€
       tion about the current timezone, timezone with the difference  between  Coordinated
       Universal  Time (UTC) and local standard time in seconds, and daylight to a nonzero
       value if daylight savings time rules apply during  some  part  of  the  year.   The
       return  value points to a statically allocated struct which might be overwritten by
       subsequent calls to any of the date and time functions.  The localtime_r() function
       does  the  same,  but  stores  the data in a user-supplied struct.  It need not set
       tzname, timezone, and daylight.
'''
and daylight and timezone are used in the calculation. I can test this to make sure it behaves the same on Monday before merging this PR.

The other race conditions can be found in the same valgrind info I sent which contained this time race condition. There were 3 race conditions listed. This was the first and the other two are TdiDoTask related.

zack-vii:
hm good point.. it probably needs the tzset() initialization.
i made the it into a get_tz_offset method

tfredian (approved):

PR #1131, opened 2017-11-20
Fix: race condition on task_xd cleanup when using Worker

zack-vii:


tfredian (approved):
Great! No more valgrind errors reported!

PR #1134, opened 2017-11-20
Gm fix reset scales

GabrieleManduchi:
Fixed broken reset scales
Optimized update queue management by removing older pending requests for the same signal


tfredian:
retest this please

tfredian (approved):
Looks ok to me but I'm no java expert.

PR #1136, opened 2017-11-20
Zck win32 pthread features

zack-vii:
enables some pthread dependent functionality on windows machines.
* TdiDoTask with timeout
* cleanup after cancelation
* etc


tfredian (approved):
Looks good.

PR #1138, opened 2017-11-20
Fix: Repair memory leak in mdstcpip on fc26

tfredian:
Always call freeaddrinfo if getaddrinfo is called.


zack-vii (dismissed):
reasonable :)

tfredian:
Did not fix leaks.

PR #1141, opened 2017-11-21
Zck valgrind fs26 32bit

zack-vii:
* removed GETHOST__ macros from pthread_port
* fixed fc26 valgrind on 32bit by suppressing third party issues (e.g. python and fork)
* improved cancel-ability by adding more FREE_ON_EXIT
* it also cleans up the MDSConnectionTest with predictable ports as tests are now running on isolated machines

### DIFF
--- a/conf/valgrind-python.supp
+++ b/conf/valgrind-python.supp
@@ -428,3 +428,11 @@
    ...
    fun:PyThread_start_new_thread
 }
+
+{
+   python 32bit problem in PyObject_Realloc
+   Memcheck:Leak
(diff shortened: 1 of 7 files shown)


zack-vii:
retest this please

tfredian (approved):
Looks much better. In theory the memory leak reported should not be happening which makes me wonder if there is a bug in gethostbyaddr_r on fc26. We could either add a valgrind suppression or possibly try switching to use getaddrinfo instead.

zack-vii:
i think suppression is the way to go will be a follow up

PR #1143, opened 2017-11-22
Fix: Repair debian installer scripts

tfredian:
Earlier consolidation of redhat and debian installer scripts caused
some problems with debian based platforms because the scripts were
using macros defined only by redhat based installers.


zack-vii (approved):
fingers crossed

PR #1144, opened 2017-11-22
Fix: Repare installer scripts

### LINKED ISSUE #1142: Fix install scripts for debian platforms
When the redhat and debian installer pre and post install scripts were combined they used the python_sitelib macro which is only available on redhat installers. The scripts need to be fixed to only use the macro on redhat platforms.

This was mentioned in https://github.com/MDSplus/mdsplus/issues/995

tfredian:
When consoladating the installer scripts for redhat and debian
packaging there were remaining references to macros that are only
defined on redhat. This change fixes this problem.

Fixes https://github.com/MDSplus/mdsplus/issues/1142


zack-vii (approved):
fingers crossed

PR #1147, opened 2017-11-23
User level installer for Windows

zack-vii:
# Project goals:
- cleanup windows installation
- enable UserMode installation


zack-vii:
On Jenkins, fc26 address sanitizer complains about ptrace:
==<pid>==LeakSanitizer has encountered a fatal error.
==<pid>==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==<pid>==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
I cannot reproduce this on my home docker though. ??

=================
fixed by adding --privileged to docker call

zack-vii:
Installer works just fine but the suppression does not

zack-vii:
fc26 takes more than 40 minutes to finish all valgrind and sanitizer tests on 64 and 32 bit. no valgrind issues have been reported in the last build and since the GLIB version on fc26 seems to be buggy anyway we should probably leave the valgrind to fc25 and take advantage of the compiler warnings added by sanitizer.

zack-vii:
retest this please

zack-vii:
RfxDevices MitDevices W7xDevices are missing and python's call results in MDSplusERROR

mayby python3.6 related when handling the string list

tfredian (approved):
Wow

Issue #1130, opened 2017-11-20
Remove obsolete #ifdef __VMS code sections


tfredian:
The current code base is not compatible with VMS and we no longer support the VMS platform so all code with #ifdef __VMS should be removed.

zack-vii:
addressed by #1137 
TODO: remove tdi code

PR #1149, opened 2017-11-27
Build: Remove VMS specific code

tfredian:
This should remove all cases in the code that is left over support
for the VMS platform. A couple dummy tdi functions,
vax() and vms(), have been retained to avoid breaking any  site specific
tdi functions or application that might still refer to them.

Fixes https://github.com/MDSplus/mdsplus/issues/1130


zack-vii (approved):
looks good

PR #1154, opened 2017-11-29
Build: Turn on --enable-werror for release, optimized, build

tfredian:
This change will enable -Werror for all builds including release
builds with -O3 compiler optmization. Some platforms are generating
unused result warning when certain functions are used and the return
value is not used. At some point we should investigate these warnings
and properly check the return values and take the appropriate action
if the return is not what is expected. For now, this change also
adds a --disable-wreturns configure option and the builds use it so that
these warnings will ignored.


tfredian:
Needs some work for latest windows build with new fc26 warnings. Updates to PR to follow.

tfredian:
retest this please

joshStillerman (approved):

PR #1161, opened 2017-12-08
Feature: Enable data(build_opaque(...))

tfredian:
This should enable the used of data() on opaque objects.


zack-vii (approved):
looks good

PR #1163, opened 2017-12-10
Zck mdsip lock

zack-vii:
This improves the IoRoutine business. however, it would break user plugins if the even exist. but we should encourage the users to update their plugins or even integrate them with mdsplus.

Also it tests IP6 and GSI, however, we might need to excluded the latter on some os' .


zack-vii:
alpine64 did not start its test at first. but otherwise the pr should be fine and one could do a "retest this please".

TODO: find all internal calls to mdsip_connection.h methods that are still using id and replace them with safer versions and FindConnectionWithLock.

* The locking mechanism is a condition per connection and a global lock.
* the state field of the connection signals usage
* The DisconnectConnection method removes the connection from the ConnectionList before destruction.
  the state CON_DISCONNECT will prevent other threads from using the connection again.
  This allows for a safe destruction of the connection object.
* NewConnection does not add the Connection to the ConnectionList before its authentication is successful.

* MDsConnectionTest tests local(Tunnel), and both versions of tcp and udt (v4, v6)

zack-vii:
* guide valgrind to use /usr/lib64/valgrind for VALGRIND_LIB as it contains x86 libs as well

tfredian:
retest this please

zack-vii:
retest this please

zack-vii:
retest this please

tfredian (approved):

PR #1168, opened 2017-12-14
Feature: Add new MdsLib threadsafe entry points

tfredian:
The following entry points have been added to MdsLib

MdsConnectionR
MdsDisconnectR
MdsValueR
MdsValue2R
MdsPutR
MdsPut2R
MdsOpenR
MdsCloseR
MdsSetDefaultR

These are basically the same as the routines without the trailing R except that
they no longer use a global connection id making them much more threadsafe. Multiple
threads could run in parallel using different connections without interference. Locking
was added recently to the use of connections so that multiple threads can transact using
the same connection but that doesn't preclude one thread from interfering with the tdi
and tree context being used on the server side if using the same connection.

The following C code demonstrates the difference of the R routines and the original API:

 /* Original API */
    int ansd;
    float answer;
    int zero=0;
    MdsConnect('myhost');
    ansd = descr(&dtype_float,&answer,&zero);
    status = MdsValue("42.",&ansd,&zero);
 /* New API */
    int ansd;
    float answer;
    int zero=0;
    int connection;
    connection = MdsConnect('myhost');
    ansd = descr(&dtype_float,&answer,&zero);
    status = MdsValueR(&connection,"42.",&ansd,&zero);


joshStillerman (dismissed):

joshStillerman (approved):

fluffynukeit:
Do these xyzR functions provide reentrant safety in the following scenarios?

- A multi-threaded process accessing local data concurrently in a single tree (i.e. without MdsConnectR call)
- A multi-threaded process accessing local data concurrently in multiple trees (i.e. without MdsConnectR call)
- Either of the above scenarios using multiple processes instead of multiple threads?

I am presuming that using the MdsConnectR call will enable all the above scenarios to work, but will introduce some small loopback and protocol overhead that can avoided for local data access.

joshStillerman:
Daniel,
  You raise some interesting questions.  We an definitely look at the thread safety of the routines you mention.  The win for the 1st suggesion and the 1st 1/2 of the last will be that the IOs could be overlapped with the computation (IO overhead, decompression, evaluation).  MDSplus takes out locks when doing IO.  They are tree/subtree specific which means that to get the big win, the multi requests would have to be from different trees or at least subtrees.  This is not usually the case for analysis tasks.  It is more true for data dispay tasks - ’Show me these ten  (or 64) interesting things about some instance of an experiment.’  Analysis tends to look a lot more like: ‘give the the 100 chunks of raw data associated with this diagnostic’. Still worth looking into.
-Josh

fluffynukeit:
On linux, at least, it looks to me like there there is the capability to lock a portion of the file and not the whole tree.  Am I extrapolating too much from this small section of code?

https://github.com/MDSplus/mdsplus/blob/alpha/treeshr/RemoteAccess.c#L1548

PR #1170, opened 2017-12-15
Fix: Repair numerous compiler warnings

tfredian:
This should fix several compiler warnings in an attempt to enable -Werror with -Wextra.

It also removes the need for -Wno-clobbered compiler option used on the routines involving threads.
It was tested using fc26 but other platforms may uncover more warnings.


tfredian:
It looks like some of the changes made to prevent the clobbered warnings may have caused problems found by sanitize address. More work needed.

PR #1172, opened 2017-12-15
Fix: Resolve compiler warnings

tfredian:
This should resolve several compiler warnings including some that were originally disabled with the -Wno-clobbered option.

### DIFF
--- a/conf/valgrind-mdsplus.supp
+++ b/conf/valgrind-mdsplus.supp
@@ -21,3 +21,9 @@
    fun:__nss_hosts_lookup2
    fun:gethostbyaddr_r@@GLIBC_2.2.5
 }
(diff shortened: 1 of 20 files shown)


tfredian:
retest this please

zack-vii (changes requested):
Why did you change this. This will not work. You only need to init tzset once. which is what pthread_once does. if helgrind still complains ten because somewhere else in the code there is another call to tzset. which wont be guarded by your lock. what one could do is storing the timezone and daylight in a private static variable an use these copies instead. but you would do this in a pthread_once as well which blocks all other threads until the first one finishes the target method.

zack-vii (approved):

PR #1173, opened 2017-12-18
Twf warnings

zack-vii:
before merging into alpha we should eventually drop the revert and the reverted commit


zack-vii:
if you run
```bash
gdb --args python /git/mdsplus/mdsobjects/python/tests/dclUnitTest.py
```
and put a break at tzset it only breaks once.
Using private statics is probably the most robust solution.

zack-vii:
pushed directly into #1173

PR #1177, opened 2017-12-19
Build: Retry docker run if it gets a container command not found error

tfredian:
For unknown reasons the docker run command sometimes gets a command
not found error. This may be an nfs cacheing problem. This change will
repeat the docker run command several times before giving up.


zack-vii (approved):
seems fine

PR #1178, opened 2017-12-19
project: threadsafe tdi

zack-vii:
attempt to make tdi threadsafe with lesser use of locks


zack-vii:
retest this please

zack-vii:
retest this please

zack-vii:
The issue seems to be that tdi does not deep copy data on TdiGetIdent when handling public var. This is probably what you would want for performance but this prevents tdi from being thread safe.
workaround would be to deepcopy publics, possibly add a CONST modifier (used like PUBLIC or P
PRIVATE) to the var makeing it readonly but allows to read the var without deepcopy.
```tdi
CONST PUBLIC _const = myconst()
PUBLIC _const = 123
%TDI-E-CONST cannot modify a constant var
```

zack-vii:
down to one segfault that only occurs on alpine 64x?

tfredian (approved):
looks good

PR #1186, opened 2017-12-29
Fix: cmdParse.y - remove overlapping arguments to strcat

joshStillerman:
The code which removed the double double quotes ("") used strcat
with overlapping arguments to copy the string on top of itself to
turn them in to single double quotes (").

This results in undefined behavior.  On some systems like RHEL7
it works fine but on others (centos -7, 3.10.0-693.5.2.el7.x86_64)
it garbles the string.

Valgrind shows this as an overlapping strcat error.

There were also some bugs in the generated code.  It needed:
local include of mdsthreadsafe.h
include of config.h
consistent naming of PVALUE - missing '_'

For now turn off the sign compare warning


tfredian:
ok to test

tfredian (approved):

PR #1190, opened 2018-01-01
REVIEW REQUIRED! Big renaming of all of the config.h includes

kgerickson:
This is the big rename change that will move all of the #include<config.h> lines to instead be #include <mdsplus/config.h>.  This does two things... 1) It moves the generated header into include/mdsplus, and 2) it renames it.  This is the simple change.  The more complicated one would be not having to install mdsconfig.h at all, but we aren't there yet.  Baby steps :)

This needs some review.  I noticed about a dozen places where we had code that generated C code that included config.h.  I don't know if I got them all.


tfredian:
ok to test

tfredian:
It looks like there are config.h includes in some of the tests that still need to be fixed.

kgerickson:
Updates pushed

tfredian (approved):
Looks good to me.

PR #1193, opened 2018-01-03
Build: Remove incorrect uses of volatile

tfredian:
In an attempt to fix various compiler warnings I inadvertently added
some "volatile" keywords to the code which were not needed. After
finding the real reason and solution for the compiler warnings I
removed most but not all of the erroneous volatile additions. This
should remove the remainder of these keywords that were added when
attempting to fix the compiler warnings.


kgerickson (approved):
LGTM

PR #1199, opened 2018-01-09
Build: Remove jar creation output from trigger console

tfredian:
The trigger console log is getting obscured by the output of
the making of the jar files for use in installers so this output
is moved to a make_jars.log file which can be included in the trigger
job artifacts.


tfredian:
retest this please

PR #1206, opened 2018-01-18
Feature:handle names longer than 12 chars

GabrieleManduchi:
Node names longer that 12 chars are now splitted and a subtree created. For example, adding a node whose name 'THISISALONGNODENAME' will ceate a subtree whose root is names 'THISISALONGN' and a son names 'ODENAME'. This feature is intended for usage in ITER IMAS where a way for mapping long names into tree nodes is required.

### DIFF
--- a/javatraverser/CompileTree.java
+++ b/javatraverser/CompileTree.java
@@ -253,8 +253,9 @@ public class CompileTree extends Thread
             if(type.equals("node"))
             {
                 try  {
-                    if(name.length() > 12) name = name.substring(0,12);
-                    nid = tree.addNode("."+name, NodeInfo.USAGE_STRUCTURE, 0);
+//                    if(name.length() > 12) name = name.substring(0,12);
+//                    nid = tree.addNode("."+name, NodeInfo.USAGE_STRUCTURE, 0);
+                    nid = addNode("."+name, NodeInfo.USAGE_STRUCTURE);
                     if(usageStr != null && usageStr.equals("SUBTREE"))
                         subtreeNids.addElement(nid);
                     tree.setDefault(nid, 0);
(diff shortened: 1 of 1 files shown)


javatraverser/CompileTree.java
```diff
@@ -395,6 +397,21 @@ void recCompile(Element node)
         }
     }
 
-
+    NidData  addNode(String name, int usage) throws Exception
+    {
+	NidData prevNid = tree.getDefault(0);
+        String currName = name.substring(1);
+	String prevNode = "";
```
zack-vii, line 404: You should probably use StringBuilder and StringBuffer to do the splitting. but thats just me.

GabrieleManduchi: I prefer avoiding as far as possible the usage of StringBuffer unless forced by performance considerations. However, as you told, it's just a personal attitude.

tfredian: This seems like a very ugly solution to me. Would it not be better to create some arbitrary unique nodes (i.e. N[0-9|A-Z]11) and assign tags with the long node name?

GabrieleManduchi: It is a long history: this thing will never be used outside the ITER IMAS context. In that context they want to uniquely map a hierarchy expressed in XML (the IMAS ontology) into a MDSplus tree. The problem is that they can use names longer that 12 chars (they describe the field names of the defined data structures). The solution adopted so far was to map the name into a 12 char one doing some hashing, but of course the transformation was not unique, and name clashes were arising.  
It would have been possible to pick a unique identifier for the node name and to same in extended NCI the actual name, but we need that the access to nodes (and therefore finding the node given the name) is efficient since this operation occurs in simulations (i.e. storing and retrieving simulation results to/from pulse files).  Since the actual tree organization of the pulse file is not exported to users (there is an intermediate software layer that translates the conceptual data view into the specific MDSplus tree structure) this solution provides an unambiguous way of storing information with minimal impact in the current implementation of the data access layer for IMAS (i.e. this change in CompileTree and a single change in a single method in the whole data access layer), and not affecting at all performance in data retrieval (which would otherwise require rewriting large parts of the data access codes in order to create pre-defined search tables  and so on).  
I would like to stress the fact that this feature does not affect anything else nor is intended to be used in any other MDSplus application. To make it cleaner, I may enable it with a specific qualifier in the CompileTree command, generating an error for names longer that 12 chars otherwise.

tfredian (approved):

PR #1207, opened 2018-01-19
fix many problems by temporarely using private context

zack-vii:
threadUnitTest with treeUnitTest now even sometimes succeeds


zack-vii:
threadUnitTest on treeUnitTest passes even helgrind now

tfredian (approved):

PR #1217, opened 2018-01-23
Feature: Update python epydoc documentation

tfredian:
There is a makedoc.sh script in mdsobjects/python which will generate html
documentation for the python modules. This documentation has not been updated
for quite a while so this commit will bring it up to date.


tfredian:
looking into making doc's during release builds instead

PR #1218, opened 2018-01-24
Zck phase xxx lookup

zack-vii:


tfredian (dismissed):

tfredian:
retest this please

zack-vii:
there is an unrelated race in dcl that i wil fix tomorrow
retest this please

joshStillerman (dismissed):
It seems that if changing the order fixes this, then there must still be a race condition.  Or am I missing something ?

tfredian:
You are missing something. There was a race condition and this fixes it. 
The tdi fun which loads the site-specific phase table sets both the 
phase names and the phase numbers into two different public tdi 
variables. It loads the phase names first and then the phase numbers. 
The function checks to see if the phase name public variable was 
allocated (initialized) and if not loads the table setting the two 
public vars. There was a race condition in that if two threads hit this 
simultaneously the first one notices that the phase name var was not 
allocated and calls the function to load the tables beginning with 
setting the phase name var. The other thread notices that the phase name 
var was allocated and then references the phase number var which was not 
yet allocated by the first thread and causes an error. The loading of 
phase tables was never done before in a threaded application so never 
caused a problem. Implementing a more complicated dispatching scheme 
could run into this issue if it was done as a multi-threaded 
application. It showed up in a new test which does a lot of various tdi 
expressions simultaneously in multiple threads. Simply changing the 
phase table access to check on the public variable loaded last instead 
of the first to see if the phase info was initialized fixes the problem.

joshStillerman (approved):

PR #1219, opened 2018-01-26
No-more no database - fix setTimeContext and more

zack-vii:
* Fixed segfault when setTimeContext with no tree opened yet
* fixed setTimeContext for cpp
* reduced warnings due to redefinition of JAVA_JAR in Makefile.in
* added some Tests
* better handling of global trees under python


zack-vii:
better solution on the way

zack-vii:
good to go

tfredian (approved):

PR #1221, opened 2018-01-29
Some race cond and pointer issues in tree and tdi

zack-vii:
* some test transfers to cpp
* segfautl in tcl dir/tag
* race in TreeDelete and NewTagSearch
* uninit var in TdiGetArgs


treeshr/TreeDeleteNode.c
```diff
@@ -255,7 +250,7 @@ extern void _TreeDeleteNodeExecute(void *dbid)
       edit->deleted_nid_list=dnid;
     }
     else
-      memcpy(edit->nci + nid.node - edit->first_in_mem, &empty_nci, sizeof(struct nci));
+      memset(edit->nci + nid.node - edit->first_in_mem, 0, sizeof(struct nci));
```
zack-vii, line 253: empty_nci is sizeof(struct nci) zeros

tfredian (approved):

PR #1223, opened 2018-01-30
Feature: Generate the html epydocs for python during package build

tfredian:
The html documentation provided in the python package use to be
included in github but it was never routinely updated. Generating
this documentation at package build time will ensure it is up to date.
The files included in the epydocs are highly dependent about the contents
of the python modules so instead depending on the expected package content
checking to include the latest documentation files this doc directory will
be excluded from the contents checking.


zack-vii (approved):
probably alright

PR #1225, opened 2018-01-30
Add new tditest tests to the standard MDSplus test suite

tfredian:
This pull request includes the addition of tests driven by the tditest application in the standard set of
tests which are performed during pull request tests and release builds. In adding this new build feature the new tests uncovered a problem with adding shared library based devices when editing an MDSplus tree. This problem is also fixed in this pull request.


zack-vii (approved):
cull extend and 2..4 needs to be fixed but they have not been broken by this commit

tfredian:
Awesome!

⁣Sent from BlueMail ​

PR #1229, opened 2018-02-05
Zck python3 tests

zack-vii:
This probes for platforms that require byteswap.. as it might not be tested at all and is likely broken


AndreaRigoni (approved):
Great job Timo !

PR #1231, opened 2018-02-09
Zck int128

zack-vii:
* use u/int64_t datatypes instead of own typedefs
* fixes many issues with octaword operations e.g. division, decompile to decimal
* use system's RNG
* removed redundant code in TdiAnd
* centralized int128 operations in int128.h header


zack-vii:
int128 is buggy dont_merge

PR #1234, opened 2018-02-18
Zck fix cull and some cleanup

zack-vii:


zack-vii:
```tdi
3..5 == 0 : 5
23..5 == 2 : 5
_a..5 == _ : 5
```
is a compiler issue. 
```tdi
3 .. 5 == 3 : 5
```
works!

zack-vii:
fixes race from fc27 #'302 64-san-thread in MdsTdiTest.log

zack-vii:
the fix for 1..3 compile is hacky but works. if anyone knows what cases it in the first place, please let me know.

I assume it has to do with the fact that both '..' (len=2) and ':' (len=1) map to the same operation

zack-vii:
retest this please

joshStillerman (approved):
As per your EMAIL I am approving...

Issue #1238, opened 2018-02-23
python expression widgets broken


joshStillerman:
The glade device setups are blanking out fields that are strings.  

See acq132 device setup for example.

The problem is that at some level the widget is looking for a 'string' - python primitive, but is getting a String - mdsplus primitive.

This can be fixed by coercing the type in the widget code, but since that code has not changed in a long time, we should track down why it broke, and fix it in the MDSplus python objects instead.

PR #1239, opened 2018-02-23
Fix: decompile must return a primitive python string

joshStillerman:
for the python widgets to use the result.

Expression widgets in python setups were being displayed as, and reset to, empty unless they were
numeric scalars.

This was because the decompile of record of a node was getting returned as an MDSplus String instead
of a python string.

Closes: #1238

### DIFF
--- a/mdsobjects/python/mdsdata.py
+++ b/mdsobjects/python/mdsdata.py
@@ -386,7 +386,7 @@ class Data(object):
     def decompile(self):
         """Return string representation
         @rtype: string"""
-        return _cmp.DECOMPILE(self)._setTree(self.tree).evaluate()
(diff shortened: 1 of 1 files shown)


tfredian (approved):

mdsobjects/python/mdsdata.py
```diff
@@ -386,7 +386,7 @@ def __fixTreeReferences__(self,tree):
     def decompile(self):
         """Return string representation
         @rtype: string"""
-        return _cmp.DECOMPILE(self)._setTree(self.tree).evaluate()
+        return str(_cmp.DECOMPILE(self)._setTree(self.tree).evaluate())
```
zack-vii, line 389: will this work on python3 as well … should be handled by __str__ but just checking

PR #1241, opened 2018-02-26
Gm fix settimecontext

GabrieleManduchi:
There were still cases in which TreeSetTimeConcest() was called with a null dbid, e.g. when called within mdsip. This check avoids crashes in mdsip that occurred using jScope (in turn using SetTimeContext via mdsip). When no dbid is defined, success is returned  (TreeSetTimeContext is void)

### DIFF
--- a/mdsobjects/cpp/testing/MdsConnectionTest.cpp
+++ b/mdsobjects/cpp/testing/MdsConnectionTest.cpp
@@ -73,6 +73,10 @@ void test_tree_open(const char *prot, const unsigned short port)
     TEST1( AutoString(data->getString()).string
            == "[0.,0.,0.,0.,0.,0.,0.,0.,0.,0.]" );
 
+    // this should not hang //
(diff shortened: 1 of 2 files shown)


zack-vii (changes requested):
did you read my comment on the issue. if you experience a uninitialized dbid you are either using an old version or you are setting the dbid to null yourself.

AndreaRigoni:
I added a simple line within the cpp tests that verifies the issue .. it is supposed to work here with the Gab changes and seems to hang the process otherwise.

tfredian:
I was able to reproduce the hang issue with calling SetTimeContect() in the MdsConnectionTest. I suspect this may be occurring because of the way the mdsip "servers" are being started which is now done using forking of the test itself instead of running mdsip processes. Perhaps the thread locks used in the tree context are getting messed up when the servers are forked from the original test process.

zack-vii:
replaced by pr #1243 which includes the test and fixes the core problem

PR #1247, opened 2018-03-05
Build: Avoid potential gpg locking problems caused by aborted builds

tfredian:
Under some circumstances file in the .gnupg directory used when signing
installers can become locked. This is rare but it can cause builds to hang indefinitely. This change makes a copy of the signing information and uses the copy for each installer creation.


tfredian:
retest this please

tfredian:
retest this please

tfredian:
retest this please

zack-vii (approved):
seems to work

PR #1248, opened 2018-03-05
Zck stack copysegmentindex

zack-vii:
Fixes a problem with updateSegment that resulted in corrupted data.
avoid recursive stack-memory expensive CopySegmentIndex.
only read and store sindex offsets and process them in correct order


tfredian (dismissed):

tfredian (approved):

zack-vii:
needs some more tests but this pr seems to have fixed the problem with compression of large files on windows systems as well :)

PR #1249, opened 2018-03-06
Fix: Repair installer signing

tfredian:
Previous attempt to prevent signing interference between concurrent builds
introduced some other errors preventing the installers from being signed. This should fix this problem.


tfredian:
retest this please

tfredian:
retest this please

zack-vii (approved):
well well

PR #1250, opened 2018-03-06
Zck scaled segments

zack-vii:
adds a new feature of scaled segmentsyou can add the value expression for a node using
```
TreeSetSegmentScale(nid, expr)
```
methods like getSegment(nid,idx) and getRecord() will look for the SegmentScale xnci and return:
```
Signal(expr, segment_data, segment_dim)
```
if the xnci exists and is not NULL. otherwise it will return:
```
Signal(segment_data, *, segment_dim)

```
, the original result.


tfredian (changes requested):
looks like a small memory leak in MdsTreeSegments.
http://jenkins2.mdsplus.org/job/fc26/437/valgrindResult/

zack-vii:
yea

tfredian (approved):

PR #1257, opened 2018-03-11
Fix: Do not call TreeSwitchDbid with NULL

tfredian:
Tdi1Using was calling TreeSwitchDbid with a NULL which caused a segfault later.


tfredian:
A similar issue might exist in javamds/mdsobjects.c which will need to be investigated.

PR #1259, opened 2018-03-12
Fix: jTraverser2 instead of jTraverser replacement

### LINKED ISSUE #1037: jTraverser and jTraverse2 conflict
Both jTraverser and jTraverser2 build a jar file names jTraverser.jar. Therefore the command jTraverser will open the version that has been last copied to /usr/local/mdsplus/java/classes in install.
It would be better to name the two tools differently and to provide a new script, say jTraverser2

zack-vii:
should fix #1037 and update javatraverser2 to new apis

### DIFF
--- a/javatraverser2/Makefile.am
+++ b/javatraverser2/Makefile.am
@@ -26,7 +26,7 @@ bin_SCRIPTS = $(JAVAROOT)
 CLASSPATH_ENV = CLASSPATH=$(JSCH):$(JAVAROOT)
 
 # Build the final jar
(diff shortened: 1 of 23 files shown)


tfredian (dismissed):
Looks ok to me.

zack-vii:
jTraverser2 only included in windows now.. the script to include it in other oses requires a change to the build scripts i think. to be done at some point but nor part of this pr

tfredian (approved):

PR #1263, opened 2018-03-19
fix: On Macintosh - listening for multiple UDP Events

joshStillerman:
On Macintosh the second event created by a process errors out with

  Cannot bind socket
  : Address already in use

This is due to the MAC socket implimentation needing BOTH
  SO_REUSEADDR
  SO_REUSEPORT
options.

Using both does not appear to break things on Linux


tfredian (approved):
looks good

PR #1268, opened 2018-03-24
Fix: Fix problem with adding extended attributes to empty nodes

### LINKED ISSUE #1267: Wrong behavior of TreeGetXNci / TreSetXNci in given sequences
When adding the mdsobject support to extended attributes I observed a strange behavior. If things occur in the following sequence:

SetXNci  on node 1

SetXNci on node 2

then GetXNci on node 1 fails while GetXNci on node 2 succeeds

HOWEVER if I repeat the sequence, both GetXNci work.

In summary it appears that something wrong occurs in the sequence of two XSetNci for nodes without that attribute already set.

tfredian:
The code for adding extended attributes mishandled nodes that
did not contain data. Instead of writing the attributes header
to the end of the datafile it instead wrote it to beginning
of the datafile. Subsquent adding of attributes to other nodes
without data also wrote it to the beginning of the datafile
overwriting any other settings found in that location. This fix
ensures that if there are no extended attributes defined for a node
the extended attributes will be written to the end of the datafile.

Tests for TCL> SET/SHOW ATTRIBUTE commands were also added to the
tditest/testing/tdi-tcl.tdi test.

This should fix https://github.com/MDSplus/mdsplus/issues/1267


zack-vii (approved):
oh yes.

Issue #1269, opened 2018-03-26
Java needs revised handling of global tree context (dbid).


tfredian:
A change made to the handling of the global tree context has broken the java interface which in turn has broken the Matlab interface. I don't fully understand the current ctx handling in the java code so can't suggest a better solution yet.

The current implementation attempts to change the global dbid context when invoking tdishr functions when doing operations which involve Tree or TreeNode instances. To do this it uses the getActiveTree() and setActiveTree() functions to get the current global tree, set the new global dbid based on the Tree or TreeNode instance, call the tdishr routine, and, in theory, return the global dbid context to its original state. This operation is now broken in many ways. First, if getActiveTree() is called and there is no tree open in the global context the code raises an exception. This is preventing any data from being returned from tree nodes. Second, the existing code is riddled with what appears to be typo errors because in many cases the code calls getActiveTree to get the "currTree", calls setActiveTree to set the "ctxTree", executes the expression, and then erroneously calls setActiveTree with the "ctxTree" instead of what I assume should be the currTree".

I believe all this code dealing with tree ctx's in  java needs to be revisited. Lots of changes have been made to the handling of the global tree ctx during a major project of Timo's to make MDSplus much more thread safe. Major changes to the python code was made in conjunction with this thread safe improvements and intricate locking was implemented to ensure thread safe behavior when switching the global tree context temporarily to evaluate expressions. The java (and perhaps the c++) code appears to need considerable work similar to the python interface to improve the thread safe behavior of the object oriented api's.

PR #1270, opened 2018-03-26
Fix: Avoid Tree Not Open errors when switching ctx in java

tfredian:
Currently the code switch the global tree context raises exceptions
if no tree is open in the global context. This suppresses that exception
to enable data in tree nodes to be accessible in expressions.

This is related to issue https://github.com/MDSplus/mdsplus/issues/1269 but only provides a quick temporary fix.


GabrieleManduchi (approved):
ok for me

PR #1271, opened 2018-03-26
Gm extended attributes

GabrieleManduchi:
Added getter and setter methods for extended tree node attributes in C++ and java


tfredian (approved):
looks good

Issue #417, opened 2016-03-17
Bug in _TreeFindNodeWild with syntax such as .*** and :***


tfredian:
This bug may have been in MDSplus for a very long time but nobody used this search syntax before. That syntax started occurring mainly because of the way TreeNode.getNodeWild was implemented.

When  the triple asterisk syntax is preceded by a period or colon, _TreeFindnodeWild seems to pass over some of the nodes multiple times sometimes causing the routine to return many more nodes than it should and in some cases infinitely recursing until the process runs out of memory.

joshStillerman:
The syntax is described at:
http://www.mdsplus.org/index.php/Documentation:Users:Tree_syntax

Keith, can you write a parser for it ?  I have an idea of how to make the search code much more simple.
-Josh

joshStillerman:
Given that some wildcard searches are broken, and there is ambiguity in the syntax for **** I propose we make the following syntax decisions:

[\[treename::]tagname{.|:}]

Four single character punctuations

1. '.' the child
2. ':' the member
3. '^' the parent
4. '~' wildcard **both** member and child

and these can be tripled to mad **breadth first** searches

1. '...' all children
2. ':::' all members
3. '~~~' all members and children
4. '^^^'  all parent

four of these is a syntax error

a special case would first translate *** to ~~~
(four of these is also a syntax error)

Some examples

```
a.b*.c  - a child named c of a child starting with b of a member or child called a 
a.b*:c - a member named c of a child starting with b of a member or child called a
a.b*~c -a member or child named c of a child starting with b of a member or child called a
a.b...c  - a child called c someplace under the child hierarchy under a.b
a.b:::c - a member called c someplace under the member hierarchy under a.b
a.b~~~c - a child or member called c someplace under the child/member hierarchy under a.b
a.b*~~~ 
etc...

```

PR #1274, opened 2018-03-28
Jas fix search

joshStillerman:
Fix:  redo TreeFindNode and TreeFindNodeWild

Given that some wildcard searches are broken, and there is ambiguity in the syntax for **** this change makes the following syntax changes:

[[treename::]tagname{.|:}]

Four single character punctuations

'.' the child
':' the member
'^' the parent
'~' wildcard both member and child

and these can be tripled to make breadth first searches

'...' all children
':::' all members
'~~~' all members and children
'^^^' all parent
four of these is a syntax error

a special case would first translate *** to ~~~
(four of these is also a syntax error)
however ****  becomes ~~~* which is OK.

Some examples

a.b*.c  - a child named c of a child starting with b of a member or child called a 
a.b*:c - a member named c of a child starting with b of a member or child called a
a.b*~c -a member or child named c of a child starting with b of a member or child called a
a.b...c  - a child called c someplace under the child hierarchy under a.b
a.b:::c - a member called c someplace under the member hierarchy under a.b
a.b\~\~\~c - a child or member called c someplace under the child/member hierarchy under a.b
a.b*\~\~\~ 
etc...

Fixes issue #417


tfredian (dismissed):
Looks great!

tfredian (approved):

PR #1277, opened 2018-03-30
Gm nodejs mdsip

GabrieleManduchi:
Restful interface to mdsip + UDP events
This node.js interfaces provides a HTTP interface listening on port 8081 for remote expression evaluation, i.e. thin client. The mdsip protocol is handled directly by the JavaScrip code.
It serves multiple clients possibly connected to different mdsip servers
In addition, it provides WebSocket interfaces for receiving MDSplus events (UDP only). Registration to the event is performed via the rest interface, and then ipon event reception, the string <event_name>:<event data> is sent via WebSocket to the connected listeners. 
Only string raw data associated with events is currently supported

 URI to call the code:
1 - http://<host>:8081/connect?ip=<ipAddress[:port]>  
      return: connection ID to be used in the following transactions

2 - http://localhost:8081/eval?expr=<expression>&idx=<connection ID>
      return: decompiled answer from remote expression evaluation

3 - http://localhost:8081/register?event=<event_name>
     register for event reception (on all clients connected via WebSocket).

WebSocket package installation is required:
   npm install websocket

Server command:
$ node mdsipRest-multi.js


tfredian (commented):
Is this intended to be registered with npm for users to install using the npm command or should it be installed as part of the MDSplus installers (perhaps java)? We found that using different repositories (i.e. pypi for python) is not a good solution since the code often needs to match the version of MDSplus being used. However since this is using thin client it might not matter.

tfredian (approved):
This looks fine depending on the mechanism planned for user installation per the previous comment.

joshStillerman:
Please wait to merge this ....
It either needs more directions or some webpack like dressing to operate.

I am investigating.

zack-vii:
any update on this one?

GabrieleManduchi:
It is ready,
    in few days I will provide sources and tutorial.

                               Gabriele

PR #1280, opened 2018-04-03
Feature: update ax_docker_build

AndreaRigoni:
Adds some features from the ultimate version of the file. Adds also the with-docker-profile
that permits to make trace of the code within the container in debug, thus the dbg will be usable in container shell.


tfredian (approved):
Looks ok to me. Need documentation on wiki though.

PR #1281, opened 2018-04-03
build:  Need to escape <backslash>~  in commit messages

joshStillerman:


tfredian (approved):

joshStillerman:
will redo against rebased

PR #1282, opened 2018-04-03
Fix: Repair a couple bugs in TreeFindNode.c

tfredian:
There was a problem with the handling of the TOP tag and the matching
of tag names. This should fix this.


joshStillerman (approved):
GREAT !

PR #1284, opened 2018-04-06
Few bug fixes

robertomendozaphd:
My first GitHub pull request so not sure if I am doing this correctly. Feedback encouraged.


zack-vii (approved):
i assume the fix works. i approve the way the bugs have been fixed. TODO to ourselves: we should cover this in a test. good find

PR #1288, opened 2018-04-09
Feature: Restore the Visual Studio MdsObjectsCppShr-VS.DLL

tfredian:
This should restore a cpp dll which is compatible for use with
Visual Studio built applications on Windows.


tfredian:
DO NOT MERGE YET

joshStillerman (approved):

PR #1289, opened 2018-04-17
Fix: Repair MDSplus sign key used in redhat stable repo installer

tfredian:
The stable redhat repo rpm was still installing the older less secure
signing public key which does not match the key used to sign the rpms.


tfredian:
retest this please

joshStillerman (approved):

PR #1296, opened 2018-04-26
Fix: Remove excess TdiData call when merging segments

tfredian:
TdiData() is called inside TreeXTreeDefaultSquish so not needed
to be called in XTreeGetTimedRecord.c.


GabrieleManduchi (approved):
Ok, as discussed

PR #1299, opened 2018-05-01
Fix: Repair older python package removal when upgrading MDSplus

tfredian:
The previous scripts for cleaning off MDSplus python packages used
pip remove in a loop but on some newer versions of pip the pip remove
command does not return a failure status if the package was not found.
This produced an infinite loop when remove or upgrading MDSplus python
installers particularly on debian based platforms. This should fix this
problem. Unfortunaly an upgrade on debian systems will cause this infinite
loop until the newer packages are installed. To prevent this infinite
looping the following files will need to be edited by hand before
removing the associated packages if the exist on the debian based system:

/var/lib/dpkg/info/mdsplus-alpha-python.prerm
/var/lib/dpkg/info/mdsplus-*devices.prerm

You need to insert an single line containing:

exit

right after the:

line at the top of the file.


zack-vii (approved):
seems ok

PR #1300, opened 2018-05-02
Gm jscope fix dtype f data

GabrieleManduchi:
Added management of DTYPE_F data items in mdsmisc/ScopeUtilities used by jScope MdsDataProvider

This fixes the issue reported by Antoine Merle


merlea:
Shouldn't the same check be done for the X array?

GabrieleManduchi:
It is highly unlikely, but it cannot be excluded, so I added also the check for X array.
For jScope exceptions with empth signals, I could not reproduce exactly the reported case, but I got an exception in another piece of code. Now a check has been added in jScope for both cases.

tfredian (approved):

PR #1305, opened 2018-05-08
Zck python args

zack-vii:
a = MDSplus.ADD(1,2);
MDSplus.MULTIPLY(a,a)
(1 + 2) * (1 + 2)
the first argument of the first sum is somewhat random


tfredian (approved):
Good job!

PR #1313, opened 2018-05-22
Build: Improve Visual Studio build configuration

tfredian:
This update corresponds to a new mdsplus/docker:windows image which places
the Visual Studio files in directories similar to the directory structures
found when Visual Studio 2017 Community edition is installed on a Windows
system. It also removes many VS files in these directories which are not
needed for building the MdsObjectsCppShr-VS.dll library.


tfredian:
retest this please

MDSplusBuilder (approved):

Issue #1314, opened 2018-05-22
Has the single character wildcard '%' been replaced with '?' ?


merlea:
Hello,

I am not sure but it seems to me like the single character wildcard '%' does not work anymore in node specifications. Furthermore it seems that it has been replaced by '?'.

Here are some examples using mdstcl.

With our currently installed version (circa sept 2016) on the TCV data server, this is what I get:

```
TCL> dir \results::dml:ck%%%

\RESULTS::DML

 :CK001        :CK002        :CK003        :CK004        :CK005       
 :CK006        :CK007        :CK008        :CK009        :CK010       
 :CK011        :CK012        :CK013        :CK014        :CK015       
 :CK016        :CK017        :CK018        :CK019        :CK020       


Total of 20 nodes.
```

With the alpha branch, this is what I get:

```
TCL> dir \results::dml:ck%%%
mdsdcl: --> failed on line 'dir \results::dml:ck%%%'


Total of 0 nodes.
TCL> dir \results::dml:ck???

\RESULTS::DML

 :CK001        :CK002        :CK003        :CK004        :CK005       
 :CK006        :CK007        :CK008        :CK009        :CK010       
 :CK011        :CK012        :CK013        :CK014        :CK015       
 :CK016        :CK017        :CK018        :CK019        :CK020       


Total of 20 nodes.
```

Thanks,
Antoine.

PR #1317, opened 2018-05-23
Fix: Allow both % and ? wildcard chars in FindNode

tfredian:
The refactoring of the findNode code changed the single character
wild card match from % to ?. To remain compatible with existing
code this fix allows both % and ? wildcard characters.


tfredian:
This should fix https://github.com/MDSplus/mdsplus/issues/1314

AndreaRigoni (approved):
Everything fine for me

PR #1331, opened 2018-06-06
Feature: Add Fedora Core 28 support

tfredian:
This will provide support for fc28 builds. Fc28 uses version 8.1
gcc compilers which added several new compiler warnings which
required changes to many of the modules. Some use pragma's to
ignore the warnings but since many of the new warnings were not
recognized in older versions of the compile the pragma's needed
to be invoked inside compiler version checks.


tfredian:
retest this please

joshStillerman (approved):

PR #1336, opened 2018-06-22
Fix: reverting FC28 changes to this file as it broke MdsValue

joshStillerman:
This commit reverts MdsLibIld.c  - the changes to this file for
FC28 compiler warnings did not work.

Symptom -

in IDL
```
set_databse, 'logbook'
DSQL('select count(*) from entries', cnt)
```
produced the error:
```
answer = lonarr(1,)
                   ^
% Syntax error.
% Compiled module: EVALUATE.
```
This is a temporary rollback of the change so that new kits will work.


tfredian (approved):
I guess this is a quick workaround disabling fc28...

tfredian:
retest this please

PR #1344, opened 2018-06-26
Fix: actlog should flush after each line

zack-vii:
flushes stdout after each line which is required if you want to have a useful output that you might wanna parse by a third party program.


tfredian (approved):
looks good

PR #1346, opened 2018-06-26
Fix: there is a typo in deploy/packaging/linux.xml

joshStillerman:
The typeo in deploy/packaging/linux.xml causes ubuntu updates
to hang.


tfredian (changes requested):
This fix is invalid. The colon was correct but the problem is probably that this htsdevice scripts were not fixed like the rest of the python device installers. Make htsdevices resemble the other device installer scripts such as w7xdevices. Btw. the colon is not a typo. In bash it is basically a noop but the script itself is wrong.

joshStillerman:
OK -

   Editing that script on the target system to change it to a ; made the 
update work.  We can talk about it tomorrow.

PR #1353, opened 2018-07-05
Fix: Avoid loading numpy in matlab if not using python bridge

tfredian:
On platforms such as Ubuntu18 there are incompatibilities between
the python numpy libraries and those used by matlab. This causes
memory issues and segfaults if the matlab scripts contain py functions
which loads the numpy module even if those py functions are not
executed. This fix uses str2func to reference those py functions
when defers the load until the users explicitly tries to use the
python bridge instead of the java bridge to access MDSplus.


zack-vii (approved):
seems ok

PR #1358, opened 2018-07-06
Zck python tab complete

zack-vii:
this finally add subnodes to the tab completion of Trees and TreeNodes. :) so easy once you learn how


tfredian (approved):
Very cool!

PR #1359, opened 2018-07-06
Fix: Repair rpm installation scripts

tfredian:
When splitting the debian and redhat installer scripts the build
script was inadvertently using the debian scripts for noarch packages
which led configuration changes to be removed during package updates.
On redhat the uninstall scripts of the installed package is invoked
after the updated package is installed and the redhat uninstall scripts
check an argument to determine if it is an update or a perminant package
deinstall so that it won't remove things if it is an update. The debian
packages don't have this problem. It will take a couple updates for the
fix to take effect though since the deinstall scripts of the installed
package will be invoked until the fixed scripts are added with new versions.


tfredian:
retest this please

tfredian:
retest this please

MDSplusBuilder (approved):

Issue #1349, opened 2018-06-29
TreeshrHooks in python


joshStillerman:
It may be useful to allow users to create Treeshr Hooks in python.

joshStillerman:
We will provide hooks for all segment operations. For this can be used to prototype things that we then want to move into statically defined extra operations.  

Hooks could be stacked then we could leave them as hooks.

tfredian:
Look into adding a useHooks flag in nci. If not set node related hooks will not be called.

tfredian:
TDI base TreeShr hooks were added by: https://github.com/MDSplus/mdsplus/pull/1467
See that pull request for more details on the new feature.

It is not yet clear how much these hooks will be used by user sites. There are use case issues with such hooks since they will be invoked on the system where running applications call the corresponding TreeShr routines. If the functions performed by these hooks need to communicate with applications running on other systems the site will have to implement a framework used by the hook. The addition of nci flags to control which nodes invoke node specific operation hooks is a fairly large project as it will require changes to a lot of code to manipulate the flags in applications such as TCL, Traverser as well as all the object oriented api's. For this reason the addition of these nci flags will be postponed until it is clear that they would be useful. The current implementation allows users to control which hook types are invoked simply by defining an environment variable, TreeHooks.

PR #1361, opened 2018-07-10
Feature: Add tdi based treeshr hooks

tfredian:
This change will enable the implementation of tdi based
treeshr hooks which could be implemented in either tdi
or python. Like the C based hooks the tdi TreeShrHook() function
would be called with a action id, a treename, a shot number and a
node number (nid). A TreeShrHook.fun-template is included to
demonstrate how it is called during various tree operations.

The code will first try to call the C based implementation and if
there is no hooks library it will try the tdi based hooks.

This should implement the feature request:

https://github.com/MDSplus/mdsplus/issues/1349

### DIFF
--- a/deploy/packaging/debian/kernel.noarch
+++ b/deploy/packaging/debian/kernel.noarch
@@ -208,6 +208,7 @@
 ./usr/local/mdsplus/tdi/treeshr/TreeSetDbiItm.fun
 ./usr/local/mdsplus/tdi/treeshr/TreeSetDefault.fun
 ./usr/local/mdsplus/tdi/treeshr/TreeSetNciItm.fun
+./usr/local/mdsplus/tdi/treeshr/TreeShrHook.fun-template
 ./usr/local/mdsplus/tdi/treeshr/TreeTurnOff.fun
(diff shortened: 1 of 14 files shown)


tfredian:
Before merging we might consider some improvements to this enhancement.
First, it might be useful to improve the way tdi loads funs (both tdi and python). If we can determine that either a fun was not found in MDS_PATH or it fails to compile perhaps we should keep track of 'broken' functions and avoid looking for and/or compiling them on future calls by a process. Subsequent calls could return the same error. A reset_public() call could clear this error cache so that someone working on a fun could retry loading the fun.

Second, since this tdi hooks feature is new perhaps we should revisit hooks in general instead of just emulating the existing hooks. The original hooks were implemented back when shots were stored on an optical jukebox to enable managing retrieval of offline shots. What might be some more modern functionality for these hooks and should the api be changed to improve their usefulness.

This PR provides a tdi implementation matching the original C hooks functionality. Before this is merge and possibly gets used by customers this might be the time to think about any changes or improvements to the api. The C hooks can remain the same for backward compatibility.

Possibly something to discuss in the upcoming MDSplus developer meeting if not before.

PR #1371, opened 2018-08-09
Fix: ssh for Windows add username if not specified; passwd and ppk auth

zack-vii:
add a few more options to the windows mdsip-client-ssh

* ssh /h is not syntax of OpenSSH
* ssh was missing "/bin/sh -l -c"
* plink does not automatically choose username and we cannot use prompt
* plink can use env:SSH2RSA *.ppk file to authenticate
* plink can use env:PASSWD to authenticate
* plink agent should try to forward agent


tfredian:
retest this please

tfredian (approved):

PR #1375, opened 2018-08-23
Feature: increase the number of args to LibCallg to 255

zack-vii:


zack-vii:
seems like there is a os specific limit?

PR #1380, opened 2018-08-27
Fix: Remove all references to MdsRoutinePrefix in idl routines

### LINKED ISSUE #1378: IDL mdsconnect broken
As of 7.46-02 (and maybe before), an auxiliary function in MDSCONNECT (IDL code) was deleted.  The problem is with the MDSDISCONNECT program, since it tries to call the missing function.

I found this (Google search), pasted it in, and had no issues...  This should be fixed either by removing completely in MDSDISCONNECT, or adding back to MDSCONNECT.

function MdsRoutinePrefix
  ;Descriptors or argc/argv convention?
  ;Shouldn't need to be changed for macos
  ;By the way, probably didn't need to port routines beginning with "Idl"
  if !version.os eq 'vms' then return,'' else return,'Idl'
end

tfredian:
The MdsRoutinePrefix fnuction was used for VMS compatibility and
was removed but references to the function was still present
in other functions. This removes those references.

This should fix https://github.com/MDSplus/mdsplus/issues/1378

### DIFF
--- a/idl/mdsdisconnect.pro
+++ b/idl/mdsdisconnect.pro
@@ -1,5 +1,5 @@
 pro mdsdisconnect,status=status,quiet=quiet, socket=socket
-  forward_function mds$socket, MdsIPImage, MdsRoutinePrefix
+  forward_function mds$socket, MdsIPImage
   image = MdsIPImage()
   status = 1
   if keyword_set(socket) then $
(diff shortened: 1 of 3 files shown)


tfredian:
retest this please

MDSplusBuilder (approved):

PR #1382, opened 2018-09-04
 Feature: allow to specify a default tree, which will open on TreeFOPENR

zack-vii:
This is useful for trees that contain only static information such as parent trees only containing subtree nodes and structures. We want to use this for test shots and the parent eval tree.

* added new named shot number -2 -> default
* forces CreateDbSlot to reopen default tree on condition change
* added status Tree-I-OPENDEF to inform the user about the default tree upon first TreeOpen
* TreeFileName will show the real file
* Edit and New will not be redirected


zack-vii:
retest this please

zack-vii:
retest this please

zack-vii:
so, any comments on this  =)

joshStillerman (commented):
Timo-
  Regardless of the merits of this new feature, it appears that this PR contains may changes to jTraverser2.  Are they related to this ?  

About the 'feature'  I don't fully understand what you are trying to accomplish.  Could you explain a bit more ?

zack-vii:
the javatraverser2 stuff should not be part of this pr. but was introduced due to a rebase onto a local alpha instead of the remote version, thanks for noticing.

The idea is to help out for two use cases.

1. You have  a main tree that inherits all diagnostics as subtrees. Some diagnostics are distributed among multiple servers and hence for performance reasons require multiple subtrees e.g. "cam1", "cam2", "cam3", ... .
In order to allow the RO of such a diagnostic to do calibrations and test shots with the entire diagnostik but without giving them control over the main tree it is useful to introduce a  dummy tree "cam" that only contains the subtrees "cam*". since the tree is never been written, it could be just a single file that is opened no matter what shot is requested. also this allow the RO for opening one shot file of regular shots that contain the entirety of his/her diagnostic without opening all other diagnostics.

2. the generation of a dynamic eval tree. We have one master eval tree that should contain the subtrees of diagnostic/topic specific eval subtrees. Since the number of subtrees might be growing over time and the ability to create evaluated data for shots in the past. it is useful to have one source "eval" tree that is edited whenever a new evaluation is added. It should be opened for each shot that is requested can therefore be used as an overview over available evaluation data.

Case 1 we have already implemented and it works nicely. Specially when editing or viewing data in traverser. It helps the ROs to navigate their 4 to 17 subtrees without getting overwhelmed by the entirety of 50.

zack-vii:
Jenkins has trouble with rsync when attempting to transfer the source.

```
time -f %e rsync -a mdsplus-jenkins2:/home/jenkins/slave/workspace/1-stable-release/ /mdsplus-docker/jenkins/workspace/raspberrypi/src
05:07:39 Permission denied, please try again.
05:07:39 Permission denied, please try again.
05:07:39 Permission denied (publickey,gssapi-keyex,gssapi-with-mic,password).
05:07:39 rsync: connection unexpectedly closed (0 bytes received so far) [Receiver]
05:07:39 rsync error: unexplained error (code 255) at io.c(226) [Receiver=3.1.2]
```

tfredian:
There is already an undocumented feature in MDSplus for accomplishing what I think you are trying to accomplish. You can defined the tree path environment variable to be '~n' and that will call a tdi function to get the file path of the tree. For example:
```
$ export mytree_path='~'
$ mkdir ~/tdi
$ cat - > ~/twf/mytree_tree_filename.fun
public fun mytree_tree_filename(IN _shot)
{
  return('myserver::/trees/mytree/mytree_model');
}
<ctrl-d>
$ export MDS_PATH='/home/myaccount/tdi;$MDS_PATH'
$ mdstcl
TCL> set tree mytree/shot=42
```
This will cause mdsplus to open the model tree for mytree no matter what shot number is used. You can use this feature to open different tree instances based on the shot number you are trying to open so it can used for a wider set of use cases. Perhaps you create one mytree instance per day and you want to use it when opening trees for the particular day based on the shot numbers. The mytree_tree_filename function can then be used to pick the appropriate instance based on the shot number being opened.
Give this a try and see if it satisfies your needs. In your case you would just need to create a cam tree model which contained all the cam1.-camn subtree references then your just open cam,shot and it will appear that you were opening cam_nnn but actually you'd be opening cam_model but all the subtree would be camn_nnnnn.

I also fixed the build problems. One of the jenkins slaves had a configuration problem preventing it from rsync'ing the sources from the trigger job.

zack-vii:
Not necessary as functionality is available thru ~n

PR #1389, opened 2018-09-10
Fix: Allow the use of MdsException(string) in python

### LINKED ISSUE #1388: Problems using `MdsException` in v7.46.1
I have installed MDSplus version 7.46.1 on a newly deployed Kubuntu system, and I am trying to use it with python. I am able to `import MDSplus`, but I have had trouble using `MdsException`s. These are raised in key parts of OMFIT.

- Older versions of MDSplus (e.g. 7.1.13) had `MdsException` subclassed on `Exception` and doing nothing (class definition contained only `pass`)
  - So, `raise MDSplus.MdsExceptin('message')` worked fine
- `MdsException` is an alias in this version: `mdsdata.py:MdsException = MDSplusException`
- `MDSplusException` is defined in `MDSplus/mdsExceptions.py`
- Inside of the `__new__(cls,*argv):` method, it tries to cast argv[0] as an int: `status = int(argv[0])`, which doesn't work if the user tries to pass a string as an error message (common practice).
- Any code expecting to handle an `MdsException` will get a `ValueError` instead, and probably won't handle it the same way / may fail. If the error isn't handled at all, and instead just left to the user, it's more confusing to read.

Could this exception class be made to accept strings, or must other applications be altered to pass integers?

tfredian:
The definition of the MdsException class in python was changed
to be equivalent to the MDSplusException class which is used to
create exceptions from numeric status codes which are returned by
MDSplus C functions. This prevents the ability for applications
to create MdsExceptions using a string argument. This change
makes MdsException a subclass of Exception and the MDSplusException
class be a subclass of MdsException.

This should fix: https://github.com/MDSplus/mdsplus/issues/1388

### DIFF
--- a/deploy/build.sh
+++ b/deploy/build.sh
@@ -23,7 +23,7 @@ SYNOPSIS
                [--distname=name] [--updatepkg] [--eventport=number]
                [--arch=name] [--color] [--winhost=hostname]
                [--winbld=dir] [--winrembld=dir] [--gitcommit=commit]
-               [--jars-dir=dir] [--make-jars] [--docker-srcdir=dir]
+               [--jars] [--make-jars] [--docker-srcdir=dir]
 
 DESCRIPTION
     The build.sh script is used for building, testing and deploy MDSplus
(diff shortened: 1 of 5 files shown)


eldond:
I think this would close #1388.

tfredian:
retest this please

tfredian:
retest this please

tfredian:
retest this please

zack-vii (dismissed):
should not break anything

tfredian:
retest this please

tfredian:
retest this please

MDSplusBuilder (approved):

PR #1390, opened 2018-09-10
Fix: Allow the use of MdsException(string) in python

tfredian:
The definition of the MdsException class in python was changed
to be equivalent to the MDSplusException class which is used to
create exceptions from numeric status codes which are returned by
MDSplus C functions. This prevents the ability for applications
to create MdsExceptions using a string argument. This change
makes MdsException a subclass of Exception and the MDSplusException
class be a subclass of MdsException.

This should fix: https://github.com/MDSplus/mdsplus/issues/1388


zack-vii (dismissed):
should not break anything

tfredian:
retest this please

MDSplusBuilder (approved):

eldond:
Thanks for taking care of this.

PR #1392, opened 2018-09-13
Zck float time

zack-vii:
The confusion is you should not need to care about second or nanoseconds too much especially when comparing mixed time vector (double [s], long [ns]). FloatToTime and FloatToDelta should do the same thing. And it would not break anything if the do. the extra day is useless when working with int64 as it is signed. the common conversion is nano_seconds vs seconds and can simply be *1e9 or /1e9. transparent for everyone.
As a result SetTimeContext would be transparent as well use int64 or Uint64 is you want to put in ns use other datatypes otherwise.


GabrieleManduchi:
ok, finally I got the point!  it seems to me that everything should work since resampling should work also with negative int64 numbers (for negative float times). If not done, you may make a few tests with bot negative and positive start and end times. Let me know if/when you have done these tests and I will approve the PR. Tom, I see a slight change also in TreeSegments.c. Could you check it?

zack-vii:
The getSegment stuff is already included in #1382

zack-vii:
retest this please

GabrieleManduchi (approved):

PR #1393, opened 2018-09-14
Gm jscope improvements

GabrieleManduchi:
Reshape dim_of results so that they have a single dimension. This prevented visualization of some signals retrieved from bidimensional arrays.


joshStillerman (commented):
Gabriele -
  Why is the expression DIM_OF(SET_RANGE(_s,_a)) ?  DIM_OF extracts a dimension from a signal.  Is this what you meant ?

GabrieleManduchi:
In the case of the signals defined by Anton, expressed as an element of a bidimensional array (he stored all the channels of a given ADC in a single segment), dim_of returned a bidimensional array where the second dimension was 1 (e.g. [[1],[2],[3]]) and in that case the following operations performed by the expressions used by jScope for dynamic resampling failed. Using set_range() transforms in any case the returned dimension in an 1D array and fixes the problem

joshStillerman:
But why Dim_of()

GabrieleManduchi:
Because that information is required to decide if and what resampling performing. The problem was  that dim_of() was returning a bidimensional array with one dimension set to 1 instead of a 1D array for Anton's signals. This is the reason of the added set_shape()

GabrieleManduchi:
hmmmm, let me have a look at the code when I come back next monday

GabrieleManduchi:
Now I remember:  dim_of() for an array such as [[1],[2],[3]] returns [0]. This is why I had to use set_shape. This happened not for a normal signal but for the result of a data() operation (this was the condition where jScope failed to display).
To summarize what happened:
- If jScope was requested to look at the signal, internally using dim_of() everything works because the signal is stored with the correct timebase.
- if jScope was requested to display data(signal), then dim_of() was applied to the result of data() (thus losing the original timebase) that turned out to be the array as above, and dim_of() returned [0] instead of an array of increasing integers as one would expect.

(Note that jScope is able to handle properly a returned signal of the form  [[1],[2],[3]] in visualization) 

A bit complicated, but apparently set_shape() fixes the problem and does not break anything.

tfredian (dismissed):

tfredian:
Gabriele, this pull request can't be merged because your branch was apparently not up to date with the alpha branch and the branch can't be rebased with the current alpha because there are conflicts. I would suggest either starting over after doing:
```
$ git checkout alpha
$ git pull
$ git checkout alpha -b gm-jscope-improvements2
```
make changes again to the scopeUtilities.c code and commit, push and create a new PR.

or... you can
```
$ git checkout gm-scope-improvements
$ git rebase alpha
```
fix conflicts in scopeUtilities.c and then
```
$ git rebase --continue
$ git push
```

tfredian:
If you chose the git rebase --continue path then you can reopen this PR and retest.

GabrieleManduchi:
Further checked and rebased

PR #1397, opened 2018-09-20
Feature: Allow MDS_DEVICES environment to identify device list

tfredian:
MDS_DEVICES can now be used to add site specific devices. You
can specify a colon delimited list of tdi functions that will
return lists of devices. If there is no MDS_DEVICES environment
variable defined it will default to including only the default
MDSplus devices defined by KbsiDevices(), MitDevices(), RfxDevices()
and W7xDevices(). If you have a local device list you could
define MDS_DEVICES to something like "mydevices:MitDevices" and your
supported device list would include your devices and the devices
listed in the MITDEVICES.fun. All python device implementations
will still be found based on the MDS_PYDEVICE_PATH environment variable
which contains a semi-colon delimited list of directories containing
device support modules.

### DIFF
--- a/tdi/dev_support/MDSDEVICES.py
+++ b/tdi/dev_support/MDSDEVICES.py
@@ -5,7 +5,7 @@ lock = Lock() # locks the cache
 def MDSDEVICES():
   with lock:
    if cache[0] is None:
-    from MDSplus import Device,tdi,version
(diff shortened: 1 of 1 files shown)


tfredian:
retest this please

joshStillerman (approved):

PR #1398, opened 2018-09-24
Gm fix j scope dimension

GabrieleManduchi:
I created a new pull request for the fix of Anton's problem in evaluating signals from jScope


GabrieleManduchi:
Sorry having created so much trouble....the changes in the code are basically the same of the other PR, just skipping an useless evaluate() operation.

tfredian (approved):

PR #1403, opened 2018-09-24
Zck segement128

zack-vii:
fixes from #1382


tfredian:
good news and bad news...
good: It looks like the tditest status is now detecting errors
bad: but the errors seem to be coming from unknown reasons. The output of test-tdishr.tdi seems to now be wrong while the output using the current alpha produces what seems to be correct and the same answers as in the test-tdishr.ans.

zack-vii:
retest this please

tfredian (approved):

PR #1404, opened 2018-09-24
Jas more findnodewild

joshStillerman:
the previous fix to findnodewild caused it to return the current default which is changed behavior.
This fixes that.


tfredian:
retest this please

tfredian (approved):

PR #1405, opened 2018-09-25
Adjustment for Windows MinGW-w64

HerveAncher:
Some adjustments to fit MinGW-w64 accross MSYS2 environment.
Correct libxml2 include path.

### DIFF
--- a/configure
+++ b/configure
@@ -8150,7 +8150,7 @@ case "$host" in
        MAKEFLAG_O=""
        IDLMDSEVENT=""
        LIBPRE=""
-       CFLAGS="$CFLAGS -fno-strict-aliasing -mno-ms-bitfields ";
+       CFLAGS="$CFLAGS -fno-strict-aliasing -mno-ms-bitfields -DWIN32 -D_WIN32 ";
        LD="$CXX -static-libgcc"
        FCFLAGS="$FCFLAGS -fno-range-check"
        FORLD="$FC"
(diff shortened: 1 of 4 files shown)


tfredian:
ok to test

tfredian:
The liblzma reference added to the mdsshr/Makefile.in prevents the windows build from succeeding since that library is not installed on the windows Docker image used for building the windows distributions.  See: http://jenkins2.mdsplus.org/job/windows/1555/console

I have removed the -llzma option and the jenkins build now succeeds: http://jenkins2.mdsplus.org/job/windows/1556/console

Similarly the addition of -DWIN32=1 and -D_WIN32=1 compiler flags should not be necessary as the MinGW64 compiler provide these. Adding these will not cause a build problem so if for some reason these are not getting defined in your build environment they can be left in. Otherwise this change is not needed.

Finally, like the previous PR, the commits do not include changes to configure.ac. All changes to configure should be performed by changing the configure.ac and both the configure.ac and configure files should be committed to the PR branch. When other configure changes are needed, developers will modify configure.ac and generate a new configure which would remove the changes requested in this pull request.

You do not need to delete your current PR branch to make these changes. Simply pull the latest version of your branch and then make the necessary changes, commit and push the commits back to github. This will trigger another test build automatically.

HerveAncher:
Hi,

Without the reference to library lzma, I can not compile in my side, because my libxml2 required this library.
May the flag '-llzma' can be added dynamically depending on dev environment?
Or the Docker image can be update to include this library?

HerveAncher:
OK, I have added the correct code to check if lzma library is present and ready to use.
I also adapt the Makefiles to use the new variable @LIBLZMA@

tfredian (approved):

PR #1406, opened 2018-09-26
Fix: removed unnecessary decompression in ReadSegment

zack-vii:
Descriptro will decompress automatically when needed.
This allows a much faster remote access when reading compressed segmetns via mdsip using

_v=*;_d=*;_s=TreeShr->TreeGetSegment(val(_nid),val(_seg),xd(_v),xd(_d));serializeout(`List(*,_s,_v,_d))

e.g. by upcomming and maybe current jTraverser2

### DIFF
--- a/treeshr/TreeSegments.c
+++ b/treeshr/TreeSegments.c
@@ -1079,13 +1079,8 @@ static int ReadSegment(TREE_INFO* tinfo, int nid, SEGMENT_HEADER* shead, SEGMENT
       for (i = 0; i < ans.dimct; i++)
         ans.arsize *= ans.m[i];
       if (compressed_segment) {
-        EMPTYXD(compressed_segment_xd);
         int data_length = sinfo->rows & 0x7fffffff;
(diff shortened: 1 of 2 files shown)


zack-vii:
Can someone look into this, please?

tfredian:
I'm not sure of the benefit of this PR. Assuming the segments are selected either manually or using the start and end values held in the segment index, they have to be decompressed anyway before they are used so what good does postponing the decompression? Please explain what I am missing?

zack-vii:
well as i mentioned if you read segmented data via thin client (like jTraverser2 does; and jScope could do for frames) you can serialize the compressed object and decompress it on the client, instead. Hence, you transfer less data and the readout is faster. I implemented the PR on our data server and reading images is up to 3 times faster. The key here is that TreeGetSegement returns the D_CA data in a D_XD. using SerializeOut on the XD prevents server side decompression entirely.

tfredian (approved):

PR #1409, opened 2018-10-02
Fix: NULL treenode search term

AndreaRigoni:
The pull request adds a test and a fix proposal for Tom based on where the debugger stopped.
The tree getNode caused a SEGFAULT if a NULL search term is passed from an empty string argument.


AndreaRigoni:
Seems that there could be something more to look at ... the fail comes from valgrind
http://jenkins2.mdsplus.org/job/fc26/1557/valgrindResult/pid=13046,0x0/

tfredian:
retest this please

PR #1412, opened 2018-10-03
Gm jscope improvements

GabrieleManduchi:
This version implements the features agreed in September meeting, namely:
- Implementation of a brand new LocalDataProvider in a pure Java implementation using pipe communication with a subprocess running mdsip-client-local
- Following the same approach a new SSHDataProvider has been added. 
- Added Command line Data Provider selection: -s <mdsip address> forces the use of MdsDataProvider connected to the specified mdsip server; -l forces the use of LocalDataProvider; -ssh <ssh address> forces the use of SSHDataProvider. In the above bases, the Network menu bar item is disabled, otherwise it is enabled and Data Server selection is performed in the usual way. NOTE: when a configuration file is passed as argumen and either -l -s or -ssh are defines, the data server specified in the configuration file is ignored.
- Made jScope compatible with DWScope configurations

PR #1413, opened 2018-10-03
Twf jscope improvements

tfredian:
Redoing https://github.com/MDSplus/mdsplus/pull/1412 because the commits in that PR are all messed up.


joshStillerman (approved):
I tested this -
  The -l switch works as expected
   It can now read in the dwscope configurations that it was having trouble with before.

PR #1420, opened 2018-10-11
Zck bootstrap

zack-vii:
since running bootstrap seems to be highly os specific lets agree on a docker image and now we can run it from anywhere with the same result.
just execute:

deploy/build.sh --os=bootstrap

One could think of a test that runs this and checks if files have changed to enforce the usage of automake


tfredian:
Basically a good idea but since we are in the habit of running mdsplus/bootstrap after changing configure.ac or Makefile.am's perhaps it might be better to change mdsplus/bootstrap to do the build command and move the commands currently in bootstrap into the docker bootstrap script.

zack-vii:
fair point i in this case changer the image to fc23 as it produces the same result but the images is smaller\

AndreaRigoni:
It's a very nice idea, but I think we should retain the possibility to do bootstrap also if you don't have docker installed as mdsplus does not require docker for basic compilation but it does for packaging... so what if we keep bootstrap as is and add a bootstrap phase to the build.sh?

A further discussion would be whether or not to ship configure and Makefiles with the code or leave any user that want to build mdsplus to do bootstrapping letting the build.sh to do its bootstrap phase by itself.

zack-vii:
hm I did not intend to revolutionized the way of how to build MDSplus in general. what about I rever the last change and we change our habit and run deploy/build.sh --os=bootstrap if required.
Some host systems may have the same versions of the automake toolschain. lets just make sure the output is the same. some versions of automake introduce redundant changes to all outputfiles in form of comments that will make it more difficult to track the important changes.
if you do ./bootstrap and git diff shows many changes you should use docker if only the intended chages show up you are good to go.

AndreaRigoni (dismissed):
I think this is a great idea ... we could even setup a distro to prepare code for pull request with also code beautifiers and the bootstrap ran always with the same tools. before doing the pull request..

AndreaRigoni:
maybe to follow the Tom advice we could add an option argument in bootstrap that let you switch to docker if you have it.

tfredian (commented):
I'm concerned that leaving bootstrap there we'll forget and the next time one of us mucks with configure.ac we'll invoke bootstrap instead of deploy/build.sh --os=bootstrap. You've spent way too much time on this already but one last suggestion might be to change the bootstrap command to require an argument specifying either -docker or -nodocker and have it either complain or prompt if the argument is missing. This will at least give us a reminder to use docker and still permit it to fall back to not using docker if necessary. It is pain when someone puts in a PR for minimal changes to configure.ac and runs ./bootstrap on a different os which causes unnecessary changes to a zillion makefiles etc.

zack-vii:
would an echo about the problem be sufficient,
```echo
WARNING: If you intend to issue a pull request on github.com, please use "deploy/build.sh --os=bootstrap" instead
```

tfredian (approved):

PR #1425, opened 2018-10-12
Revert "Fix: implemented missing timeout in mdstcpip for windows"

tfredian:
Reverts MDSplus/mdsplus#1424
Sorry stupidly ignored the DO NOT MERGE label!


zack-vii:
well its actually just the fist argument which should be tablesize and not 1.. also i was hoping to use teh same code for linux as poll and select are supposed to be interchangeable but if i do so, the cpp MdsConnetcionTest fails..

Also under windows mdstcl 
dispatch/command/server=localhost type test
 now always (but the first time) prints 
Dispatcher select loop failed
Last client addr = 127.0.0.1

zack-vii (approved):

PR #1430, opened 2018-10-16
Zck win socket timeout

zack-vii:
should work but mdsdtcl is still reporting 

TCL> dispatch/command/server=localhost:8100 type test
TCL> dispatch/command/server=localhost:8100 type test
Dispatcher select loop failed
Last client addr = 127.0.0.1
TCL>

### DIFF
--- a/deploy/os/windows.env
+++ b/deploy/os/windows.env
@@ -1,4 +1,5 @@
 export WINEDEBUG=-all
+export WINEPREFIX=/workspace/winebottle
 export PYTHONPATH="Z:\\python27\\Lib;Z:\\python27\\Lib\\site-packages"
 export PYTHONHOME="Z:\\python27"
 export PyLib=python27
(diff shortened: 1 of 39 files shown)


zack-vii:
looks like we were lazy with select most of the time
nfds should be the largest file identifier in the list +1
We used FD_SETSIZE which is 64 under windows but sockets can surely have an id larger than that.

MS says the parameter is ignored, though.

Anyway, big bug was that fds under windows are of type SOCKET aka. (void*) under linux they are int. in some functions the mapping was wrong and only set the first 4 bytes of the SOCKET.
In return Connections from Windows could not be reused.

zack-vii:
there are more bug to it

zack-vii:
retest this please

zack-vii:
could be good now

tfredian (approved):

PR #1437, opened 2018-10-24
Fix: Require consistent dispatch table context

### LINKED ISSUE #1302: minor bug when dispatch/build is not called again after opening another shot
mdsdcl hangs possibly due to conditional actions (WHEN is required ACTION node):
e.g.:
```tcl
set tree main
create pulse 1
set tree main/shot=1
dispatch/build
dispatch/phase INIT
...
set tree main
create pulse 1
set tree main/shot=1
dispatch/phase INIT
```
If the init phase contains conditional actions, the dispatch blocks after the first Action in the dependency list.

minor bug as it only happens during manual testing.

tfredian:
The dispatch table will be only valid when the context is referring
to the tree the dispatch/build command used to generate the table.
The dispatch table is now attached to the tree context so if you
open another tree or close the original tree you will get an error
if you try to issue the dispatch/phase. If you open another tree
and then set the tree context to the original the dispatch table
will be available.

Fixes https://github.com/MDSplus/mdsplus/issues/1302


tfredian:
retest this please

zack-vii (approved):
tested it and it work as expected
maybe we should add a test somewhere

zack-vii:
retest this please

PR #1444, opened 2018-10-29
Zck python fun

zack-vii:
TODO: improvement; add python methods loaded from tdi into a designated module in order to prevent any kind of name clash with existing python modules


tfredian:
retest this please

zack-vii:
the windows version is a bit hacky but works fine.. for some reason windows cannot lock the file properly. gdb reports a segfault in _lockfile when it tries to do PyRun_SimpleFileExFlags.

edit: execfile might not be available on all python versions :(

tfredian (approved):

PR #1445, opened 2018-11-02
Zck python fun

zack-vii:
fixes execfile and exec for python3 in both TdiExtPython and .py
TODO: test Py, addfun, and pyfun in either tditest or python tests - separate PR ?!

### DIFF
--- a/conf/valgrind-python.supp
+++ b/conf/valgrind-python.supp
@@ -455,14 +455,21 @@
 }
 
 {
-   python 32bit problem during thread shutdown via fork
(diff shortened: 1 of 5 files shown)


zack-vii:
@tfredian: if the test succeeds, could you verify and merge this PR. you already merged the previous one that does not work with python3.

zack-vii:
ok tested it for linux and windows, both python 2.7 and 3.4. should work for other versions too.

tfredian (approved):

PR #1450, opened 2018-11-07
Fix: select can return EINPROGRESS - which is OK

joshStillerman:
The select in mdstcpip can return 0 with errno == EINPROGRESS
if the connect returned it.  This is not an error.  I do not think
the select needs to be redone, but do not have a good way to verify
this.


tfredian:
The EINPROGRESS error in errno is set by the connect when a connection timeout is set using the MDSIP_CONNECT_TIMEOUT environment variable. The select is used to wait for the connection to complete and should return non-zero unless the connection attempt has timed out. That is the point of the select. If MDSIP_CONNECT_TIMEOUT is not defined then the connect is done without NONBLOCK and the select is not needed. If a site is getting an "Error in connect" error then either something is wrong with their network or they need to increase the value of their MDSIP_CONNECT_TIMEOUT.

Issue #1423, opened 2018-10-12
Issuing (e.g. from tdic) TreeSetDefault('\null')  crashes tdishr


GabrieleManduchi:
Apparently only names that start with \null  crash tdishr. Every other  name (correct or incorrect) does not crash the system.

tfredian:
Is this still a problem? If so, can you provide an example to demonstrate it? This was entered after Josh's PR to fix an similar issue https://github.com/MDSplus/mdsplus/pull/1416
I tried the following code and did not get a crash:

```
#include <stdio.h>
extern int TreeOpen();
extern int TreeSetDefault();
int main() {
  char *node="\0";
  int status=TreeOpen("cmod",-1);
  printf("status=%d\n",status);
  status = TreeSetDefault(NULL);
  printf("status=%d\n",status);
  status = TreeSetDefault(node);
  printf("status=%d\n",status);
  return 0;
}
```

GabrieleManduchi:
It is the name '\null'   that crashes. Very easy to reporduce with tdic:

[mdsplus@scmcl6 national]$ tdic
TDI> Treeopen('test',-1)
265388041
TDI>  TreeSetDefault('\xxx')   ->a node that does not exist
265388290                             -> status wrong, as expected
TDI> TreeSetDefault('\null')  -> or any name that starts with \null, e.g. \nullxxxx
Segmentation fault (core dumped)

PR #1455, opened 2018-11-13
Fix: Avoid segfault in FindNode with invalid nodename syntax

tfredian:
If a bad syntax is passed to FindNode or FindNodeWild a segfault
occurs. This should fix https://github.com/MDSplus/mdsplus/issues/1423


joshStillerman (approved):

tfredian:
retest this please

PR #1467, opened 2018-11-21
Feature: Add tdi based treeshr hooks

tfredian:
Add various treeshr hooks which will call a TreeShrHook fun with
one argument which is a Dictionary. Depending on the hook the
Dictionary could contain fields such as:

  "type" - the name of the type of hook
  "tree" - the name of the tree
  "shot" - the shot number of the tree
  "nid"  - the node being operated on
  "data" - the data being written to the tree

The type of hooks currently provided are:

  "OpenTree"     - Tree being opened
  "WriteTree"    - Save of edited tree
  "GetNci"       - Getting node characteristics
  "GetData"      - Retrieving data stored in a node
  "CloseTree"    - Closing an opened tree
  "RetrieveTree" - Tree not found so hook to retrieve from storage
  "OpenTreeEdit" - Opening a tree for edit
  "OpenDatafileWrite" - Open tree datafile for write
  "MakeSegment"  - Store a segment in a node
  "MakeSegmentFull" - Store a segment in a node (data passed to hook)
  "MakeTimestampedSegment" - Store timestamped segment
  "MakeTimestampedSegmentFull" - Store segment (data passed)
  "PutSegment"      - Store segment
  "PutTimestampedSegment" - Store timestamped segment
  "PutTimestampedSegmentFull" - (data passed)

Hook selection is determined by the TreeHooks environment variable.
You can select particular hooks by listing the hook types in a
comma delimited list (i.e. export TreeHooks=OpenTree,GetNci,CloseTree)
or you can use the keywords allhooks and fullhooks. If the string,
"allhooks", is found in the TreeHooks environment variable all of the
hooks which do not have full in their type name will be called. If
the string, "fullhooks", is found all the hooks with full in their type
will be called. If you include allhooks and fullhooks in the TreeHooks
environment variable all the hooks will be called. The names included
in the TreeHooks environment variable will be case insensitive.

### DIFF
--- a/treeshr/RemoteAccess.c
+++ b/treeshr/RemoteAccess.c
@@ -321,6 +321,7 @@ int ConnectTreeRemote(PINO_DATABASE * dblist, char *tree, char *subtree_list, ch
 	  info->flush = (dblist->shotid == -1);
 	  info->header = (TREE_HEADER *) & info[1];
 	  info->treenam = strcpy(malloc(strlen(tree) + 1), tree);
+	  TreeCallHookFun("TreeHook","OpenTree",tree,dblist->shotid, NULL);
 	  TreeCallHook(OpenTree, info, 0);
 	  info->channel = conid;
 	  dblist->tree_info = info;
--- a/treeshr/TreeAddNode.c
+++ b/treeshr/TreeAddNode.c
@@ -748,6 +748,7 @@ int _TreeWriteTree(void **dbid, char const *exp_ptr, int shotid)
                 MDS_IO_LOCK(info_ptr->channel, 1, 1, MDS_IO_LOCK_RD | MDS_IO_LOCK_NOWAIT, 0);
                 status = TreeNORMAL;
                 (*dblist)->modified = 0;
+		TreeCallHookFun("TreeHook","WriteTree",info_ptr->treenam, (*dblist)->shotid, NULL);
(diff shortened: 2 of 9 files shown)


GabrieleManduchi:
Is the hook available only in python?

GabrieleManduchi:
If I am not wrong, there are possible memory leaks in TreeCallHook _strcasestr() (the strings obtained by strdup should be freed?) and in isEnabled() (enabledHooks not freed in return statements?)

GabrieleManduchi:
Not clear to me why the following two consecutive calls:

    TreeCallHookFun("TreeNidHook","PutData",info_ptr->treenam, info_ptr->shot, nid, NULL);
    TreeCallHookFun("TreeNidDataHook","PutData",info_ptr->treenam,
		    info_ptr->shot, nid, descriptor_ptr, NULL);

Seems like the hook for is called twice with type = PutData, the second time with a dictionary argument that is just the dictionary of the first call plus data. Am I missing anything?

tfredian:
Good finds Gabriele. I guess eventually I should add some tests of calling hooks as this leak would have been found by sanitizers or valgrind. The second call to TreeCallHookFun should have been PutDataFull as it passes the data to the hook. These should now be fixed.

GabrieleManduchi (approved):

PR #1468, opened 2018-11-22
Refactor Thick Client; support as if distributed

zack-vii:
Do CTX_PUSH and CTX_POP dbid in c.
*memory fix for TreeGetDatafileSize
*TreeCleanDatafile can auto close tree before compressing: should it reopen after?
*provides _TdiExecute(void\*\*dbid,...) -> TdiExecute(...) generically for all Tdi functions
*provides _mdsdcl_do_command_dsc(void\*\*dbid,...) -> mdsdcl_do_command_dsc
*eliminates tree._TreeCtx in python api
*lock/unlock and free cleanup in treeshr/RemoteAccess.c
*added python device do test-dev-py.tdi


zack-vii:
added a new spin-off of the tunnel protocol. it spawns a thread instead of a new process. its not much of an improvement in regular use but makes it a lot easier to debug mdsip in gdb as both threads client and server are in one process.

zack-vii:
If this succeeds it need some testing of it working together with old-version servers and old-version clients.
this will possibly be replaced by a new pr.

zack-vii:
retest this please

PR #1483, opened 2018-12-13
Fix: add a knob to rate limit cryocon daq

joshStillerman:
The Cryocon18i gets unhappy if you talk to it too fast.  This
commit adds a new knob which sets the rate in Hz for streaming
DAQ


tfredian (approved):
Looks ok to me.

PR #1488, opened 2018-12-17
Fix: include _version.py modules in module used for tests

tfredian:
The tests were using the source directory tree for the MDSplus
module during tests which did not include the _version.py module
identifying the version of the python module. This fix changes
the tests to use the buildroot copy of the python module which
does contain the version info.

This also cleans up the install target in the top Makefile.in when
copying the _version.py files from the build directory to the
target root.


tfredian:
retest this please

joshStillerman (approved):
Clear as mud 
I HATE MACROS

PR #1489, opened 2018-12-18
Fix: no need for extra module and only print if properly installed an…

zack-vii:
…yway


tfredian:
This defeats one of the main causes of mismatched versions. Some users are fetching the MDSplus sources to get the MDSplus python module code and installing the module from sources using tools like pip and then having problems because the current sources don't match the version of the MDSplus libraries were installed on their system. This would disable the warning in this case which is precisely where the warning is needed. Also since you removed the extra version_check.py module the packaging content checking for the python installer would need to be modified. I am closing this PR now and will continue trying to enable the checking for all cases including the "make tests" operation during builds.

PR #1490, opened 2018-12-27
Fix: Use MDS_END_ARG in MdsLibIdl Treeshr calls

joshStillerman:
The change from MDS_END_ARG=0 to MDS_END_ARG=1 made the non standard
(non macro) use of 0 to terminate argument lists wrong.

This caused MdsOpen, 'tree', shot  to segfault on some platforms.
Changed calls to TreeOpen and TreeClose to use the MDS_END_ARG macro

### DIFF
--- a/mdslibidl/MdsLibIdl.c
+++ b/mdslibidl/MdsLibIdl.c
@@ -69,7 +69,7 @@ EXPORT int IdlMdsClose(int argc, void **argv)
   int status;
   BlockSig(SIGALRM);
   if (argc > 1)
-    status = TreeClose((char *)argv[0], ((char *)argv[1] - (char *)0));
+    status = TreeClose((char *)argv[0], (int)(intptr_t)argv[1]);
   else {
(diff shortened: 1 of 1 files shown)


zack-vii:
in fact the change was from -1 to 1 and does not apply for TreeOpen as TreeOpen does not take and varargs input. the terminating 0 was the readonly flag. TreePutRecord does not take one either.

zack-vii:
@joshStillerman: Can you make a case where the issue occurs. Your fixes wont work, i.e. they break it even further (readonly=1, utility_update=1).

tfredian:
In both cases the argument in question was the shot number. Perhaps the handling of the argv could be done better but this apparently had been working fine for a very long time.

zack-vii:
ah not the original change was. josh changed the trailing ', 0) ' into MDS_END_ARG), the cast of the shot number was my doing before i realized that TreePutRecord also takes no var args.

PR #1491, opened 2019-01-02
Fix: Repair thick client create pulse

tfredian:
At some point the argument list of the tdi fun for
calling TreeShr->TreeCreatePulseFile were changed to
match the order of the arguments of the c library call. The
thick client implementation was still calling the tdi fun
with the old argument list order so doing a create pulse
of a tree with subtrees using thick client resulted in an error.

This change calls the TreeShr function directly instead of using
the tdi fun which should be more efficient and has the correct
argument order.


zack-vii (approved):
should do the trick

PR #1498, opened 2019-01-08
zack.vii playground

zack-vii:
* length of integer to str conversion inconsistent
* tcl("dir/full") failed with tdi devices
* c/lib devices wont work without python
* running python code without python caused sigsegv
* added lib-device and tdi-device add-test to tditest treeshr test

### DIFF
--- a/servershr/ServerSendMessage.c
+++ b/servershr/ServerSendMessage.c
@@ -539,8 +539,6 @@ static void ResetFdactive(int rep, SOCKET sock, fd_set* active){
   DBG("reset fdactive in ResetFdactive\n");
 }
 
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wclobbered"
 static void ReceiverThread(void *sockptr){
   atexit((void*)ReceiverExit);
   CONDITION_SET(&ReceiverRunning);
@@ -594,7 +592,6 @@ static void ReceiverThread(void *sockptr){
   fprintf(stderr,"Cannot recover from select errors in ServerSendMessage, exitting\n");
   pthread_exit(0);
 }
-#pragma GCC diagnostic pop
 
 int is_broken_socket(SOCKET socket)
 {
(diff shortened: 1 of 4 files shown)


tfredian:
I don't understand why you added back the old MDSDEVICES.fun which was replaced by MDSDEVICES.py which has more features. I don't recall whether tdi looks for py funs before tdi funs but we certainly don't need both since only one would be invoked. Perhaps this was unintentional?

In the future I would recommend breaking up pull requests like this into smaller ones. It is difficult to review PR's with 60 files being changed particularly if the changes include a variety of unrelated fixes and/or enhancements and include more descriptive comments in the commits.

zack-vii:
Yes i agree well my intend was to introduce the opcode_t and while i was doing it i stumbled upon these other issues.. i will break it up once i am satisfied.
I added the tdi fun back in as it is used by shared lib devices (the once that have their methods written in c.)
I changed the code to try tdi if it found the .py file but could not load it (no python). This way you still can add those devices without python

tfredian:
I'm not sure we need to worry about things not working if no python is available. As we implement more fun's in python it is questionable whether we spend much effort implementing workarounds for systems which don't have python or don't have the MDSplus module installed, especially the old c based device types since they are mostly only usable on linux systems which should have python and the MDSplus module installed automatically.

tfredian:
In addition the mdsplus and the mdsplus-mitdevices packages require python and mdsplus-python packages to be installed. The platforms that don't support camshr (such as windows and osx) don't include things like mitdevices anyway which is the only package now that includes c based devices.

zack-vii:
ok well i stumbled uppon it while i tried to get teh tditest treeshr test working for all platforms.. since some of them dont have python and well in the end i had to give up on it because of windows

tfredian:
Perhaps we might think about making some of the test suites platform specific. Could be that tditest is the only one causing problems. We could add an argument to the script which could be platform (redhat, debian, windows, macosx, alpine, etc..) and invoke test-[name]-[platform].tdi and compare to test-[name]-[platform].ans if the corresponding tdi file exists.

zack-vii:
not a bad idea.

PR #1500, opened 2019-01-11
introduce typedefs where applicable

zack-vii:
This introduces typedefinitions that will allow compile time checks of consistency with arguments and field also this will improve readability of gdb prints

extends #1499


tfredian (approved):
Cleans things up quite a bit. Maybe they'll do the same to VMS since things like mdsdescrip.h was a modified version of the VMS OS header file. ;-)

PR #1501, opened 2019-01-14
Feature: Add "code in" pydevice support

tfredian:
This enables the feature to develop pydevices in which the device
methods are stored in the tree. The only time the python module file
is referenced is to add the device to an MDSplus tree. From that point
on the methods executed on the device are implemented in code living
in the trees datafile.

### DIFF
--- a/deploy/packaging/debian/python.noarch
+++ b/deploy/packaging/debian/python.noarch
@@ -26,6 +26,7 @@
 ./usr/local/mdsplus/mdsobjects/python/tests/dataUnitTest.py
 ./usr/local/mdsplus/mdsobjects/python/tests/dclUnitTest.py
 ./usr/local/mdsplus/mdsobjects/python/tests/devices/TestDevice.py
+./usr/local/mdsplus/mdsobjects/python/tests/devices/TestDeviceCodeIn.py
 ./usr/local/mdsplus/mdsobjects/python/tests/devicesUnitTest.py
 ./usr/local/mdsplus/mdsobjects/python/tests/exceptionUnitTest.py
 ./usr/local/mdsplus/mdsobjects/python/tests/images/mdsplus_logo.jpg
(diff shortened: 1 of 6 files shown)


tfredian:
There are many potential problems with the use of this feature so we should have some discussions as to whether it is a feasible long term solution. Some things to consider:

1) What if the python code used in the device implementation now stored in the tree becomes obsolete or some fixes or enhancements in the python to MDSplus abi break the the device methods or static functions stored in the tree. Do we need some kind of update capabilities and how might they be managed?
2) Do we need some way to cache the device code associated with the nid to avoid rereading and compiling the python code stored with the device instance in the tree?
3) Should we provide a devfun() tdi function to enable the use of a function implemented in the code in the tree instance? i.e. devfun(mydevice,'fun_name',args...) so that the device could store an expression which used a function stored in the device instance.

tfredian:
PR test failed apparently due to some py3 compatibility issues. I'll explore.

zack-vii:
Some comments to
1. No we dont. even if the device should either not play a role anymore as data would be acquired already and if the data is required for reading the data and may break although we intend to keep the api compatible it was bad by design or can easily be fixed by writing the content f the new version into the device node. depending on how we want to store the data i would suggest as argument of the conglum object that is stored in the head node.

2. managing the caching we could define a either nid to name or device name to name convention and "import" the code as module. a second load would simply check for the module first and only load if it is not available. Overhead would be comparable.. alternatively on could hash the source on store and use that hash for module name. we would store that hash along in the device node and would be able to only load the driver once for multiple copies in the tree or even different trees of the same session.

3. for this we should reuse TreeDoMethod and let the internal code decide how to load the source. but as i said the method should not do any hacks so it will work as long as methods work in general, other than that: anything that can be expressed as tdi expressions with preferably build ins should be stored in the respective node as expression. Complex, otherwise external tdi.fun however, would qualify for this solution.

:: all in all before we add features i think we should start with a simple how to read the source from the tree. I would not make it devices specific but optional for any python device. merely the read method getDevice would have to be able to understand the difference. I will work on an implementations some time this week. depending on if the final waves are ok.

zack-vii:
replaced by #1544

PR #1507, opened 2019-01-16
Zck maxloc

### LINKED ISSUE #1506: reported that maxloc is broken in apha
"The issues resulting from the latest MDSplus update just keep on appearing... Earlier this morning Jonathan discovered that your ip_004.fun doesn't work in the new version (7.49-0) of MDSplus. I did a bit of debugging and it seems that the TDI function MAXLOC is producing rubbish: it always returns a float very close to zero (e.g., 2964.393875047479D-324 in case of plasma current Rogowski measurement in pulse #5157). This causes ip_004.fun to pick the wrong time window."

I checked the maxloc documentation, and in the spirit of perfect documentation, I was surprised that the value return is 0..N for the columns and 1..N for the rows. That is probably normal, but the unsuspecting user might not assume it, so a mention in the documentation file might help. On the other hand, the documentation might already be saying that since I am not certain of the senses of extent or offset vector.

>> Otto's issue

Here's a little test to demonstrate the issue I'm talking about. The following is what I get on my Windows laptop (MDSplus 7.7.9):
TDI> _test = [1.,2.,3.,4.,3.,2.,1.]
[1.,2.,3.,4.,3.,2.,1.]
TDI> maxloc(_test)
3
TDI>

That just shows that MAXLOC used to work as expected. However, on Smaug (MDSplus 7.49-0.el7):
TDI> _test = [1.,2.,3.,4.,3.,2.,1.]
[1.,2.,3.,4.,3.,2.,1.]
TDI> maxloc(_test)
4203.9E-48
TDI>

zack-vii:
Fix: fixes #1506


tfredian:
retest this please

tfredian (approved):

Issue #1514, opened 2019-01-18
Race condition in distributed client configuration


GabrieleManduchi:
In the current alpha version a race condition arised creating several malfunctions and crashes in out current experiment configuration
After investigating the problem  (thanks Cesare!!!) and discovered that at line 860 of treeshr/RemoteAccess.c the macro definition of IO_LOCK and IO_UNLOCK have been commented, therefore not using anymore the lock around the routines defined there.  When we uncommented that macro definition, all the problems disappeared. Therefore the problem is very likely related to some race condition that has been uncovered removing those locks.

zack-vii:
I deactivated the lock as it did not seem to protect anything.
Yet to discover is what the lock actually protects and if it is the cure or if the real problem is somewhere else.
@GabrieleManduchi: Can you somewhat describe what you are doing when the issue is triggered. Can't be a simple read or write or dispatch as that would have been caught be the tests, helgrind at least.

GabrieleManduchi:
The configuration that exposes the issue is a separate thread that writes on different nodes in distributed client configuration, concurrently with other data access programs. The effect is neat, so I strongly suspect that that lock is actually protecting something. In any case I will investigate further.

GabrieleManduchi:
Looking at the routines that are protected via IO_LOCK macro, the number of routines recursively called is high and it is very difficult to state for sure that there are nowhere static structures that are not thread safe.
In any case, that lock protects several routines that call MdsIoRequest(). This routine sends a message over a mdsip connection and reads the response. If two MdsIoRequest() sharing the same connection id are interlaced (recall that two threads can share the same tree object) than you get unexpected responses. No surprise that in most cases the malfunctions occurred in in GetAnswerInfoTS().

zack-vii:
But if this is the case the lock is simply in the wrong place.
First step would be to stick the lock into  MdsIoRequest.
eventually, a individual lock for each connection would be better so multiple servers can be accessed in parallel.

GabrieleManduchi:
I confirm that protecting MdsIoRequest alone resolves the issue. Let me know if you want we make  the fix and issue a PR.
HOWEVER , UNLESS an exhaustive check of all the called routines has been done (I fear this is very difficult in practice, look at  what happened now) AND a measurable improvement in performance is demonstrated, I feel a bit uncomfortable removing the other locks. As you have seen these are the worst errors to be detected and may happen after years of successful operation.

Timo, Tom, Josh, Andrea please advice!

PR #1516, opened 2019-01-20
Fix: move lock into MdsIoRequest;TODO: make threadsafe Send/Receive

zack-vii:
this addresses issue #1514
* We should add a test that faces the issue
* we should secure the mdsip send and receive using msg id to protect remote io
  otherwise different implementations would still interfere, e.g. python has its own lock


GabrieleManduchi (approved):
It seems ok to me

PR #1525, opened 2019-01-26
Build: Avoid need to copy python to build dir

tfredian:
The Makefiles were modified to copy all the python sources
into the build directory so the version checking of the python
modules would not break the tests if _version.py was not in the
python source directory tree. This causes problems with debugging
and testing changes to the python code in the source directory.
Instead this change adds an environment variable MDS_DIS_VCHECK
which can be used to disable the warning message done by the
version checking code which was causing test failures simply
because the output of the tests did not match expected output.


tfredian:
retest this please

zack-vii (approved):
in pr #1524 i added a similar env called MDSPLUS_CHECK_VERSION in pyhton that allows a user to disable the warning by setting this env to no off of 0, which will suppress the check and the warning in case the user decides to use customized python code.
I suggest merging this pr. and i will rebase mine onto of this one but change the env name to MDSPLUS_CHECK_VERSION and discard the __init__.py of this pr

Issue #1519, opened 2019-01-22
Failures reading data in 7.49-0.el7 (stable)


oasunta:
I'll start by apologizing that this explanation is rather lengthy. It's because I wanted to try to test as much as I could to narrow down the problem (I'm not sure how well that worked out in the end, though...). 

We have a problem in 7.49.-0 with nodes that used to work in 7.7.-4. For testing, take two nodes:

(1)
TDI> mag.rog.mct:i
Build_With_Units(Build_Param(Build_Signal(.MAG.ROG.MCT:I_GAIN * (GETNCI(FIND_ANALOGUE_SIGNAL("I_ROG_MCT", "ain"), "record") - .MAG.ROG.MCT:I_DRIFT * (DIM_OF(GETNCI(FIND_ANALOGUE_SIGNAL("I_ROG_MCT", "ain"), "record")) - .002) - .MAG.ROG.MCT:I_OFFSET), *, DIM_OF(GETNCI(FIND_ANALOGUE_SIGNAL("I_ROG_MCT", "ain"), "record"))), "CURRENT IN THE TOP MC COIL AND CASE", *), "A")

(2)
TDI> psu.bvl:i
Build_With_Units(Build_Param(21810.3 * COMPILE(find_analogue_signal("I_BVL_PSU", "ain")) / 10., "PSU current", *), "A")

I can always read the TDI strings that are stored in the nodes. The problem arises when trying to retrieve data(). Both nodes call find_analogue_signal(), a function that simply finds the name of the node where the raw data is, but the rest is just references to other nodes in the tree and standard TDI.

In TDI, if I read data in node (1) first, everything works:

TDI> treeopen('st40',5157);
TDI> data(mag.rog.mct:i)
Set_Range(211000,-50.98459301688293D0 /*** etc. ***/)
TDI> data(psu.bvl:i)
Set_Range(211000,44.7968 /*** etc. ***/)

If I try to read node (2) first, TDI freezes.

TDI> treeopen('st40',5157);
TDI> data(psu.bvl:i)
^C

The fact that getting data from (2) works after getting data from (1), but not before, makes it look as if the TDI function would somehow have a memory.

I also thought that maybe COMPILE(find_analogue_signal("I_BVL_PSU", "ain")) could be the problem because that was one of the differences between the nodes. If I execute the external function call on its own first, the COMPILE() works.

TDI> treeopen('st40',5157);
TDI> find_analogue_signal("I_BVL_PSU", "ain")
[...]

PR #1530, opened 2019-01-30
Fixes: (part of) #1519 for stable

zack-vii:
This should resolve part of the problems in #1519 
* also updated the python tests so they have the same coverage as alpha
* commented out tests of not yet supported features


zack-vii:
retest this please

tfredian (approved):

PR #1533, opened 2019-01-31
wave8: python3.7 GIL improvements

zack-vii:
* improves but sadly not fixes locking problem on fc29


zack-vii:
funny that the now correct fc29 image has a setup (i think its python version) that does not clean up well after a thread cancel.

zack-vii:
merged as part of wave 9 #1534

PR #1543, opened 2019-02-05
Build: rpmsign not prompting for pass phrase

tfredian:
On more recent fedora releases, the rpmsign command no longer
prompts for a pass phrase if the sign key does not require a pass phrase.
The build process used to expect the prompt and provide an empty line
response but with the new rpmsign the packages are being signed but there
are misleading error messages generated. The builds were completing ok
with signed packages but this change should prevent those error messages.


zack-vii (approved):
looks good

PR #1544, opened 2019-02-05
zck: added optional feature include_source

zack-vii:
* allows to embed any python device (from a single source) in the device head node as Conglum filed qualifier via switch add_source in Device.Add()
* change default behaviour via env PYDEVICE_ADD_SOURCE to any other than no or 0
* the module is loaded once per session and version control in case of multi version per session is ensured via a hash value that is stored in the name field.
* Device has a new hidden method called _update_source in case the code can be updated without changing the tree structure (node characteristic); if possible we encourage the user to remove the old device and add a clean device build.
* device method check_source() returns unified diff between source in record and file
* dclUnitTest shows how to use the new tools


tfredian (approved):
Nice job. Much cleaner than my attempt.

Issue #1548, opened 2019-02-07
Different values of MAX_DIM


zack-vii:
Should we limit the number of dimensions to 7 (the number supported by the mdsip protocol)
Currently its 8 but i have seen higher numbers in different libs before.. (stable should still have them)

joshStillerman:
I know that we have been telling people 8 for many years.  I doubt that 
anyone has used 8 though.  If it is easy to change mdsip to accommodate 
8, that would be better than making it 7 everywhere.

zack-vii:
Actually teh message looks like this.. so the fill value would be the eighth dim no sweat, right? 
```c
typedef struct _msghdr {
  int msglen bits32;
  int status bits32;
  short length bits16;
  unsigned char nargs;
  unsigned char descriptor_idx;
  unsigned char message_id;
  unsigned char dtype;
  signed char client_type;
  unsigned char ndims;
#if defined(__CRAY) || defined(CRAY)
  long dims[(MAX_DIMS_R + 1) / 2];
#else
  int dims[MAX_DIMS_R];
  int fill;
#endif
} MsgHdr;
```

zack-vii:
I will make a PR so we can test it for different client/server version combinations

zack-vii:
tested it with tditest and our old version server mds-data-2, looks like the server sets ndims to 8 if 8 is teh number of dims however it can only set 7 dims so the client can reconstruct the 8th dim from the message length if it is a multiple of the numbytes value upto the 7th dim. now you can even read 8dim array from old servers.

zack-vii:
MAX_DIM := 8

PR #1552, opened 2019-02-07
Fix: MAX_DIMS 8 for mdsip; make use of fill value

zack-vii:
addresses issue #1548

### DIFF
--- a/include/ipdesc.h
+++ b/include/ipdesc.h
@@ -9,10 +9,7 @@
 typedef int SOCKET;
 #define INVALID_SOCKET -1
 #endif
-#ifdef MAX_DIMS
- #undef MAX_DIMS
-#endif
-#define MAX_DIMS_R 7
+#define MAX_DIMS 8
 #define DTYPE_UCHAR   2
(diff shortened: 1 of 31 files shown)


joshStillerman (commented):
Can you write a test for this ?  local 8 dims ?  remote 8 dims ?  I would be surprised if anyone has every made an 8 dimensional object. 

I also wonder why remote was explicitly 7 dims

zack-vii:
will do but looks like i have to do some more code replacements before it even matters
will try to grap zero([1,1,1,1,1,1,1,1],1) in MdsConnectionTest

zack-vii:
7 wonders, 7 dwarfs, 7 dims .. maybe maybe not.

zack-vii:
tested it with tditest and our old version server mds-data-2, looks like the server sets ndims to 8 if 8 is teh number of dims however it can only set 7 dims so the client can reconstruct the 8th dim from the message length if it is a multiple of the numbytes value upto the 7th dim. now you can even read 8dim array from old servers.

joshStillerman (approved):

PR #1558, opened 2019-02-12
Fix: remote access disconnect host_list

zack-vii:


zack-vii:
So could someone approve this?

joshStillerman (approved):

PR #1561, opened 2019-02-13
Fix: add timeout and return not triggered on timeout

santorofer:
The device acq435st has been modified to deal with trigger timeout (including error handling): it will return Not Trigged on timeout.


joshStillerman (commented):
please return 1
try to trim lines to ~ 80 chars

joshStillerman (commented):
Why did every line change ?

joshStillerman (approved):
Note that for devices that are not brand new - the ones we don't know all of the users of 
we do not add nodes / change the tree structure.  
this is because, if people have models with these devices in them changing the node list will break their models.

In this case it is fine, since we know exactly who is using this device

zack-vii:
FYI: python device methods should  not return any status code .. please return nothing on success and raise an mdsplus exception on error.
returning an int is interpreted as status value only for backward compatibility and deprecated.

joshStillerman:
OK -

   Fernando will change it back.

-Josh

Issue #1563, opened 2019-02-15
Threads started by devices can loose their tree contexts


joshStillerman:
The streaming acq devices create a thread on init to handle the stream.  When these are dispatched to the same action server the stream can find that self.tree is either not defined or not open (not sure yet).  

This is not a problem with a simple tcl script that inits the devices in a row.  There must be some non thread-safe handling to tree contexts. 

It is possible that this problem is not new, and we are just seeing it now that we are doing 'streaming devices' which fire up threads.  In the past most devices like this (cameras etc...) ran on dedicated servers which would have masked this problem

zack-vii:
maybe we can add something like #1569 to Device class that would simplify this procedure

joshStillerman:
I thought I had it: PR #1571 seemed to work.  However I am finding intermittent behavior when the actions are run by an mdsip server.  With the code as committed
```
do /meth node init
```
works fine.  While 
``` 
dispatch node:init_action
```
The server spits out (sometimes)
```
[jas@cmodws102 pydevices]$ mdsip -s -p 9999 -h /etc/mdsip.hosts 
Cannot setuid - run server as root!Fri Feb 22 16:25:42 2019 (7) (pid 14514) Connection received from jas@cmodws102.psfc.mit.edu [198.125.183.102]
Connected: jas
Fri Feb 22 16:25:43 2019, Doing \TEST::TOP:J:INIT_ACTION in TEST shot 3
Warning:
  The MDSplus python module version (7.62.7) does not match
  the version of the installed MDSplus libraries (1.0.0).
  Upgrade the module using the mdsplus/mdsobjects/python directory of the
  MDSplus installation.
doing \TEST::TOP:J(ACQ435ST).INIT()
Python error in ACQ435ST.INIT:
'ACQ435ST' object has no attribute 'copy'

Traceback (most recent call last):
  File "/usr/local/mdsplus/tdi/dev_support/PyDoMethod.py", line 58, in PyDoMethod
    return domethod(methodobj,args)
  File "/usr/local/mdsplus/tdi/dev_support/PyDoMethod.py", line 33, in domethod
    return methodobj(*args)
  File "/home/jas/mdsplus/pydevices/HtsDevices/acq435st.py", line 161, in init
    thread = self.Worker(self)
  File "/home/jas/mdsplus/pydevices/HtsDevices/acq435st.py", line 102, in __init__
    self.dev = dev.copy()
[...]

PR #1564, opened 2019-02-17
Fix: change from threads to subprocess

joshStillerman:
**NOTE:**  This changes the conglomerate for this device.  All instances
must be removed and re added to the tree

Threads were stepping on each others device instances.  Which still
needs to be address as per: #1563

This code is better regardless.
Instead of creating a thread

- write a script in tmp
- execute it
    - capturing output in a log file
    - recording in nodes
        - script name
        - logfile name
        - pid of creating process
    - change stream to a method used by the script above
        - add additional stream exit condition: parent pid is unavailable
        - regardless of reason(stopped, timed-out, max-segments, parent-dead, on exit, record contents of output in node remove the files

**NOTE:**  Timo's comment that  methods do not need to return 1 is not quite right.  If they don't then the following happens when using do/method
```
TCL> do /meth j stop
doing \TEST::TOP:J(ACQ435ST).stop()
Error executing device method
Error message was: %SS-W-SUCCESS, Success
mdsdcl: --> failed on line 'do /meth j stop'
```

### DIFF
--- a/pydevices/HtsDevices/acq435st.py
+++ b/pydevices/HtsDevices/acq435st.py
@@ -28,12 +28,6 @@ import MDSplus
 import tempfile
 import threading
 
-def threaded(fn):
-    def wrapper(*args, **kwargs):
-        thread = threading.Thread(target=fn, args=args, kwargs=kwargs)
-        thread.start()
-    return wrapper
-
 class ACQ435ST(MDSplus.Device):
     """
     D-Tacq ACQ435 Digitizer real time streaming support.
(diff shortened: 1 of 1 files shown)


zack-vii (approved):
I assume you know what you are doing. looks clumsy. Would fixing #1563 help? As long as it helps.

joshStillerman:
it would help.... but i think having separate process is a win. I would 
love for it to be less clumsy, but did not want to keep any of the info 
in memory. I had trouble capturing the output of the threads, to put in 
either a file or the tree.  This solves all of those problems.  Do you 
have a suggestion for it to be less clumsy? This, at least, does work.

-Josh

joshStillerman:
I have decided (after Timo's comment) to at least move all of the data about the running process into a dictionary stored in the running node.  This is still ugly but not quite as ugly.  That will be a new PR

PR #1569, opened 2019-02-20
Feature: added persistent property to Device for device private stati…

zack-vii:
…cs, e.g. threads

maybe this helps for #1563. and other devices

### DIFF
--- a/mdsobjects/python/tests/dclUnitTest.py
+++ b/mdsobjects/python/tests/dclUnitTest.py
@@ -141,6 +141,10 @@ class Tests(_UnitTest.TreeTests,_UnitTest.MdsIp):
         pytree.readonly()
         self.assertTrue(pytree.TESTDEVICE_I.INIT1_DONE.record <= pytree.TESTDEVICE_I.INIT2_DONE.record)
         self.assertTrue(pytree.TESTDEVICE_S.INIT1_DONE.record <= pytree.TESTDEVICE_S.INIT2_DONE.record)
+        self.assertTrue(pytree.TESTDEVICE_I.INIT2_DONE.record <= pytree.TESTDEVICE_I.STORE_DONE.record)
+        self.assertTrue(pytree.TESTDEVICE_S.INIT2_DONE.record <= pytree.TESTDEVICE_S.STORE_DONE.record)
+        self.assertEqual(pytree.TESTDEVICE_I.DATA.data().tolist(),[0,1,2,3,4,5,6,7,8,9])
+        self.assertEqual(pytree.TESTDEVICE_S.DATA.data().tolist(),[0,1,2,3,4,5,6,7,8,9])
 
     def timeout(self,full=False):
         def test_timeout(c,expr,to):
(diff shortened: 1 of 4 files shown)


tfredian:
hmmm... Not sure if this makes much sense or what is causing the issue in the first place. If the tree associated with the device is getting garbage collected somewhere (which it shouldn't if the device instance is being used for doing a method) then your _unique method wouldn't work anyway. I think we have to fully understand the problem described by the issue before coming up with a solution. I seem to recall we did a lot of work convincing ourselves that the garbage collection handling was all working now and we jumped through hoops to avoid recursive references that prevent garbage collection of trees from being executed properly.

zack-vii:
no the GC collects the instance and hence the reference to the thread. The static field _persistent is tied to the Device class and remains referenced until the end of the session. it is up to the device implementation to open the tree in the thread. I would recommend to ensure private context however i made this unnecessary for most dbid stack operations (TreeCtxPush/Pop set the private context on its own). I just know that the static field workaround is required. 

Since the TdiDoMethod calls into PyDoMethod from c the device node argument is always a new instance and hence there is no way to communicate between two phases/tasks without the usage of statics. I used it on many devices and Czgnet4k has a terrible implementation of this for the same reason.

zack-vii:
To the second part: If I am not mistaken, the GC critical stuff i got rid of when I eliminated the _TreeCtx class.

Which only applies to alpha, though.

tfredian:
Still not fully understanding... So if you generate a thread to execute a method which implies the method already has a tree context since a device is a subclass of tree node. Would that tree context not persist anyway and if not is it up to the method to reopen using self.tree=Tree(self.tree.tree,self.tree.shot)??? If self.tree got closed in another thread then the _unique method would stop working since it couldn't get tree name and shot (unless that is nolonger tied to the tree context). Also what happens if multiple threads are executing device methods on the same device? Should the _unique method include a thread id in the key names? Perhaps if you included a reimplementation of one of your devices using it in the PR it would be clearer why this is needed and how one might use it.

zack-vii:
called from tdi the only context is the global context . Tree() no args. so the context is the same if expt and shot did not change. however the instances of the first call will become unreachable unless they are tied to a static. unreachable objects will be collected afaik. i dont know how the thread will behave,,, if it will get canceled if it depends on its daemon state. all i know is, that the persistent filed will simplify device implementations. maybe not in this case but its worth a try.  nonetheless this pr is useful for some of our local devices.

zack-vii:
its possibly worth setting up a task pair that addresses this issue in dclUnitTest. i will see what i can do tomorrow

zack-vii:
OK I added a use cast to TestDevice and one more useful method 'copy' which simplifies the creation of a thread-safe device ref.
@joshStillerman: How to apply this to your device .. Its probably better to reset it to use threads first.

zack-vii:
retest this please

tfredian (approved):
ok, This makes much more sense to me after the addition of copy.

PR #1571, opened 2019-02-21
Fix: acq435st back from process to thread

joshStillerman:
Looks like the problems in: #1563 were related to garbage collection not to a race condition.

This commit removes the process stuff and uses a defined thread class to do the work.

I (Tom's idea) do not think there is a need to store the device and or thread persistently in the class.  It may be good enough for the thread to make a copy of the device.

Note:  It is sufficient for the init of the thread to copy its argument, because the init is done synchronously before the thread is started, so 'self' is still in scope during its execution

Note: Note: capturing stdout from the thread is problematic as all the threads in a process share the same 
FD table.  The solution for this to use a logger - which can be implemented later.  For now the output is mixed in with the main process's output.

### DIFF
--- a/pydevices/HtsDevices/acq435st.py
+++ b/pydevices/HtsDevices/acq435st.py
@@ -90,8 +90,20 @@ class ACQ435ST(MDSplus.Device):
         parts.append({'path':':INPUT_%2.2d:OFFSET'%(i+1,),'type':'NUMERIC', 'value':1, 'options':('no_write_shot')})
     del i
 
+    class Worker(threading.Thread):
+        """An async worker should be a proper class
+           This ensures that the methods remian in memory
+           It should at least take one argument: teh device node  
+        """
+        def __init__(self,dev):
+            super(ACQ435ST.Worker,self).__init__(name=dev.path)
+            # make a thread safe copy of the device node with a non-global context
+            self.dev = dev.copy()
+        def run(self):
+            self.dev.tree.normal()
+            self.dev.stream()
+
     debug=None
-    data_socket = -1
 
     trig_types=[ 'hard', 'soft', 'automatic']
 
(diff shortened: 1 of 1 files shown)


pydevices/HtsDevices/acq435st.py
```diff
@@ -90,8 +90,20 @@ class ACQ435ST(MDSplus.Device):
...
+        """An async worker should be a proper class
+           This ensures that the methods remian in memory
+           It should at least take one argument: teh device node  
+        """
+        def __init__(self,dev):
+            super(ACQ435ST.Worker,self).__init__(name=dev.path)
+            # make a thread safe copy of the device node with a non-global context
+            self.dev = dev.copy()
+        def run(self):
+            self.dev.tree.normal()
```
zack-vii: self.dev.tree.normal() is redundant as copy opens the tree in normal mode

pydevices/HtsDevices/acq435st.py
```diff
@@ -145,32 +157,9 @@ def init(self):
...
-                                stdout=logfd, 
-                                stderr=logfd).pid
-        if self.debugging():
-            print('The streaming pid is %d' % pid)
-            print('Parent Pid is %d' % os.getpid())
-        fd.close()
-        logfd.close()
-        self.running.record = rundata
+        thread = self.Worker(self)
+        thread.start()
```
zack-vii, line 161: dont you want to register in with the persistent filed? dont you want to do error handling in deinit/store?

pydevices/HtsDevices/acq435st.py
```diff
@@ -145,32 +157,9 @@ def init(self):
...
         return 1
```
zack-vii, line 162: No need to return a status code. simply raise an exception if not ok

joshStillerman: I saw this comment of yours about Fernando's PR last week.  Turns out do /meth in TCL is a bit unhappy if you do not return something odd.  
This could probably be fixed there but for now, I want to leave it.

tfredian (approved):

PR #1579, opened 2019-02-24
Fix:  acq2106_423_Nst devices to acq435 threads

joshStillerman:
In order to test the threading behavior under realistic conditions
we need at least 2 devices.  The acq435st was changed to use a
thread that copies the device object.

Issue: #1563 outlines some potential problems along these lines.

This commit changes to acq2106 to match the new acq435 code so they
can be tested on the SPARC network.

### DIFF
--- a/pydevices/HtsDevices/acq2106_423_1st.py
+++ b/pydevices/HtsDevices/acq2106_423_1st.py
@@ -62,3 +62,6 @@ class ACQ2106_423_1ST(Acq2106_423st):
                      'type':'NUMERIC', 
                      'value':1, 
                      'options':('no_write_shot')})
+    def __getnewargs__(self):#this line
+        return (ACQ2106_423_1ST.__str__(self),)
+
--- a/pydevices/HtsDevices/acq2106_423_2st.py
+++ b/pydevices/HtsDevices/acq2106_423_2st.py
@@ -62,3 +62,6 @@ class ACQ2106_423_2ST(Acq2106_423st):
                      'type':'NUMERIC', 
                      'value':1, 
                      'options':('no_write_shot')})
+    def __getnewargs__(self):#this line
+        return (ACQ2106_423_2ST.__str__(self),)
+
--- a/pydevices/HtsDevices/acq2106_423_3st.py
+++ b/pydevices/HtsDevices/acq2106_423_3st.py
@@ -62,3 +62,6 @@ class ACQ2106_423_3ST(Acq2106_423st):
                      'type':'NUMERIC', 
                      'value':1, 
                      'options':('no_write_shot')})
+    def __getnewargs__(self):#this line
+        return (ACQ2106_423_3ST.__str__(self),)
+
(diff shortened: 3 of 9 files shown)


zack-vii:
This is exactly why i don't like Devices to be shipped with MDSplus packages. device development seems to rely on commits to alpha before they can be properly tested. I would advice updating MDSplus on all involved servers to newest alpha and add the device with the Device.Add(tree,path,add_source=True).
if not already the case combine the carrier and module files into one. This way you can make changes to the device driver update the source in the node and you can be sure it will use the right code. once everything works as expected you can still PR the final version. Would this not speed up the development phase?
I read this in a mail:
1. There is supposed to be a threading issue if you have two devices. there should never be a race condition as long as you work on different nodes. the key of the static dict contains trename:shot:head_nid and consequently each device will have its own private entry.
2. TreeNode.copy() creates a new database that is invisible to other threads. whenever it requires the global context to be changed e.g. to do old tdi calls it activates PrivateCtx first so it should be threadsafe.
Using Worker(self.copy()) is redundant as the __init__ method of the Worker is executed in the main thread.

tfredian:
I don't completely understand how putting in the device code in the node improves the ability to test new device implementations. You certainly don't need to commit the device code to alpha to easily test new code. You simply use MDS_PYDEVICE_PATH to point to your development directory to test new implementations.

joshStillerman:
This last commit addresses the questions that Fernando and I sent earlier. Adding __getnewargs__ to the concrete classes enabled copy to call new successfully.  Tested with do /meth.  Will test with dispatch shortly.

zack-vii:
__getnewargs__ does not work the way you think maybe. you told it to instanciate the copy with teh path, i.e. str(self), as argument which would use the global context agian.

The copy method that we need is not the builtin copy method but the one defined for MDSplus.Device.

since your process does not seem to find it i assume you need to update mdsplus-alpha on the worker running mdsip and restart the mdsip process to reload the MDSplus package.


> 
> 
> This last commit addresses the questions that Fernando and I sent earlier. Adding **getnewargs** to the concrete classes enabled copy to call new successfully. Tested with do /meth. Will test with dispatch shortly.

zack-vii (changes requested):
revert this commit as the approach is unnecessary and implementation wrong. see comment

zack-vii:
please make sure the mdsip action server is running 
    alpha_release-7-63-0
check with MDSplus.\_\_version\_\_
to check if the action server 'ACTIONSERVER' is loading the right version use
```tdi
_host_port = getenv("ACTIONSERVER")
mdsconnect(_host_port)
mdsvalue('py("v=MDSplus.__version__","v")')
```

zack-vii:
You broke the copy method by overwriting with from copy import copy:
- list() will create a copy
- lets not have one file per device if they depend on a super class. if you with to use the PyDevice functionality lets rename the py file to acq2106_423.py and make PyDevice check teh py file if module name startswith filename.

zack-vii:
Please check if PR #1592 satisfies your needs.

PR #1589, opened 2019-03-01
fix: exec-prefix must be set to ${prefix} if NONE

AndreaRigoni:
If user provides --prefix argument the exec-prefix by default must be set to that prefix too

We inadvertently added this bug with Fernando setting exec-prefix before prefix. The fix is simply a swap between the two set lines. In addition the if clause was moved to AS_IF macro, and also I fired the bootstrap command using the Timo docker approach in order to cleanup all the Makefiles.in


AndreaRigoni:
retest this please

GabrieleManduchi (approved):

PR #1594, opened 2019-03-05
Zck pydevices

zack-vii:
get more devices out of one pydevice module file


tfredian (approved):
Looks ok as far as I can tell. I assumed this is tested during pr builds.

PR #1599, opened 2019-03-06
Zck TreeDoMethodA avoids MDS_END_ARG

zack-vii:
- tcl_do_method did not handle new python device methods which return None on success


zack-vii:
retest this please

tfredian (approved):

PR #1600, opened 2019-03-06
Zck python test cleanup

zack-vii:
- only pyhton Tests improvements.

### DIFF
--- a/conf/valgrind-mdsplus.supp
+++ b/conf/valgrind-mdsplus.supp
@@ -120,3 +120,19 @@
    fun:allocate_stack
    fun:pthread_create*
 }
+
+{
+   dont complain about loaded libs in loadLib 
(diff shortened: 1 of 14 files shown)


tfredian:
retest this please

tfredian (dismissed):
Interesting. My fix was a whole lot easier but perhaps this will handle more cases of misuse.

zack-vii:
since SendArg and GetAnswerInfo belong together they should handle the lock in a special way.. in particular they should abort and return an error if the connection has been marked for disconnection,
SendArg should lock from the first message and for special mds_io messages. otherwise it should verify the CON_SENDARG state. GetAnswerInfo does the latter and unlocks.

joshStillerman (approved):

PR #1605, opened 2019-03-10
Zck python setup

zack-vii:
refactor of windows installer

more options and some fixes

### DIFF
--- a/deploy/packaging/windows/create_installer.sh
+++ b/deploy/packaging/windows/create_installer.sh
@@ -39,7 +39,7 @@ fi
 #fi
 
 pushd ${MDSPLUS_DIR}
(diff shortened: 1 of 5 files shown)


tfredian (dismissed):
Complicated! Looks ok but hopefully you will fully test the Windows installer before merging!

zack-vii:
I think it is good now. this MUI2 update look nice and allows for descriptions for each component.
Also the add and remove from *_path envs works now. It now removes teh entry from the list instead of removing the entire env. also it has a memory of the last install dir is the installer is run as update.
Also i took the liberty to remove old versions of Labview as they are not maintained anymore and can be downloaded from github or are already install thru an older version of MDSplus.

tfredian (approved):

PR #1607, opened 2019-03-11
Fix: Ident should keep its treeref

### LINKED ISSUE #1606: Problem with tdi ident in python not retaining context
A user analyzing MDSplus data ran into another tree context issue when using python to access data stored in an MDSplus tree. In this case the node he was accessing contained tdi expressions which referenced other nodes in the tree and used tdi variables:

```
_SIG = .GAS.MKS:B_BOT_CAL * .GAS.MKS:B_BOT_RAW, _DAT = [_SIG], _DIM = DIM_OF(_SIG), Build_With_Units(Build_Signal(_DAT - SUM(_DAT[0 : 99 : *]) / 100, *, _DIM), "mtorr")
```
The python used to access this node does the following:

```
>>> time=tree.getNode(('\\b_bot_mks')).dim_of().data()
```

This results in a TREE-W-NOT_OPEN error. Since the dim_of() method is returning the _DIM tdi variable it appears that idents defined in a node expression are not retaining the tree context.

There is a quick fix:

```
>>> time = pressure.tdiExecute('data(dim_of(\\b_bot_mks))').data()
```

but I believe the previous code in the analysis script used to work and this context issue should be fixed if possible.

zack-vii:
should fix #1606

### DIFF
--- a/mdsobjects/python/mdsscalar.py
+++ b/mdsobjects/python/mdsscalar.py
@@ -440,7 +440,7 @@ class Pointer(Scalar):
         return cls(value.value,is64)
 _dsc.addDtypeToClass(Pointer)
 
-class Ident(_dat.Data):
+class Ident(_dat.TreeRef,_dat.Data):
     """Reference to MDSplus Ken Variable"""
     dtype_id=191
     def __init__(self,name):


tfredian (approved):
That was easy! ;-)

PR #1610, opened 2019-03-12
Fix: A commit message content can break the release tag jaon

tfredian:
The release tag sent to github is a jason which contains a list
of the new commits added to make the release. If this contains
certain charactors it might break the jason format which prevents
the release from being tagged. This first tries with the comments
and if that fails will just send the tag with a comment indicating
the changes could not be described in the tag info.


zack-vii (approved):
It's a valid workaround

zack-vii:
retest this please

PR #1614, opened 2019-03-15
Build: alpine packaging fixes for 3.6 [stable]

zack-vii:
… build script from wiping git source folder


zack-vii:
retest this please

deploy/packaging/alpine/build_apks.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+pckdir=$(dirname $(realpath $0))
+cd /workspace
+signkeys="/sign_keys/mdsplus@mdsplus.org-589e05b6.rsa"
```
WhoBrokeTheBuild, line 5: If we're hardcoding the path to the key here, why don't we just download it here as well?
The Dockerfile calls wget to download the key, we could instead add a check here for the file and download it if it doesn't exist, that way there wouldn't be any hardcoded file paths between the image and the build scripts

deploy/packaging/alpine/build_apks.sh
```diff
@@ -0,0 +1,23 @@
...
+if [ -f $signkeys ]
+then
+  mkdir -p ~/.abuild
+  echo -e "PACKAGER_PRIVKEY=\"$signkeys\"\n" > ~/.abuild/abuild.conf
+else
+  abuild-keygen -aqn # generate new keys
+fi
+# we have to unset srcdir as it is a reserved name in abuild
+# in the process of building the apks it will wipe $srcdir
+srcdir= \
```
WhoBrokeTheBuild, line 17: I'm confused by the \ here, are those intentional?

deploy/packaging/alpine/build_apks.sh
zack-vii: you cannot instlal the key wiothout sudo rigjhts and in jekins you are not necessarely sudo

deploy/packaging/alpine/build_apks.sh
zack-vii: it ties the lines together as if there were no new line.
```sh
A=a B=a cmd
```
wil run cmd with envA set to a and env B set to b without modifying the env of the caller

tfredian (approved):

PR #1626, opened 2019-03-21
Zck python wsgi

zack-vii:
* added template scripts and conf files for an easy setup with apache2.
* fixed most of wsgi's tree context problems.
* moves js and html under wsgi.
* integrated mdsplus_wsgi.py in wsgi.\_\_init\_\_.py

from MDSplus.wsgi import application


joshStillerman:
Is this ready for test/review ?  You keep pushing commits to it

zack-vii:
Oh yes it is.. just working thru more improvements. I wont be pushing more then so this can be milestone.

joshStillerman (approved):
This looks great !  thanks!  I look forward to trying it all out.  I think we should look to web based GUis for a lot of things which will require good web based APIs

PR #1627, opened 2019-03-22
Adding increment function to python MDSplus.Tree

WhoBrokeTheBuild:
Adding `MDSplus.Tree.increment(treename)` function to mirror functionality from `set current treename/increment`

### DIFF
--- a/mdsobjects/python/tree.py
+++ b/mdsobjects/python/tree.py
@@ -874,6 +874,16 @@ class Tree(object):
             raise _exc.MDSplusException(status)
         return _scr.Uint64(dt.value).date
 
+    @staticmethod
+    def increment(treename):
+        """Increment the current shot number for the specified treename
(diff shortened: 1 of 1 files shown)


joshStillerman (commented):
How about 
tree.increment() too ?  (instance method)

joshStillerman:
Yep sounds good

mdsobjects/python/tree.py
```diff
@@ -874,6 +874,16 @@ def getVersionDate():
             raise _exc.MDSplusException(status)
         return _scr.Uint64(dt.value).date
 
+    @staticmethod
```
zack-vii, line 877: I have a way of implementing methods that can be used as static and as instance base on the context it is used. check out ther usage of @classmethodX used by e.g. compressDatafile

PR #1628, opened 2019-03-22
Feature: make set/getCurrent classstaticX and added incrementCurrent

zack-vii:
reversion of PR #1627

### DIFF
--- a/mdsobjects/python/tree.py
+++ b/mdsobjects/python/tree.py
@@ -2456,6 +2456,12 @@ class TreeNode(_dat.TreeRef,_dat.Data): # HINT: TreeNode begin  (maybe subclass 
         """
         self.compress_segments=flag
 
+    def setDefault(self):
+        """Set compress segments state of this node
+        @rtype: None
+        """
+        self.tree.setDefault(self)
+
     def setDoNotCompress(self,flag):
         """Set do not compress state of this node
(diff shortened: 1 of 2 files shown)


mdsobjects/python/tree.py
```diff
@@ -874,6 +876,22 @@ def getVersionDate():
...
+        @param tree: Name of tree
+        @type tree: str
+        @param inc: Increment (default: 1)
+        @return shot: New current shot number for the specified tree
+        @rtype int
+        """
+        if isinstance(self,(Tree,)): # instancemethod: args shifted by one
+            if tree is not None: inc = tree
+            tree = self.tree
+        shot = Tree.getCurrent(tree)+inc
```
WhoBrokeTheBuild, line 891: You can't use `Tree.getCurrent(tree)` here because if there is no current shot it returns an exception, you need to instead use `TreeGetCurrentShotId` (which will return 0) or add an exception handler

WhoBrokeTheBuild (changes requested):

mdsobjects/python/tree.py
zack-vii: but also if you hav no current you cannot increment it so the exception is justified, no?

WhoBrokeTheBuild (approved):

PR #1629, opened 2019-03-22
Added reader/writer threads with queues to ACQ423

WhoBrokeTheBuild:
ACQ423 devices now spawn an MDSWorker thread to write data to MDS, and a DeviceWorker thread to read data from the device.
The workers create and recycle buffers as needed using two queues.


joshStillerman:
retest this please

joshStillerman (approved):

PR #1631, opened 2019-03-26
Zck python; refactor ScopeUtilities

zack-vii:
might still have memory issues

### DIFF
--- a/include/dtypedef.h
+++ b/include/dtypedef.h
@@ -2,7 +2,7 @@
  *	atomic data types 0 - 37, 51-55
  */
 DEFINE(MISSING		,0)	/* no data */
-DEFINE(V_depricated	,1)	/* aligned bit string */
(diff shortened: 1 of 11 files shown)


zack-vii:
this might be it

tfredian (approved):
I don't know much about jScope related features so can't really tell if this will break anything. I'll assume it is ok and approved it but perhaps Gabriele should probably review this soon before merging it.

zack-vii:
please review followup #1634 instead

GabrieleManduchi:
The changes to ScopeUtilities seem reasonable, but it is hard for me to verify in the code due to the large number of modifications.
Let's cross fingers. I will make some more extensive tests in order to make sure that nothing got broken.

zack-vii:
cool thanks. if there is anything different now it should be easier to adapt though.

PR #1636, opened 2019-03-28
Fixed acq423 time duplication bug by truncating calculations

WhoBrokeTheBuild:
The time was accumulating due to MDSplus rounding to a precision of 3, and python rounding to a much higher precision, leading to time values being duplicated at the end/beginning of segments.


joshStillerman:
Ok to test

joshStillerman (approved):
This should be further looked at to see why it is needed.  For now, it works so lets do it.

PR #1638, opened 2019-03-29
Fix: added average resample mode to xtreeshr and made it default

zack-vii:
This one is tricky.. the default resampling method in xtreeshr is worthless if used for evaluation and has not real use (at least i cannot think of any non artificial) This PR will change the default resampling method to moving average.

### DIFF
--- a/mdsmisc/ScopeUtilities.c
+++ b/mdsmisc/ScopeUtilities.c
@@ -177,13 +177,13 @@ inline static int64_t estimateNumSamples(const mdsdsc_t *const dsc, mdsdsc_t *co
   int nextRow, segmentSamples, numActSegments, segmentIdx;
   EMPTYXD(xd);
   int nid = recIsSegmented(dsc);
-  if(!nid) return -1;
+  if(!nid) goto return_neg1;
   int status = TreeGetNumSegments(nid, &numSegments);
-  if STATUS_NOT_OK return -1;
+  if STATUS_NOT_OK goto return_neg1;
   if(xMin != NULL || xMax != NULL) {
     if(xMin) XTreeConvertToLongTime(xMin, &startTime);
     if(xMax) XTreeConvertToLongTime(xMax,   &endTime);
-    if(numSegments < NUM_SEGMENTS_THRESHOLD) return -1;
+    if(numSegments < NUM_SEGMENTS_THRESHOLD) goto return_neg1;
     startIdx = 0; //If no start time specified, take all initial segments
     if(xMin) {
       while(startIdx < numSegments) {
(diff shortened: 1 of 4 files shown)


zack-vii:
can we vote on this one.. its changing the default behavior but i don't think this is an issue.

tfredian:
I think Gabriele will need to review this to see if this might break existing user codes etc.. Gabriele did most if not all of the xtreeshr implementation.

zack-vii:
we probably should add some api to set the resample and squish mode without extended Attributes. 
maybe as additional args to an extended XTreeSetTimeContext(start,end,delta,squish,resample).

zack-vii:
The original default sampling mode was interpolation which makes only sense when oversampling, i.e. when delta_requested is smaller than delta_timebase. however, judged by the sloppiness of the output timebase (nodes are processed segment by segments which leads to inconsistencies at the segment edges.) I assume xtreeshr is meant to be used rather for data reduction than for conversion of timebases of different signals.

zack-vii:
added some tests for the other modes. they all seem to work now.

zack-vii:
retest this please

zack-vii:
I added an env that allows to set the default resampling mode. Maybe we can have another less permanent way to set the params. Maybe a way that is private and does not involve writing to the tree. Maybe we can improve the handling of segment bounds and make the reduction available for non segmented nodes as well. Maybe for this we should make the data reduction implementation more transparent and predictable. Maybe we should be working (optional) with (double or native) timestamps. But that would be somewhere in the future.

zack-vii:
@GabrieleManduchi : could you comment on this please

GabrieleManduchi:
A detailed check is very difficult due to the large amount of changed code. In any case the code appears to be improved, but I would recommend to avoid the definition of macros for segments of code and using functions instead, leading to the same effect in compacting the code, but allowing a better debug. Debugging a code with macros for pieces of code is a nightmare!

Pleas fix also the bug I have reported on slack:

--------------------------------------------------------------------------------------------
Found the bug. It happens when the signal is not a segmented one (at least on my use case, I think you have tested on segmented signals)
in estimateNumSamples() of ScopeUtilities.c, when recIsSegmented() returns 0 (as happens with non segmented signals), you must set the out arguments dMin and dMax before returning -1
In my code I changes as follows:
int nid = recIsSegmented(dsc);
 if(!nid)
 {
     const double xmin = to_doublex(xMin->pointer,xMin->dtype,-INFINITY,TRUE);
     const double xmax = to_doublex(xMax->pointer,xMax->dtype, INFINITY,TRUE);
     *dMin = xmin;
     *dMax = xmax;

   return -1;
 }
and it works.

zack-vii:
well in this case the macro is the only way to avoid code duplication because tge only code change is the data type. in case we want to debug something i would recommend to expand the code beforehand just for this purpose. i will make the code change right after this meeting

GabrieleManduchi:
My concern was for  INTERPOLATION_FORLOOP(type), but, yes, you are right. This is a C and not a C++ program for which templates would fit.
After you have reported the fix I will approve the PR and after merging I will make some further tests on jScope dynamic resampling.

zack-vii:
i extended your mdsmisc fix by evaluating and setting min max for any -1 result

GabrieleManduchi (approved):

GabrieleManduchi:
There are still several problems, look at slack

GabrieleManduchi:
Ok, after a sequence of 5 bugs my test case seems to work. 
Please refer to slack for the list of bug fixes, or let me know if you want I make a PR

PR #1643, opened 2019-04-06
Fix: use double to resample to increase range

zack-vii:


GabrieleManduchi:
From my side it seems to work. Of course we need to test it a while.

GabrieleManduchi (approved):

PR #1648, opened 2019-04-10
Zck cleaup spaces

zack-vii:
cleanup space mess using 
sed -i 's/\s*$//' $1
sed -i 's/\(^\|\t\)\x20\{1,7\}\t/\1\t/g' $1
sed -i 's/\(^\|\t\)\x20\{8\}/\1\t/g' $1


zack-vii:
should we add tools that help keeping the code clean:

e.g. python code should not contain tabs for py2/3 compatibility etc.

GabrieleManduchi (approved):
Looking at 1301 files changed is a bit scaring.........
I guess you did it with some automated tool......

Issue #1652, opened 2019-04-12
mdsip non compatibility between stable and alpha


GabrieleManduchi:
It appears that in distributed client configuration (I did not check thin client) stable clients cannot write data remotely using alpha mdsip server.  I have not checked but I fear that the problem in the opposite configuration  (alpha clientm, stable mdsip server) is still there.

This represents a  problem for large systems: we already experienced this in our lab. After upgrading mdsip server, all the clients not updated to alpha were broken.

PR #1653, opened 2019-04-12
Fix: restore compatibility with stable

zack-vii:
According to http://man7.org/linux/man-pages/man2/write.2.html

On Linux, write() (and similar system calls) will transfer at most
0x7ffff000 (2,147,479,552) bytes, returning the number of bytes
actually transferred.  (This is true on both 32-bit and 64-bit
systems.)

we dont need to return int64_t


zack-vii:
this should fix #1652

joshStillerman (approved):

PR #1659, opened 2019-04-22
Build: Repair repository creation on redhat and alpine

tfredian:
Since the MDSplus repositories were moved to a nfs share exported
by a new nfs server there were locking errors running the createrepo
command. A workaround was added for the publish operation to use
a /tmp directory for the repodata. This workaround was not implemented
for the release operation but is necessary unless another nfs server is
used to serve the release directory tree. To avoid this need for using
a separate nfs server the same /tmp workaround will now be used for both
the /release and /publish repositories.

The use of two nfs server mounts also can cause problems if some of
the jenkins slaves fail to mount one of these mounts which can cause
problems if the release build and the publish build do not occur on the
same slave. With the above workaround the build and publish diretories
will by served by the same nfs server.

There was also a bug in the alpine repository creation script which
caused each alpine platform build to rsync the entire repository
directory from release to publish instead of just rsync'ing the repository
specific to the architecture being built. This should fix that problem
too.

### DIFF
--- a/deploy/platform/alpine/alpine_docker_build.sh
+++ b/deploy/platform/alpine/alpine_docker_build.sh
@@ -55,5 +55,6 @@ buildrelease() {
 }
 
 publish() {
-    rsync -a /release/${BRANCH} /publish/
+    mkdir -p /publish/${BRANCH}
+    rsync -a /release/${BRANCH}/${ARCH} /publish/${BRANCH}/
 }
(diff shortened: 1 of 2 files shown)


zack-vii (approved):

tfredian:
retest this please

tfredian:
retest this please

PR #1670, opened 2019-04-29
Fix: Segfault in get_segment_limits

### LINKED ISSUE #1669: Bug in get_segment_limits
"I have found a problem in treeshr get_segment_limits (TreeSegments.c) when called for a segment  with int64 start and end definition, with the first argument retStart set to 0 (a legal combination when only retEnd is requested).
In this case the program crashes. Looking in the code it appears that in lines:"

```
if (retStart) {
      q_d.pointer = (char *)&timestamp;
      loadint64(&timestamp,buffer);
      MdsCopyDxXd(&q_d, retStart);
    }
    if (retEnd) {
      const int filled_rows = get_filled_rows_ts(&vars->shead,vars->sinfo,vars->idx,(int64_t*)buffer);
      if (filled_rows > 0) {
        loadint64(&timestamp,buffer + (filled_rows-1) * sizeof(int64_t));
        MdsCopyDxXd(&q_d, retEnd);   --------------------------------------------> HERE q_d is not initializedin case retStart == 0
      } else
        MdsFree1Dx(retEnd, 0);
    }
```
"apparently line         q_d.pointer =(char *)&timestamp;
is missing"
Reported by @GabrieleManduchi

tfredian:
Fixes #1669

### DIFF
--- a/treeshr/TreeSegments.c
+++ b/treeshr/TreeSegments.c
@@ -1446,16 +1446,17 @@ static int read_segment(void* dbid, TREE_INFO * tinfo, int nid, SEGMENT_HEADER* 
 
 static int get_segment_limits(vars_t* vars, struct descriptor_xd *retStart, struct descriptor_xd *retEnd){
   INIT_TREESUCCESS;
-  struct descriptor q_d = { 8, DTYPE_Q, CLASS_S, 0 };
+  int64_t timestamp=0;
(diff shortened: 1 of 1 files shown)


zack-vii (approved):
Would have been sufficient to swap line 1469 and 1470, but hey this should work too.

PR #1674, opened 2019-04-30
Fix:Conversion to double in ScopeUtilities

GabrieleManduchi:
Removed the division by 1e9 if DTYPE_Q or DTYPE_UQ to be converted, that broke jScope dynamic resampling for absolute times. No assumption must be made when time is expressed in 64bit long format (EPICS Time counts nanoseconds, UTC time counts milliseconds), so the value (converted to double) must be returned as it is.


AndreaRigoni (approved):
ok !

PR #1676, opened 2019-04-30
Feature: Add error reporting in SendMdsMsg

tfredian:
This will add some error reporting for failures occurring when
send messages to mdsip servers. This should provide more information
when diagnosing mdsip communication errors.


GabrieleManduchi:
Tom, if I am not wrong, if for any reason send returns always 0 the program enters an infinite loop because in this case tries is never increased.

tfredian:
I changed the code to only reset tries to 0 if send returns > 0.

GabrieleManduchi:
Yes, but it seems to me that if send keeps on returning 0, tries is never increased......am I wrong?

tfredian:
oops. ok I think I got it now.

GabrieleManduchi (approved):

PR #1678, opened 2019-05-06
Feature: Add default_tree_path

### LINKED ISSUE #1667: Add default_tree_path configuration option
Add a `default_tree_path` configuration option / environment variable. This would then be used with the current logic as such:
```
if env[tree + '_path']
    path = env[tree + '_path']
else if env['default_tree_path']
    path = env[tree + '_path']
else
   error tree + '_path or default_tree_path not found'
```

The immediate use case is to configure a docker container with:
```
export default_tree_path="/trees/~t/"
```
Which would allow any tree to be mounted under `/trees` with little to no configuration.

tfredian:
You will now be able to define a default tree path using the
default_tree_path environment variable. For example defining

export default_tree_path=/trees/~

You could do something like:

$ mkdir /trees/mytree1
$ mkdir /trees/mytree2
$ mdstcl
TCL> edit/new mytree1
...
TCL> edit/new mytree2
...

without needing mytree1_path and mytree2_path environment variables.

fixes https://github.com/MDSplus/mdsplus/issues/1667

### DIFF
--- a/treeshr/TreeOpen.c
+++ b/treeshr/TreeOpen.c
@@ -137,6 +137,8 @@ EXPORT char *TreePath(char const *tree, char *tree_lower_out){
   if (tree_lower_out)
     strcpy(tree_lower_out, tree_lower);
   path = TranslateLogical(pathname);
+  if (path == NULL)
+    path = TranslateLogical(TREE_DEFAULT_PATH);
   if (path) {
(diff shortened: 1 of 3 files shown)


WhoBrokeTheBuild (approved):
Looks great!

PR #1687, opened 2019-05-15
Fix: Obey deny rules in mdsip.hosts

### LINKED ISSUE #1684: MDSip is not obeying the ! (denial) syntax anymore
It was always possible to deny standard mdsip connections by providing a exclamation point at the beginning of a line in mdsip.hosts followed by a match string. When a client connects the lines in mdsip.hosts are checked starting from the top until a matching rule is found. If a match is found on a line starting with an exclamation point the connection should be denied. It no longer appears to be doing that. An easy test is to begin mdsip.hosts with a line containing "!*" which should deny any connections but does not.

This was reported by a site which runs stable releases so it must have been broken for a while.

tfredian:
Lines beginning with exclamation points should deny access if
the connection matches the match string after the exclamation point.

This should fix: https://github.com/MDSplus/mdsplus/issues/1684

### DIFF
--- a/mdstcpip/CheckClient.c
+++ b/mdstcpip/CheckClient.c
@@ -113,7 +113,6 @@ int CheckClient(char *username, int num, char **matchString)
   if (strncmp(hostfile, "TDI", 3)) {
     FILE *f = fopen(hostfile, "r");
     static int zero = 0, one = 1;
-    static unsigned short two = 2;
     if (f) {
       static char line_c[1024];
       static struct descriptor line_d = { 0, DTYPE_T, CLASS_S, line_c };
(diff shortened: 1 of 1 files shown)


zack-vii (approved):
looks like the previous code tried to do the same but it should have uses one for the offset instead of two.

tfredian:
Actually the two was correct (a ported VMS routine). The problem was that the match string was stripped of the leading exclamation point in the for loop (not the while) so it could then match subsequent checks without the exclamation point which could then be accepted. I originally put a break after the ok=2 but if the deny match string didn't match it would try the next in the for loop comparing it to the string after the exclamation point was removed and if that matched it would also approve the connection. I didn't explore back to the older code to see how that used to be handled but manually testing the new code seemed to fix it.

PR #1688, opened 2019-05-15
Fix: Obey deny rules in mdsip.hosts

tfredian:
Lines beginning with exclamation points should deny access if
the connection matches the match string after the exclamation point.

This should fix: https://github.com/MDSplus/mdsplus/issues/1684


zack-vii (approved):
see #1687

PR #1690, opened 2019-05-20
Fix: TreeShr hook funs not working

tfredian:
The feature of enabling treeshr hooks as tdi funs was only working
when compiled without optimization. The code was declaring structures
within if else blocks and referencing those declarations later and
the optimizer was reusing the memory of those structures before
they were used. This commit provides the hook feature with
optimization enabled.

### DIFF
--- a/treeshr/TreeCallHook.c
+++ b/treeshr/TreeCallHook.c
@@ -89,7 +89,11 @@ static int isEnabled(char *hookName) {
 }
 
 void TreeCallHookFun(char *hookType, char *hookName, ...) {
-  if (!isEnabled(hookName)) return;
+  static int (*TdiExecute)() = NULL;
+  int status;
+  status = LibFindImageSymbol_C("TdiShr", "TdiExecute", &TdiExecute);
+  if (!(isEnabled(hookName) && STATUS_OK)) return;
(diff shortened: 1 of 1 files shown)


treeshr/TreeCallHook.c
```diff
@@ -89,7 +89,11 @@ static int isEnabled(char *hookName) {
 }
 
 void TreeCallHookFun(char *hookType, char *hookName, ...) {
-  if (!isEnabled(hookName)) return;
+  static int (*TdiExecute)() = NULL;
+  int status;
+  status = LibFindImageSymbol_C("TdiShr", "TdiExecute", &TdiExecute);
+  if (!(isEnabled(hookName) && STATUS_OK)) return;
```
joshStillerman, line 95: Insert:
if ((status & 1) == 0) return;

joshStillerman (changes requested):
I added 1 suggestion

treeshr/TreeCallHook.c
tfredian: That's what STATUS_OK does. (See Timo's include/status.h)

joshStillerman (approved):

PR #1694, opened 2019-05-22
Gm streaming

GabrieleManduchi:
Added method stream() to python class Event
Minor fixes to node.js server and to the html page

### DIFF
--- a/mdsobjects/python/event.py
+++ b/mdsobjects/python/event.py
@@ -213,6 +213,46 @@ the e.join() would return exiting the problem.
         _exc.checkStatus(_MdsShr.MDSQueueEvent(_ver.tobytes(event),_C.pointer(eventid)))
         return eventid.value
 
+    @staticmethod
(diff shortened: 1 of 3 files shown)


tfredian:
Would it be too complicated to send MDSplus serialized data in the STREAMING event? If the consumer could deserialize the data it would be much more efficient (network and cpu) than converting the data to strings, sending, and converting back. The STREAMING event could even send a dictionary or list object containing the signal name, shot, times and values.

tfredian:
The advantage of sending serialized data would depend highly on the typical size of streaming data packets.

tfredian (approved):
This is fine unless sending serialized data is a better option.

GabrieleManduchi:
The MDS event payload is going to be received by a JavaScript program (Node.js). That's why the data packed is converted in ASCII. In any case the packet is composed of one or few samples since the streaming is for visualization

PR #1695, opened 2019-05-22
Fix: mdsmisc.GetXYSignalXd set TimeContext based on SegLimit, isLong?

zack-vii:
It Dimensions are stored as int64 delta need to be converted to sec in order to get correct data.

### DIFF
--- a/mdsmisc/ScopeUtilities.c
+++ b/mdsmisc/ScopeUtilities.c
@@ -647,12 +647,13 @@ EXPORT int GetXYSignalXd(mdsdsc_t *const inY, mdsdsc_t *const inX, mdsdsc_t *con
       delta = estimatedSegmentSamples/10.;
     //In any case don't make it too big in respect of the single segment size (at minimum 10 samples )pairs) per segment
     delta *= (estimatedDuration/estimatedSamples);
-    if (isLong) delta /= 1e9;
+    if (isLong) delta /= 1e9; // quick compensation of xtreeshr conversion
     deltaP = (delta>1e-9) ? &deltaD : NULL;
   } else deltaP = NULL;
   //Set limits if any
   int status = TreeSetTimeContext(xMinP,xMaxP,deltaP);
   if STATUS_OK status = TdiEvaluate(inY, &yXd MDS_END_ARG);
+  TreeSetTimeContext(NULL,NULL,NULL); // reset timecontext
   if STATUS_NOT_OK goto return_err;
   // Get Y, title, and yLabel, if any
   EMPTYXD(title);EMPTYXD(xLabel);EMPTYXD(yLabel);
--- a/mdsmisc/ScopeUtilities.c
+++ b/mdsmisc/ScopeUtilities.c
@@ -524,11 +524,11 @@ inline static int getNSamples(const mdsdsc_xd_t *const yXd, const mdsdsc_xd_t *c
   if(yXd->pointer->class != CLASS_A) return TdiINVCLADSC;
   if(xXd->pointer->class != CLASS_A) return TdiINVCLADSC;
   if(yXd->pointer->dtype == DTYPE_F) {
-    const int status = TdiFloat((mdsdsc_t *)yXd, yXd MDS_END_ARG);
+    const int status = TdiFloat(yXd->pointer, yXd MDS_END_ARG);
     if STATUS_NOT_OK return status;
   }
   if(xXd->pointer->dtype == DTYPE_F) {
-    const int status = TdiFloat((mdsdsc_t *)xXd, xXd MDS_END_ARG);
+    const int status = TdiFloat(xXd->pointer, xXd MDS_END_ARG);
     if STATUS_NOT_OK return status;
   }
   //Number of samples set to minimum between X and Y
--- a/mdsmisc/ScopeUtilities.c
+++ b/mdsmisc/ScopeUtilities.c
@@ -166,7 +166,7 @@ static int recIsSegmented(const mdsdsc_t *const dsc) {
 **/
 
 #define MAX64 0x7FFFFFFFFFFFFFFFL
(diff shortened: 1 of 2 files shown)


zack-vii:
retest this please

GabrieleManduchi:
You are assuming that in case long is used, the time unit is nanosecond. This is not true, indeed we often use int64 to store UTC time expressed in milliseconds.

GabrieleManduchi:
When we decided to use double representation, we got rid of the implicit conversion of all times in nanoseconds unit, and therefore now every time is a number related to the chosen units, so no division / multiplication for 1e9 must be done.

zack-vii:
this is just because in xtreeshr we do the same. i would be happy to get rid of it in xtreeshr but this would break current behaviour. bottomline is we need to treat int64 in mdsmisc thesame way as we do in xtreeshr.

GabrieleManduchi:
Ok, but what xtreeshr does is totally transparent outside XTreeGetTimedRecord.  
In ScopeUtilities.c,  estimateNumSamples does an estimation of the number of samples based on the actual delta value, that is totally unrelated to the way XTreeGetTimedRecord works. 
By the way, I remember to have  already removed that division by 1e9 some time ago since it broke jScope when zooming into waveformes with absolute UTC time.

GabrieleManduchi:
We may later think about a different implementation of XTreeshr without the implicit conversion to nanoseconds, but this is unrelated to ScopeUtilities

GabrieleManduchi:
...and if I am not wrong you already worked out the xtreeshr stuff.....

GabrieleManduchi:
...ok, but without removing the 1e9 stuff....

zack-vii:
hmm well the connection with xtree is duevto thebsettimecontext. and yes you are right i need to fix min and max too. the 1e9 in mdsmisc is only to compensate the 1e9 in xtreeshr. the output shoult then be in tge same data type and scale as the source.

GabrieleManduchi:
But, again, you should disregard completely the 1e9 stuff outside the XTreeGetTimedRecord() internals.

zack-vii:
min max is not affected as it will use NULL instead. 
again if mdsmisc want to make use of settimecontext we need to treat long differently. we can use a dtype_q delta or compensate the 1e9. as i said the output is not affected.

GabrieleManduchi:
Maybe I am missing something.....why should I be concerned about 1E9  even if I use SetTimeContext passing int64 times? In normal use cases int64 is used to represent absolute times that may have ns units (EPICS times) or ms unit (UTC times). In any case the passed min and max to SetTimeContext will be consistent with the signals being read.  I don'care If they are internally multiplied /divided by 1E9 in xtreeshr, what I am interested in is that signals read afterwords (of course with times consistent to what has been passed to SetTimeContext) return only the portion within min and max.
In other words, no time unit is assumed by SetTimeContext and XTreeGetTimedRecord().

zack-vii:
What you are missing is that the delta is calculated based on the segment Limits with would be a direct cast of int64 to double. Before mdsmisc put delta in settimecontext as is (direct cast as double not int64). xtreeshr well interpret the as-is delta in seconds. It will also notice the data's dim is in int64 and interpret it as nanoseconds. Consequently it will convert the one in the other and introduce the factor 1e9 to delta. To make it compare as is units with as is units we can compensate delta by dividing 1e9.

GabrieleManduchi:
Ah, I see.... if I understand, you are in this way computing delta in double. In case segment limits were given in int64  (i.e.  jScope is dealing with absolute times) there is a misinterpretation by xtreeshr of the information stored is SetTimeContext. Why not converting also delta in int64 (from the computed double value) in this case (absolute times)  instead of exposing the internal xtreeshr 1E9 stuff in mdsmisc?

zack-vii:
Basically we could do that but dividing delta by 1e9 is a one-liner while changing the datatype of delta would require some more elements. At minimum you would need to change deltaP to use DTYPE_Q and point to a int64.

GabrieleManduchi:
I think that  changing  deltaP to use DTYPE_Q and point to a int64 when isLong is returned true by estimateNumSamples()  is much better, even if it requires some lines more.

In this case the xtreeshr internals are not exposed elsewhere and if we decide to change anything inside xtreeshr there are no side effects.

zack-vii:
maybe, but if we change xtreeshr we can as well delete the line. in fact, we would want to get rid of the entire extra handling anyway. so we need to touch the file anyway.

GabrieleManduchi:
"""""maybe, but if we change xtreeshr we can as well delete the line. in fact, we would want to get rid of the entire extra handling anyway. so we need to touch the fike anyway"""""

This is exactly what I want to avoid!!  In few months no one will remember this and in case xtreeshr will be changed sometimes in future,  jScope will break and it will take time to figure out what happened. Another thing is getting rid of the entire extra handling. If we'll do it after changing xtreeshr, good, we have cleaned code, if we forget or decide to postpone it, nothing gets broken anyway.

zack-vii:
well that would be the changelog:
```c
-    if (isLong) delta /= 1e9; // quick compensation of xtreeshr conversion
-    deltaP = (delta>1e-9) ? &deltaD : NULL;
+    // since xtreeshr has a special handling for int64 timestamps we need to address this here
+    // one way would be to compensate the 1e9 factor or we pass delta as DTYPE_Q
+    if (isLong) {
+      if (delta >= 1.5) {
+        deltaP = &deltaD;
+        *(int64_t*)&delta = (int64_t)(delta+.5);
+        deltaD.dtype = DTYPE_Q;
+      } else deltaP = NULL;
+    } else
+      deltaP = (delta>1e-9) ? &deltaD : NULL;
```
i would rather add a test that will alarm us and add a comment to xtresshr that will remind us. but I think it is unlikely that we will change xtreeshr. It suites us well with combined dimensions in double sec and int64 ns plotted in on scope.

GabrieleManduchi (approved):
Ok, what is important is to make sure hat we don't forget it in future

PR #1699, opened 2019-05-28
Fix:handle the case the expression refers to a segments nid, but chan…

GabrieleManduchi:
…ges timebase

In case the expression being displayed on jScope refers to a segmented node, but re-defines timebase, the mechanism for resampling based on segment information fails. This has been fixed by making recIsSegmented() more conservative so that it gives off if a SIGNAL or FUNCTION descriptor is found when traversion the expression


AndreaRigoni (approved):
Ok

PR #1701, opened 2019-05-28
Fix: removed reference to RELEASE_COMMIT and RELEASE_BRANCH

tfredian:
Now available as GIT_COMMIT and GIT_BRANCH
If a RELEASE_VERSION has not been set by trigger.sh the RELEASE_TAG will be
equal to GIT_TAG. And the GIT_TAG has the unique commit hash inside.


tfredian:
I will eventually clean up the deploy scripts to simplify the git commands used to determine whether a minor or patch release is required and to reduce the number of environment variables passed around to specify release info which is now determined by configure. The trigger jobs could detect a new release and add a local release tag to be used by the build jobs, This will be done in a subsequent pull request.

AndreaRigoni (approved):
Thanks Tom !! Everything seems correctly added.
We should wait for Josh to test it too.

Issue #1704, opened 2019-06-03
Windows installer fails if MDSplus dll's are in use (i.e. mdsip server is enabled)


tfredian:
I may be wrong but I think this used to work. When MDSplus is updated and the system had a running mdsip service the MDSplus dll's being used by that service are not updated. Since the MDSplusVersion mechanism changed mdsdcl cannot find the data structure used in the mdsshr.dll so it fails to start. Similarly the python module cannot find the required structure and complains when importing MDSplus. If the mdsip service is stopped and you deinstall MDSplus and then reinstall it, it seems to fix the problem.

The original scheme which I thought worked was to rename the dll's that were in use since they cannot be removed and then install the new dll's. I did find some renamed dll's in system32 so perhaps that happened as expected but maybe Windows is now using the renamed dll's until all MDSplus processes exit. If this is the case then simply restarting the mdsip service would fix the problem. I did not try this. I don't know what the standard way of dealing with this in windows installers is. If the rename no longer works properly then perhaps the installer needs to detect dll's actively in use and stop the processes using those dll's during the deinstall of MDSplus.

zack-vii:
Well the designed workaround this problem is to setup a postinstall and make the user restart the system. Stopping the mdsip server before installing should also do the trick. What I found was that the installer just fails to move the dlls from the install folder to system32. In that case you need to manually stop the services or all running mdsplus binaries in general and move the remaining files.
If renaming fixes it, it is still a very dirty solution. as those files will likely not be removed by the uninstaller.
we should implement the following
https://blogs.msdn.microsoft.com/windows_installer_team/2005/09/18/insights-into-the-underlying-windows-installer-reboot-behavior/

zack-vii:
Since we alway run uninstall before installing a new version we should force the uninstaller to remove&rename files thoroughly, as intended by microsoft.

zack-vii:
looks like we could make use of the [LockedList](https://nsis.sourceforge.io/LockedList_plug-in) plugin for nsis. It would pop a dialog if files cannot be deleted.

tfredian:
The LockedList plugin looks promising. We could also probably stop/start the mdsip services when updating:

```
sc stop "MDSplus 8000 >nul"
sc stop "MDSplus 8100 >nul"
```
invoke LockedList plugin (perhaps in a loop until nothing is locked or the uninstall is canceled?)
uninstall/remove files
install mdsplus
```
sc start "MDSplus 8000 >nul"
[...]

PR #1705, opened 2019-06-05
Zck win installer

zack-vii:
Windows installer:

* will notify the user if things went wrong in system folders (install and uninstall).
* will properly stop and remove "MDSplus 8000" and "MDSplus 8100" service when uninstalling.
* will rename files that cannot be removed but have been ignored by the user
* renamed file will be scheduled for removal during the next reboot

* fixed a rar mem access issue in xtreeshr


zack-vii:
this addresses issue #1704

tfredian (approved):
looks good

PR #1711, opened 2019-06-12
Feature: Reorganize the MDSplus python module

tfredian:
In the past the MDSplus python module code was in directory
called mdsobjects/python. To facilitate simpler use of the python
module the directory of the module is now python/MDSplus. One
can easily access the module simply by including something like
/usr/local/mdsplus/python in the PYTHONPATH environment variable
and python will look for MDSplus in that directory when importing.
The MDSplus installers will now avoid copying the python code from
/usr/local/mdsplus/python/MDSplus but instead simply create a
symlink to that directory in the python site-packages directory.
This will greatly avoid the problem of the python module getting out
of sync with the MDSplus libraries.


AndreaRigoni:
Hi Tom, I would suggest you to have a look at how setup.py gets the release number if you install from directory.. it seems to try importing mdsplus_version that doesn't exist: 
from mdsplus_version import mdsplus_version
release=mdsplus_version

This resolves to me as having always the same egg package 1.0 installed in site-packages 

Do we prefer this ?

tfredian:
Andrea, actually with these changes there should never be an egg package unless someone uses pip to install the package (mostly for pip --user support). The new setup.py will just install a symlink to /usr/local/mdsplus/python/MDSplus in site-packages. Another possibility would be to remove the setup.py completely which would require the PYTHONPATH env to include /usr/local/mdsplus/python. The problem we're trying to resolve is the case where the version of the python module is out of sync with the installed MDSplus libraries etc. We have users who want to run their own python program which is a different version than the system python programs. The use the setup.py to install in their own python and then when an system wide MDSplus update occurs their installed MDSplus python version gets out of sync. The first attempt to solve this was to do the version comparison and complain when it gets out of sync. This probably caused confusion instead of really solving the problem. Just adding an MDSplus  symlink to the directory containing the python code in site-packages seems to solve the problem but is not a standard practice and when using a tool like pip it is difficult to even determine where it needs to create the symlink. The use of PYTHONPATH would work but some users might define PYTHONPATH in their own scripts without merging in whatever PYTHONPATH was previously defined so the script might omit the MDSplus module. I haven't researched the many options for distutils and/or setuptools to see if there is some better and/or standard way to deal with this. I agree that if we continue to provide a setup.py it should handle the version info properly for cases where it won't just create a symlink. I don't know what problems it might cause but we could just make setup.py print a message to use PYTHONPATH and exit with an error. Any better ideas on how to handle this are certainly welcome. Perhaps a topic of discussion at the MDSplus developers meeting if not solved before then.

tfredian:
I think I may have figured out how to get setup.py to create a .pth file instead of doing the symlink thing. I got it to work with the /usr/local/mdsplus/python directory hardcoded  in the setup.py but will have to change that to try to detect the location of the setup.py and use it's parent directory for the .pth file. I will also fix the release handling in the setup.py and push changes to the PR on Monday hopefully.

tfredian:
Do not merge yet. Ran into a problem with Windows installer with python module.

PR #1712, opened 2019-06-17
Zck mdip stuck recv

zack-vii:
mdsip seems to get stuck in recv when hosting. This replaces the blocking recv with a timeout loop select.
Furthermore the requested username of the client will be store in connection struct for debugging purposes (since the process uid might have mapped it to something else or not map at all).

Finally thick clients wont keep connections open when they become unused. this should improve the situation with dangling connections as a cleanup currently requires the user to trigger disconnect old connection. TODO: if the keep allive feature for unused connections should be required we should spawn a thread cleaning up old connection after some time out if not reuse by then.

### DIFF
--- a/mdstcpip/Connections.c
+++ b/mdstcpip/Connections.c
@@ -201,6 +201,7 @@ void DisconnectConnectionC(Connection *c){
   FreeDescriptors(c);
   free(c->protocol);
   free(c->info_name);
+  free(c->rm_user);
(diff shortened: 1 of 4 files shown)


tfredian:
retest this please

tfredian (approved):

zack-vii:
windows had a hick up last test but succeeded in run .#3560

PR #1713, opened 2019-06-17
Feature: Reorganize python module

tfredian:
The directory structure for the MDSplus python module
is being changed from mdsplus/mdsobjects/python to
mdsplus/python/mdsplus. This enables the use of the module
by simply defining PYTHONPATH=<mdsplus-dir>/python.

The setup.py in the module has been changed to add an mdsplus.pth
file to the python site packages which directs python to reference
the module in the mdsplus/python/MDSplus directory similar to
what changing PYTHONPATH would do.

The windows installer will also compile the python code in the mdsplus/python/MDSplus during
installation.


deploy/packaging/windows/mdsplus.nsi
```diff
@@ -1017,7 +1023,7 @@ Section uninstall
 	Push $R2
 	IfFileExists "$INSTDIR\python\MDSplus\setup.py" 0 skip_python_remove
 	SetOutPath "$INSTDIR\python\MDSplus"
-	nsExec::ExecToLog /OEM /TIMEOUT=10000 'python setup.py remove'
+	nsExec::ExecToLog /OEM /TIMEOUT=10000 'python setup.py remove && del /S *.pyc && del /S __pycache__'
```
zack-vii: I would not do that ... should be part of setup.py remove if at all

deploy/packaging/windows/mdsplus.nsi
```diff
@@ -757,40 +757,46 @@ SectionGroup /e "!APIs" apis
...
   SectionEnd ; python_pp
   Section "run 'python setup.py install'" python_su
 	SetOutPath "$INSTDIR\python\MDSplus"
-	File mdsobjects/python/setup.py
+	File python/MDSplus/setup.py
 	nsExec::Exec /OEM /TIMEOUT=10000 "python setup.py install"
 	Exch $R0
 	Pop  $R0
   SectionEnd ; python_su
+  Section "Compile python" python_comp
```
zack-vii, line 794: If you want to make this optional you might have to add it to the selection logic, handling it same as python_su. 

* You would want to do this for pyhton2 and 3 if both are available.
* It is only required for AllUser installations where the users might not have the chance to store compiled files at runtime.

python/MDSplus/__init__.py
```diff
@@ -86,18 +86,18 @@ class MDSplusVersionInfo(ctypes.Structure):
...
   The MDSplus python module version (%s) does not match
   the version of the installed MDSplus libraries (%s).
-  Upgrade the module using the mdsplus/mdsobjects/python directory of the
-  MDSplus installation.
-''' % (__version__, _ver ))
+  Upgrade the module using the mdsplus/python/MDSplus directory of the
+  MDSplus installation or set PYTHONPATH=/usr/local/mdsplus/python/MDSplus.
+''' % (__version__, _verchk ))
     version_check()
-del version_check, _ver
```
zack-vii, line 100: well the idea was to remove all trace of the version_check for the final module. which is why I del all used vars.. I at some point rename _ver to version to make it more transparent but rebasing or something caused it to scramble it seems.. hte _mimport should write to version and all _ver.* should be version.*

python/MDSplus/setup.py
```diff
@@ -25,82 +25,53 @@
...
+      'name'         : 'MDSplus',
+      'extra_path'    : {'mdsplus',pth_dir},
+      'version'      : release,
+      'description'  : 'MDSplus Python Objectsi - '+release_tag,
       'long_description': """
       This module provides all of the functionality of MDSplus TDI natively in python.
       All of the MDSplus data types such as signal are represented as python classes.
       """,
-      'author'       : 'Tom Fredian,Josh Stillerman,Gabriele Manduchi',
+      'author'       : 'MDSplus DEvelopment Team',
```
zack-vii: typo DE ->De

python/MDSplus/tests/dclUnitTest.py
```diff
@@ -24,6 +24,8 @@
 #
 
 import sys
+import os
```
zack-vii: why, it is not used, right?

zack-vii (changes requested):

deploy/packaging/windows/mdsplus.nsi
zack-vii: I added the logic and the doc string

python/MDSplus/__init__.py
zack-vii: I rename the _ver to version

deploy/packaging/windows/mdsplus.nsi
tfredian: Didn't realize "python setup.py remove" removed the whole directory tree. I took out the del cmds in the nsi.

zack-vii (approved):

PR #1715, opened 2019-06-24
Fix:fix random crashes in Distributed Client

GabrieleManduchi:
When calling remote disconnect, disconnectFromMds was called with wrong id in case a connection whith h.connections == 0 was found in the host list. This can happen (and indeed it happened often in our configuration) if some subtree is missing.

In addition, binary not has been used in place of the wrong logical not  in DisconnectConnections

### DIFF
--- a/mdstcpip/Connections.c
+++ b/mdstcpip/Connections.c
@@ -214,11 +214,11 @@ int DisconnectConnection(int conid){
   if (c) {
     c->state |= CON_DISCONNECT;
     pthread_cond_broadcast(&c->cond);
-    if (c->state & !CON_DISCONNECT) {
+    if (c->state & ~CON_DISCONNECT) {
       struct timespec tp;
       clock_gettime(CLOCK_REALTIME, &tp);
       tp.tv_sec += 10; // wait upto 10 seconds to allow current task to finish
-      while (c->state & !CON_DISCONNECT && pthread_cond_timedwait(&c->cond,&connection_mutex,&tp));
+      while (c->state & ~CON_DISCONNECT && pthread_cond_timedwait(&c->cond,&connection_mutex,&tp));
     }
     // remove after tast is complete
     if (p)      p->next = c->next;
--- a/treeshr/RemoteAccess.c
+++ b/treeshr/RemoteAccess.c
@@ -147,9 +147,9 @@ static int remote_access_disconnect(int conid, int force){
     DBG("Connection %d< %d\n",conid,host->h.connections);
   } else DBG("Disconnected %d\n",conid);
   for (host = host_list; host ;) {
-    if ((force && host->h.conid == conid) || (host->h.connections <= 0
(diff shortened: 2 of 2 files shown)


tfredian:
retest this please

GabrieleManduchi:
retest this please

GabrieleManduchi:
retest this please

mdstcpip/Connections.c
```diff
@@ -214,11 +214,11 @@ int DisconnectConnection(int conid){
   if (c) {
     c->state |= CON_DISCONNECT;
     pthread_cond_broadcast(&c->cond);
-    if (c->state & !CON_DISCONNECT) {
+    if (c->state & ~CON_DISCONNECT) {
```
zack-vii, line 217: should be (!(c->state & CON_DISCONNECT )) instead.

treeshr/RemoteAccess.c
```diff
@@ -147,9 +147,9 @@ static int remote_access_disconnect(int conid, int force){
     DBG("Connection %d< %d\n",conid,host->h.connections);
   } else DBG("Disconnected %d\n",conid);
   for (host = host_list; host ;) {
-    if ((force && host->h.conid == conid) || (host->h.connections <= 0
+    if ((force && host->h.conid == conid) || (host->h.conid == conid && host->h.connections <= 0
```
zack-vii, line 150: the second 'host->h.conid == conid' would eventually leave old connections alive when USE_TIME is true. this second loop seems to only make sense with USE_TIME as it all happens in a HOST_LIST lock anyway.
i will clean it up.

zack-vii (commented):
I think it might be better to file a new PR.
can you briefly summarize when you issue pops up. i recall from some email.. it was threaded writing of segments into different nodes of the same tree mounted using distributed client (host::/path/to/tree/files) ?!

treeshr/RemoteAccess.c
GabrieleManduchi: In any case the previous version was wrong because if any element is found with h.connections == 0, then DisconnectFromMds() was called ***with the wrong conid*** that is the conid passed to remote_access_disconnect().

This was the reason for the error below that drove us crazy for months

Error in SendArg: mode = 1, status = 65554
Error in SendArg: mode = 6, status = 65554
Error in SendArg: mode = 6, status = 65554
Error writing device data: DatabaseException: %TREE-E-FAILURE, Operation NOT successful
null

mdstcpip/Connections.c
GabrieleManduchi: You want to enter the block if any flag bit except CON_DISCONNECT is set
this corresponds to c->state & ~CON_DISCONNECT

if you use !(c->state & CON_DISCONNECT ) the result is true if bit CON_DISCONNECT is false,  I don't' think this is the condition you expected.

zack-vii:
might have been fixed by #1723

PR #1720, opened 2019-06-25
Fix: TCL do/method with args segfault

joshStillerman:
The call to process the arguments in TclDoMethod was passing the
address of a descriptor to TdiIntrisic.  It needed to pass the address
of this pointer because the input argument is declared:
```
struct descriptor *list[]
```


joshStillerman:
Fixes #1720

tfredian (approved):

PR #1721, opened 2019-06-25
Fix: uniform thread and tunnel implementation

zack-vii:


zack-vii:
~Thread formally was  a dirty copy of ~Tunnel to ensure this will not break anything. However this makes it hard to maintain. This solved the problem in a similar manner as the tcp/udt protocol.

tfredian (approved):

PR #1723, opened 2019-06-25
Zck distributed connections

### LINKED ISSUE #1722: Distributed client performance issues
It is unclear what exactly caused this change in behavior but when opening a tree with a path containing multiple remote server connections which has many subtrees with similar tree paths the connections to those servers are opened and closed many times. At MIT the tree paths point to a new shots server, a model tree server and an archived shots server so when opening an older shot a connection is made to the new shots server, tree open fails, connection closed, connection to models sever, tree open fails, connection closed,... Just opening an old cmod tree containing subtrees results in more than 100 connections and closures taking more than 1 minute to open the tree. Using older MDSplus releases the same operation resulted in < 5 connections and took only 1 second. Looking at the older code it seems likely that the intended connection closures were not really happening so it reused the existing connections. In any event, this situation is a very big performance deterioration for distributed client usage.

Josh made a suggestion that we might consider. If we moved the connection cleanup/closure operations to a separate routine we could cleanup connections after TreeOpen operations complete instead of doing it for each file open. In other words, implement some CloseInactiveConnections routine that would be called at the end of TreeOpen*() calls which would close all unused connections. We could also add a time period (perhaps 10 minutes or so) since a connection to a host was in use (i.e. successful file open and close) and only close stale connections. This would reduce the creation of new connections if an application opens and closes many shots sequentially.

zack-vii:
attempt to fix #1722 and issue with distributed client implementation addressed by #1715

### DIFF
--- a/mdstcpip/Connections.c
+++ b/mdstcpip/Connections.c
@@ -224,6 +224,7 @@ int DisconnectConnection(int conid){
       while (c->state & ~CON_DISCONNECT && !pthread_cond_timedwait(&c->cond,&connection_mutex,&tp));
       if (c->state & ~CON_DISCONNECT)
         fprintf(stderr,"DisconnectConnection: Timeout waiting for connection %d state=%d", conid, c->state);
+      c = _FindConnection(conid, &p); // we were waiting, so we need to update p
     }
     // remove after task is complete
     if (p)      p->next = c->next;
--- a/mdstcpip/Connections.c
+++ b/mdstcpip/Connections.c
@@ -91,7 +91,7 @@ Connection *FindConnectionWithLock(int id, con_t state){
   CONNECTIONLIST_LOCK;
   c = _FindConnection(id, NULL);
   if (c) {
-    while (c->state && !(c->state & CON_DISCONNECT)) {
+    while (c->state & ~CON_DISCONNECT) {
       DBG("Connection %02d -- %02x waiting\n",c->id,state);
       pthread_cond_wait(&c->cond,&connection_mutex);
     }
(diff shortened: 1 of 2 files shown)


GabrieleManduchi:
In DisconnectConnection() you want to enter the block if any flag bit except CON_DISCONNECT is set
this corresponds to c->state & ~CON_DISCONNECT, which is different from  !(c->state & CON_DISCONNECT) 

if you use !(c->state & CON_DISCONNECT ) the result is true if bit CON_DISCONNECT is false,

 I don't' think this is the condition you expected.

zack-vii:
well it the condition that I intended today, but I have to overthink if that was the condition I intended back then. 

edit: but you are right the other way would be nonsense given teh c->state |= CON_DISCONNECT

tfredian:
This change improves the performance problem described in #1722 but doesn't do as well as my PR (which I closed because of this PR). With these changes the time for opening of a cmod tree is reduced to 16 seconds with a total of 18 connections. Just setting USE_TIME=60 in current alpha drops the time to open the tree back down to 1 second. Changing the timeout in this pr to 60 seconds did not help.
Also the DEBUG macro is still defined so this PR will cause debug printouts.

If this can't be tweaked to give the 1 second open speed I can reopen my PR to simply enable the 60 second USE_TIME value.

tfredian:
I think you may need to use a time field in the host_t struct. Just having a thread periodically cleaning unused connections does not solve the issue of excessive reconnections to servers.

tfredian:
I put in a commit that makes this work. It now opens that cmod tree in 1 second like it used to and does the host_list_cleanup only every 10 seconds after any connection activity.

tfredian (approved):

PR #1725, opened 2019-06-25
Fix:  reduce forced and reset only if needed

zack-vii:
Adds the idle state of the clean up thread back in.
This prevents the rar but yet possible interference with a dis-/connect in sync with the 10sec timeout.
Also it minimizes the overhead.
Forced way to disconnect now only disconnects conid, honoring the timeout for all other connections.
Connect does only reset the timer but never arms it.
Disconnect only arms the timeout if connection counter requires it and never resets.

### DIFF
--- a/treeshr/RemoteAccess.c
+++ b/treeshr/RemoteAccess.c
@@ -77,20 +77,25 @@ typedef struct host_list{
 } host_list_t;
 
 static host_list_t *host_list = NULL;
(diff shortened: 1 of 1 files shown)


tfredian:
no idea why this is needed

zack-vii:
I think this may improve performance

tfredian (approved):

PR #1728, opened 2019-06-26
Feature: added int64 return values from lib calls and wrapped __ALPHA

zack-vii:
turns out it was not possible to do LIB->CALL:Q()

also this reduces code doubling and skips code that is of no use ifndef __ALPHA

### DIFF
--- a/include/mdsserver.h
+++ b/include/mdsserver.h
@@ -28,11 +28,6 @@ typedef enum { build_table_begin = 1,
   done
 } MonitorMode;
 
-#ifdef __ALPHA
-#pragma member_alignment save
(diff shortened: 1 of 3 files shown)


tfredian:
Hmmm, I just now noticed that you removed the checks for __vms 2 years ago and left the __ALPHA code blocks in place. I believe the __ALPHA checks were there to handle vms systems running OpenVMS on the VAX Alpha hardware. I think all the references to the __ALPHA macro could have been removed when we discontinued vms support so this code could be greatly cleaned up.

zack-vii:
great

tfredian (approved):

PR #1731, opened 2019-06-26
Twf promote stable alpha

tfredian:


zack-vii (approved):
I am a fan of alpha as you know ;)

PR #1735, opened 2019-07-02
Fix: Small fixes for acq435st.py.

seanalsop:
This is my first PR to MDSplus - and was just the changes I needed to make to get an acq1001+acq435 combo to stream. Let me know if there is anything I need to change.

### DIFF
--- a/pydevices/HtsDevices/acq435st.py
+++ b/pydevices/HtsDevices/acq435st.py
@@ -26,8 +26,6 @@ import numpy as np
 import MDSplus
 import threading
 import Queue
-import socket
-import time
 
 try:
     acq400_hapi = __import__('acq400_hapi', globals(), level=1)
@@ -206,6 +204,8 @@ class ACQ435ST(MDSplus.Device):
                 self.running = False
 
             def run(self):
+                import socket
+                import time
                 if self.debug:
                     print("DeviceWorker running")
 
--- a/pydevices/HtsDevices/acq435st.py
+++ b/pydevices/HtsDevices/acq435st.py
@@ -24,7 +24,6 @@
 #
 import numpy as np
 import MDSplus
-from MDSplus import Event,Range
 import threading
 import Queue
 import socket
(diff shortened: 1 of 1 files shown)


joshStillerman (changes requested):
Thank you for this contribution.  A few comments:
1 - since this is a 32 channel device, our standard is to use 2 digit channel number
2 - changing the device structure is problematic because of existing trees (this is mitigated by the fact that devices reference their parts by offset from the head node, so technically changing the node names does not break things)
3 - Many of the changes you made are stylistic
      - moving the imports to the top of the file
      - I personally like them in the functions, especially (but not just) when they are only used in one function
4 - The PR talks about things that were syntax problems.  It is difficult to determine which of the changes are in reference to this and which are stylistic
5 - I do agree on some of the stylistic changes (renaming the worker for example)
I
 am sorry if this seems overly critical, because we do really really welcome your participation in the project.  Please resubmit with:
1 - not moving the imports
2 - not changing the node names
3 - label which changes are:
   Syntactic  corrections
   Semantic corrections
   Stylistic corrections

Finally:  The title must start with 'Fix:' or our revision/release system will complain.

Thanks again!
-Josh

seanalsop:
Hi Josh,

I don't think you're being overly critical at all, I'd like to get this right.

1 - There was a mix of 2 and 3 digit channel numbers in the file. This was throwing
an error. I have changed all of the channel numbers to 2 digits now.

2 - I think the node name changes were the digits in the channel numbers?

3 - The imports actually weren't present at all. The script was error-ing out. I
agree, it's definitely cleaner to have them inside the functions. I've made a
commit to move them into their relevant functions. I have not yet moved Queue.
Since it is used in two functions inside the class it would require an import
and then adding "self." to the module calls. If you want this implemented I am
happy to do so. I had a look for Queue in the git history but couldn't find
anything.

4. Syntax problems were mostly just missing "self." that were throwing errors.

5. Without renaming Worker to MDSWorker it was throwing an error. I changed this
just to get the script to run as I saw that Worker was renamed to MDWorker in an
earlier commit but the call to it wasn't changed.

There was one part that I wanted to highlight - I added a "first = True" inside
the run function. I'm not sure if it is correct for it to be there. Again this
was a change I made just to get the script to stream from the system, so I know
it works, I just wanted to make sure this was okay.

Once you're happy with the changes I'll close this PR and open a new one
starting with "Fix:".

Thanks for taking the time to look through the PR.

joshStillerman (commented):
This is great!  Thanks alot.  We are looking at how the non working code made it into the alpha.  I will merge now.

Josh

joshStillerman (approved):

PR #1743, opened 2019-07-17
Gm add treetdimethods

GabrieleManduchi:
Added Tree methods tdiData(), tsiEvaluate(), tdiExecute(), tdiCompile() in order to avoid the non thread safe usage of setActiveTree. Now C++ interface is on step with the python one.

Added the corresponding test unit.
In addition, the test unit for TreeSegments has been fixed (I wonder if it was actually executed when making tests....) , and the node.js interface for streaming updated

### DIFF
--- a/include/mdsobjects.h
+++ b/include/mdsobjects.h
@@ -267,7 +267,7 @@ public:
     /// Return the result of TDI decompile
     char *decompile();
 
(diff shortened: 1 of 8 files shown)


AndreaRigoni (approved):
seems ok !!

Issue #1655, opened 2019-04-17
Problems and changed behavior in python treenode nci properties


tfredian:
The data types of the nci properties of TreeNode instances in the MDSplus python module has changed between stable and alpha. This has the potential of breaking user code which might depend on nci properties being returned as MDSplus data type instances. It also can cause problems if one tries to use these in python based tdi fun's since sometime they produce errors when trying to convert the nci property back into an MDSplus data type which tdi fun's need to do.

Here is an example of the behavior currently in the alpha branch:

```
>>> n=Tree('mytree',-1).MYNODE
>>> type(n.rfa)
<type 'long'>
>>> type(n.time_inserted)
<type 'int'>
```

but in the stable branch:

```
>>> type(n.rfa)
<class 'MDSplus.mdsscalar.Uint64'>
>>> type(n.time_inserted)
<class 'MDSplus.mdsscalar.Uint64'>
```

If one of these values is returned by a python based tdi fun it is automatically converted to an MDSplus data type but the current alpha has problems doing this:

```
>>> makeData(n.rfa)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python2.7/site-packages/MDSplus/mdsscalar.py", line 108, in __init__
    self._value = self._ntype(value)
OverflowError: Python int too large to convert to C long
```

One of the advantages of nci properties being returned as MDSplus data types is one used to be able to do things such as:

```
>>> print n.time_inserted.date
17-APR-2019 12:33:59.44
```

It is fairly trivial to convert MDSplus data types to native python data types if needed so it is difficult to justify such a change in the API/ABI of the python module. If there is a consensus to keep the new behavior at a minimum the problem above regarding makeData(n.rfa) needs to be addressed.

zack-vii:
I agree on the makeData part can you tell me what rfa is.. a long but what value
the issue lies within int/long conversion int mdsplus types. is this on python2 i assume

tfredian:
The rfa field is defined in treeshrp and is 48 bits so needs to be returned as a 64-bit value. As the error message shows above it is python2.7

zack-vii:
can you give me the return value of n.rfa please

point is if i do the above i have no problem with makeData on python2
the error should only occur if _ver.bit_length(value)<=31 but value still does not fin in int32.. i wonder when that should be.

tfredian:
with stable:
```
>>> n.rfa  
4294966784QU
```
with alpha:
```
[...]

PR #1750, opened 2019-07-24
Fix: cast rfa to Uint64 in python api

zack-vii:
addresses #1655

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -219,7 +219,7 @@ class Nci(object):
     _IS_MEMBER=2
     SET_FLAGS          =( 1,_C.c_uint32,4,int)
     CLEAR_FLAGS        =( 2,_C.c_uint32,4,int)
-    TIME_INSERTED      =( 4,_C.c_uint64,8,int)
+    TIME_INSERTED      =( 4,_C.c_uint64,8,_scr.Uint64)
     OWNER_ID           =( 5,_C.c_uint32,4,int)
     CLASS              =( 6,_C.c_uint8 ,1,int)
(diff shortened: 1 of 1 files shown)


tfredian (commented):
This may solve the problem of rfa causing an error but at least for some of the nci items there are advantages in returning MDSplus scalar types instead of native python types. It may be the only case where this is true is time_inserted which used to return a Uint64 type which has properties such as date and time. One might be able to create a Uint64 instance with the return value but since there is not much else an application might want to do with the time_inserted value there is no reason not to return a Uint64.

zack-vii:
retest this please

zack-vii:
retest this please

zack-vii:
looks like an issue with mdstcpip
http://jenkins2.mdsplus.org/job/fc28/2752/
free(): invalid pointer
 ... again/still possibly when a connection is dropped.

http://jenkins2.mdsplus.org/job/fc28/2752/artifact/tests/64/python/MDSplus/tests/connectionUnitTest.log/*view*/

should not be related to this PR though

zack-vii:
retest this please

pfbuxton:
This issue is also present when one uses the alternative Python method for accessing data:
```python
conn = Connection(MDSplus_IP_address)
conn.openTree(tree,pulseNo)
time_inserted = conn.get("getnci("+place_to_read+",'TIME_INSERTED')").data()
```

Also when accessing data in Matlab.

In all three cases the overflow behaves differently, so a different number is generated.

Thanks,
Peter

joshStillerman:
retest this please

joshStillerman:
retest this please

joshStillerman (approved):

PR #1756, opened 2019-07-31
Modified the Python Tree() class, specifically the dir() method to return the default path to a node and not always the TOP value.

santorofer:


santorofer:
The method dir() belonging to the Tree() class was returning only the value of the TOP node path:

self.top.dir()

Now it will return the value of the default node path:

self.default.dir()

The old behavior is now in a method called dirTop().

zack-vii:
well dir prints from the node it is callen from. and top is the top node. please dont chsnge that. instrad you could add a dir method tho thebtree class that prints from default.

On 31 Jul 2019 18:50, Fernando Santoro <notifications@github.com> wrote:

The method dir() belonging to the Tree() class was returning only the value of the TOP node path:

self.top.dir()

Now it will return the value of the default node path:

self.default.dir()

The old behavior is now in a method called dirTop().

santorofer:
Hi Timo,
Well, the dir() method I modified is in the Tree() class, not the one in the TreeNode(). 
The way Tree() dir() was written was to always print top, am I right? Shouldn't be printing the relative path from the default node (if it is set to one different from TOP)?

zack-vii:
i see, well then at least i woikd avoid topdir as new api since node.tree.top.dir() would do the trick from any node.

santorofer:
Hi Timo, 
I understand, it makes sense. I can remove that new method knowing that we can achieve the same thing "manually".
Cheers!
Fernando

zack-vii (approved):
makes sense this way.

Issue #1760, opened 2019-08-08
Some dispatched action replies are not getting sent   / recieved


joshStillerman:
A user reported that his dispatched actions were hanging about 50% of the time.
I verified that it is exactly 50% of the time (every other)

If you run a program that dispatches an action and exits
and then you run that program again it hangs.

This is because the socket management code in `SendReply` in `servershr\ServerQAction.c` is not working.

Given a script:
```
$ cat t.tcl
set tree lh
create pulse 'P1'
set tree lh /shot = 'P1'
dispatch .HARDWARE.SEND_INIT
$ 
```
and an action:
```
TCL> deco .hardware.send_init
Build_Action(Build_Dispatch(2, "SUBMIT_SERVER", "INIT", 1, ""), LHInitialize($SHOT), "")
TCL>
```
This HANGS
```
[jas@cmodws107 ~]$ mdstcl
TCL> @t.tcl 33
TCL> exit
[jas@cmodws107 ~]$ mdstcl
TCL> @t.tcl 34
......... STUCK .........
```
if the process stays around it works repeatedly
```
[jas@cmodws107 ~]$ mdstcl
TCL> @t.tcl 35
TCL> @t.tcl 36
TCL> @t.tcl 37
TCL>
this is fine
so it only has a problem when it is a new process
```

Some more information:

1.  the treatment of the return values for send with  MSG_DONTWAIT do not seem to be being handled correctly. see: https://developer.ibm.com/tutorials/l-sockpit/
2. the case that the 2nd `send` in SendReply fails was not causing a retry
3. the print statement that the connection was lost was trashing the 1st and 3d terms of the IP address (the debugger showed them to be correct)
4. I don't think this is relevant but is the global variable that holds the list of client sockets thread safe ?

I am positing that AttachPort and RemoveClient in that c file are not working correctly.

I tried removing the DONTWAIT and checking for success on the 2nd Send.
This did not help.

Stopping for now.

zack-vii:
hi josh,
[...]

PR #1762, opened 2019-08-14
Zck server reply

zack-vii:
This should be a cleaner version of pr #1761 and still resolve issue #1760 .

### DIFF
--- a/servershr/ServerQAction.c
+++ b/servershr/ServerQAction.c
@@ -29,6 +29,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <string.h>
 #include <stdio.h>
 #include <time.h>
+#include <errno.h>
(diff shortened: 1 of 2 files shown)


tfredian:
Great find that a send of length 0 detects the broken socket! I guess you need a #pragma to avoid the clobbered warnings if the warning is not really a problem. I wonder if we could also check the client list periodically by issuing a couple 0 length sends to the sockets.

zack-vii:
i think the send 0 length only detects the broken pipe because we send the header first. so looping through tge clients would not work. the clobered is due to the pthread_cleanup eventually i only need to split tge staus in allocation and initialisation. will check it later when maya takes a nap.

tfredian (approved):

tfredian:
Yep.  I experimented with two 0 byte sends on a disconnected socket and it didn't produce an error return. Oh well... I guess it is unlikely the client list would accumulate many disconnected sockets anyway.

PR #1764, opened 2019-08-14
Fix: Restore menu updates for Windows

### LINKED ISSUE #1763: Windows installer not creating MDSplus start menu shortcuts.
I tried installing the latest alpha windows installer (7.84-0) on my Windows 10 computer and there are no MDSplus related items in the Windows start menu. I can create a command prompt and type mdstcl and that works fine so MDSplus is installed ok but the menu items are missing.

I then tried installing the latest stable and it did not produce menu items either so this must have been broken for some time.

image

Even if other menu folders are selected no shortcuts are created. The "Do not create shortcuts" option is not checked. I tried selecting that option in case the logic was reversed but again no shortcuts were created. The install for everyone or just for me option had no effect either.

tfredian:
The latest Windows operating system updates seemed to deprecate
the use of a obsolete Windows message used to trigger the update
of the Start Menu items after new shortcuts are added or deleted during
a package installation or removal. This fix changes the message
type sent after the MDSplus installation to ensure the menus are
updated.

This fixes #1763

### DIFF
--- a/deploy/packaging/windows/mdsplus.nsi
+++ b/deploy/packaging/windows/mdsplus.nsi
@@ -996,7 +996,7 @@ Function .onInit
 FunctionEnd ; .onInit
 
 Function .onGUIEnd
-	SendMessage ${HWND_BROADCAST} ${WM_WININICHANGE} 0 "STR:Environment" /TIMEOUT=5000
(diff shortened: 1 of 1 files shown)


tfredian:
retest this please

zack-vii (approved):

PR #1767, opened 2019-08-18
Build: fixed JRE and UIL def in configure

zack-vii:


zack-vii:
the definitions for JRE and UIL were broken but everything seemed to be good. no that the definitions are correct the packaging of some OS's is broken. which one is right? should we just remove the lines?

tfredian:
retest this please

zack-vii:
both vars are/can be defined globally. its save to remove the lines

tfredian:
Several of these JRE and UIL assignments in the configure.ac seemed to have been added by Fernando six months ago. Looking at the changes I am not entirely sure why they were added. It probably is no longer critical to even support 32-bit motif applications but if we do intend to support them it is important that /usr/bin/uil32 is used when building the uid's for 32-bit platforms. I don't know if these changes will cause the 64-bit uil to be used instead of uil32 on platforms where the docker image contains both uil and uil32.

As far as JRE is concerned, I couldn't find where the value JRE is even being used except perhaps in cconfigure.ac where it is deciding whether to build java or not.

zack-vii:
Well thing is the curretn version does not set the var either due to a syntax error (extra spaces ' '). So unless we are facing uil problems under 32 bit already... this should not change the outcome of the build but instead cleanup the configure.

santorofer:
Hi Tom,

I remember making those changes when I was working on the issue of installing MDSplus from source for Mac. But I cannot find the original commits in Git. I’m trying to remember what was the original issue that needed these changes.
I will go back to my notes and check.

Thank you Tom,
Fernando

santorofer:
Hi Timo,
I was wondering, what was the issue that you are trying to fix? Or how did you notice that there was a problem?
Cheers!

zack-vii:
Well if you run configure on all oses i checked you will notice that these lines are infact syntax error and have no effect. (" = " should be replaced by "=")
```log
14:30:05 checking for visual studio... no
14:30:05 /mdsplus-docker/jenkins/workspace/fc21/src/configure: line 8588: JRE: command not found
14:30:05 /mdsplus-docker/jenkins/workspace/fc21/src/configure: line 8589: UIL: command not found
14:30:05 checking size of long... 8
```
So in short, I figured this is not OK. 
It might be that the os you intended to fix allows the syntax. Basically the spaces need to be removed but this causes issues during build. so if you added those lines, maybe you can check if they do what you need them to do. Possibly you get the same result without them.
If not can you explain what you intended to do or what os is affected by this change.

santorofer:
Hi Timo,
What I remember is that I made some changes to configure.ac so that Java programs can be used in Mac, and the UIL compiler can be used to compile the Motif applications.

santorofer:
The OS that can potentially be affected will be Mac OS X. I believe.

zack-vii:
hm well for mac it was/is
```bash
UILCMD="uil"
*apple-darwin*) echo "Configuring for MacOS X";
       JRE = "";
       UIL = "/usr/bin/uil";
```
So even osx should interpret both lines as a function call instead of a env var assignment.

zack-vii:
When you run configure watch out for
```
checking for uil... /usr/bin/uil
```
which should not be changed by this commit.

santorofer:
Hi Timo,
So, this is strange because I remember being able to fix the issue of not having some of the Motif application working in Mac OSX after adding the line to the “uil” compiler, and not only that, the location of that compiler was /usr/bin/. Now it seems that that compiler, well, a link to the compiler, is located under /usr/local/bin :

/usr/local/bin/uil@ -> ../Cellar/openmotif/2.3.8/bin/uil

So, in other words, for Mac it should be now:

UIL="/usr/local/bin/uil”;

Am I right?

For Linux, I couldn’t find any uil, only uic which is actually the compiler for QT GUI toolkit….
Cheers,
Fernando

zack-vii:
retest this please

zack-vii:
So you a compiling for mac on your mac where it cannot find uil? jenkins macosx slave in alpha configures as follows
e.g. ttp://jenkins2.mdsplus.org/job/macosx/1767/consoleFull
```
17:30:59 /Users/twf/jenkins/workspace/macosx/src/configure: line 8699: JRE: command not found
17:30:59 
17:30:59 Error: additional UIL source file: /usr/bin/uil was ignored
17:30:59 
17:30:59 Severe: error opening source file: =
...
17:31:06 checking for uil... /usr/local/bin/uil
```
So setting UIL to /usr/bin/uil obviously did not work.
Question though: does this PR produce the correct motif applications on jenkins.
I triggered a new build just now. you can download the binaries from its workspace once its done.

santorofer:
Hi Timo,

If I use:

UIL="/usr/local/bin/uil”;

And then run autoconf, I get a configure script that doesn’t show those errors.
Ok, I will test your new build as soon as is done.

Thanks

zack-vii:
its done. this pr still results in 
```
17:57:42 checking for uil... /usr/local/bin/uil
```
so the configure script finds the file and there is no need to hardcode it.

santorofer:
Hi Timo,

Yes, I understand. Removing those hardcoded paths in configure.ac sounds good to me.

I actually went into Jenkins and downloaded, for example, dwscope from the Workspace:

<http://jenkins2.mdsplus.org/job/macosx/ws/build/macosx/zck_jre_uil/test/dwscope/>

And I then downloaded dwscope into my Mac and it works fine it seems.
So, I think this looks good.
Cheers,
Fernando

zack-vii:
Could someone please approve this PR?

santorofer (approved):
Tested dwscope created for Mac OSX. It works correctly.

PR #1770, opened 2019-08-27
Fix:TreeNode::data()

GabrieleManduchi:
After the changes in setActiveTree() usage, method data() must be changed in order to set the nid-associated tree before evaluating. The mutex has been changed to recursive in order to avoid deadlocks.


AndreaRigoni (approved):
Good !

Issue #1362, opened 2018-07-17
MDSIP connection waits should handle aborts/SIGINT


zack-vii:
currently if you do e.g.
```tcl
show server myserver:port
```
The request times out after several seconds if the  "Transport endpoint is not connected".
I those cases it would be great if one could either set the timeout or abort the attempt via SIGINT (Ctrl+C).

The service should be able to handle such aborted waits/requests merciful and eventually abort any further execution of the dispatched command.

tfredian:
Providing SIGINT handling of the show server command is easy to do and I tested an implementation of it. Aborting a executing command on the server if it is not responding to a show server command is another story which might be extremely difficult to implement since the mdsip process on the server must be in some unresponsive state in the first place. If show server does not get a response from the server the action it is performing must be preventing the main mdsip process thread from receiving connections or executing expressions sent from the client since that is how show server works. This would indicate a problem with the command being executed on the server which should be investigated.
It is unlikely the mdsip process would respond to any message sent to the server over the port the server mdsip process is listening so even abort or stop messages would be blocked as well. One could imagine implementing a tdi function that finds the pid of the server process and sends it a signal so that the client could make an mdsip connection using a different port to kill or interrupt the mdsip action server but that would require the use of an account with sufficient privileges.

So... should I implement the SIGINT interruption of the show server command which would only affect the client mdstcl process and have no affect on the hung server process?

joshStillerman:
Is this now resolved then ?

zack-vii:
does not seem to be the case
```tdi
show server 8.8.8.8:80
Checking server: 8.8.8.8:80
^C^C^C^C^CError in connect to service: Transport endpoint is not connected
Error connecting to server
```

zack-vii:
could eventually be fixed in the process of #1443

zack-vii:
fixed by #1781

PR #1781, opened 2019-09-02
Fix: SIGINT handling for readline and mdstcpip

zack-vii:
Originally the PR was intended to fix the Ctrl+C behaviour in readlind applications mdsdcl and tditest. Turned out this also allowed for user interruptions of select except for Windows. In order to enable interruptions for Windows a socket hack had to be used. looks quite compliated but its basically a self-connected DGRAM socket that is included in the select fd_set. The signal handler closes teh socket and cases select to return. While i was investigating i found that teh unix signel(2) method should only be used to set the handler to SIG_IGN or SIG_DFL. so I updated teh occurences in teh vincinity of the touched code. On that matter is routines used ABORT() to cleanup SocketList via shutdown(). I does this now under a few more circumstances. Unfortunately, there is no consistent way to guarantee cleanup procedure.

### DIFF
--- a/include/pthread_port.h
+++ b/include/pthread_port.h
@@ -157,6 +157,7 @@ _CONDITION_UNLOCK(input);\
  #ifndef _WIN32
   #define INITIALIZESOCKETS
  #else
+  #define SHUT_RDWR 2
   static pthread_once_t InitializeSockets_once = PTHREAD_ONCE_INIT;
   static void InitializeSockets() {
     WSADATA wsaData;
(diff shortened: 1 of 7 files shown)


zack-vii:
this pr addresses issue #1362

tfredian (approved):
Complicated but seems ok. If I am not mistaken, when we switch to always using squash and merge for pull requests the main pull request comment is included in the commit text so it would be very useful if you edit the top pull request comment prior to merging to be much more descriptive and explain the intent of the pr instead of just having a list of commit titles.

PR #1784, opened 2019-09-03
Updated FLIR & BASLER cameras

fmolon:
Changes only in FLIR and BASLER cameras (python device & device support)


GabrieleManduchi (approved):
Ok

PR #1787, opened 2019-09-04
Fix: MdsGet1DxA's align should be able to handle size=0

### LINKED ISSUE #1786: Floating point exceptions in MdsGet1DxA.c
I am getting some Floating point exceptions in MdsGet1DxA.
According to gdbm the culprit is the align macro called at [line 97](https://github.com/MDSplus/mdsplus/blob/8e68e5cb06d15a5232ae23d34279929e810d326e/mdsshr/MdsGet1DxA.c#L97)
It seems to me like align_size can be 0 if dtype is not DTYPE_T and *length_ptr=0 and this is causing the floating point exception ...

The use case is when trying to access the first element of an empty array by using an expression such as:
data(GetSegment(\PATH_TO_NODE,0))[0]
where the node is not filled.

tfredian:
Fixes https://github.com/MDSplus/mdsplus/issues/1786

Which describes a segmentation fault that can occur when subscripting an empty array.

A regression test for this problem was also added.

### DIFF
--- a/mdsshr/MdsGet1DxA.c
+++ b/mdsshr/MdsGet1DxA.c
@@ -74,7 +74,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <mdsshr.h>
 #include <STATICdef.h>
 
-#define align(bytes,size) ((((bytes) + (size) - 1)/(size)) * (size))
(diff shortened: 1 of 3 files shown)


tfredian:
retest this please

zack-vii (approved):
should do the trick, i changed the title to something more useful

PR #1788, opened 2019-09-04
Feature: Convert nids in funs to paths

tfredian:
When a node reference is hardcoded in side a tdi fun the node
is converted into a nid (node id reference) at compile time when
the fun is first referenced. This will often cause problems if
subsequently the current process opens other trees since the nid
compiled into the function may not resolve to the same node in
the new tree context.

This change will automatically convert any nids found in a tdi fun
to a tree path instance. When the function is executed, instead of
trying to reference the node by a nid, the function will now find
the node referenced using a tree path string. This should avoid
the potential conflicts when the current tree context is changed
after the initial loading of the tdi function.

For a test I used the sample main tree and I created a new main
shot 1 as follows:

TCL> edit/new main/shot=1
TCL> add node numeric/usage=numeric
TCL> put node numeric 42
TCL> write

then in TDIC:
TDI> treeopen('main',-1)
265388067
TDI> getnci(numeric,'nid_number')
4
TDI> treeopen('main',1)
265388041
TDI> getnci(numeric,'nid_number')
1
TDI> public fun test(){return(numeric);}
Fun PUBLIC test () {
        Return (NUMERIC);
}
TDI> test()
42
TDI> treeopen('main',-1)
265388067
TDI> test()
42

Using the alpha libraries the following is the behavior:

treeopen('main',1)
265388041
public fun test(){return(numeric);}
Fun PUBLIC test () {
        Return (NUMERIC);
}
test()
42
treeopen('main',-1)
265388067
test()
[1,2,3,4,5,6,7,8,9,10]

In this alpha case the numeric nid  (1 in shot 1) is stored in the fun and when switched to shot -1 it does not reference the numeric node but instead returns the content of the member node which in that tree is identified by nid 1.

### DIFF
--- a/tdishr/TdiYaccSubs.c
+++ b/tdishr/TdiYaccSubs.c
@@ -39,6 +39,9 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <libroutines.h>
 #include <tdishr_messages.h>
 #include <mdsshr.h>
+#include <treeshr.h>
+#include <string.h>
+#include <strroutines.h>
 
 
 
@@ -185,6 +188,20 @@ int TdiYacc_IMMEDIATE(struct descriptor_xd **dsc_ptr_ptr)
   return status;
 }
 
+static int fixup_nid(int *nid, void *unused __attribute__ ((unused)), struct descriptor *path)
+{
+  char *path_c = TreeGetPath(*nid);
+  if (path_c) {
+    struct descriptor path_d = { 0, DTYPE_T, CLASS_S, 0 };
+    path_d.length = (unsigned short)strlen(path_c);
+    path_d.pointer = path_c;
+    StrCopyDx(path, &path_d);
+    TreeFree(path_c);
+  }
+  return 1;
+}
+
+
 /*--------------------------------------------------------------
 	Automatic evaluation of constants does what is safe to do.
 	Examples: F_COMPLEX(1,5) becomes DTYPE_FC: CMPLX(.1E1,.5E1).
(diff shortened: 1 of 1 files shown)


zack-vii:
i cannot check my self atm but could you verify if the treepath matches the string that defined the nid. i assume if you would reconstruct the path from the nid, you cannot preserve tge quality of the path (normal, full, minimal). this may break relative paths as default node may change.

tfredian:
TreeGetPath returns a full path string (including the \tree:: part). In the above example I just tried defining the test fun in shot 1 and then switching to shot -1, changing default to .subtree and executing the test() fun and it returned the numeric node as it should. No hurry if you want to do some more testing before reviewing the pr.

zack-vii:
i will see to it tomorrow. have a nice day

tfredian:
Here is a cryptic illustration of what the code now produces when it compiles the fun:
```
>>> from MDSplus import Tree
>>> type(Tree('main',-1).tdiCompile('public fun test(){return(numeric);}').getDescs()[1].getDescs()[0])
<class 'MDSplus.tree.TreePath'>
>>> repr(Tree('main',-1).tdiCompile('public fun test(){return(numeric);}').getDescs()[1].getDescs()[0])
'\\MAIN::TOP:NUMERIC'
```
Here is what it used to do:
```
>>> type(Tree('main',-1).tdiCompile('public fun test(){return(numeric);}').getDescs()[1].getDescs()[0])
<class 'MDSplus.tree.TreeNode'>
```
Changing the default node on the tree does not change anything. When the fun is loaded it gets compiled and the node references are made into nid's if the nodes exist but just before the compilation compiles it converts all nids in the fun to treepath instances.

Issue #1785, opened 2019-09-04
tdi function pointer to subtree/node breaks across pulses


VadimNV:
Dear MDSplus developers team,

  There is a bug in the way TDI functions caches/stores their "pointers" to the nodes and/or subtrees. It seems that the issue is due to a given node changing its NID_NUMBER between some pulse number ($SHOT) After looking into this for a while, here is the information that can be useful. 
  If there is a TDI command that can read node data by the NID_NUMBER rathar than by the node name/path - please share, this would be very useful.
  Thanks in advance! 

1) simple test tdi function "simple_test.fun":
PUBLIC SIMPLE_TEST(IN _name, IN _IOtype) {
  WRITE(*,data(raw_mag.acq2106_034:_DT));
  WRITE(*,getnci("raw_mag.acq2106_034","NID_NUMBER"));

  WRITE(*,data(pss.PF_CONFIG));
  WRITE(*,getnci("pss","NID_NUMBER"));

  RETURN("TESTING");
}
2) here "ST40" is a main tree, "raw_mag" and "PSS" are subtrees. In my tests ".acq2106_034:_DT" is a NUMERIC node storing a value of 1.0; "PF_CONFIG" is a NUMERIC node node storing 1.0 or $ROPRAND depending on $SHOT. There is also a tree called "ECE" with a HELP node storing a string "This tree stores data from the Electron Cyclotron Emission"

3) Hypothesis: when you connect to MDSplus server for the first time, a TDI function caches (stores) the subtree and/or node id numbers ( getnci("nodeName","NID_NUMBER") ). When, within the same session/connection, one opens a tree where the node of interest has a distinct NID_NUMBER between different $SHOT numbers, the TDI function tries to read from the wrong place. Sometimes, by chance, it returns the data from a different node, other times it returns /* bad */ if no node exists under the cached NID_NUMBER.

4) from command line, start TDI and run tests
$tdic
TDI> treeopen('st40',6066)
TDI> simple_test("I_TF_PSU","ain")
       1.00000E0
    50331649
                        $ROPRAND                              # CORRECT OUTPUT
   234881024                                                    # SAME NID_NUMBER - OK
"TESTING"
TDI> treeopen('st40',6067)
265388067
TDI> simple_test("I_TF_PSU","ain")
       1.00000E0
    50331649
                        $ROPRAND                                                     # CORRECT OUTPUT
   234881024                                                                           # SAME NID_NUMBER - OK
"TESTING"
[...]

PR #1789, opened 2019-09-05
Feature: convert path-like nids in tdi funs to tree_paths

zack-vii:
When a node reference is hardcoded inside a tdi fun the node
is converted into a nid (node id reference) at compile time when
the fun is first referenced. This will often cause problems if
subsequently the current process opens other trees since the nid
compiled into the function may not resolve to the same node in
the new tree context.

This change will cause TdiYacc to convert any nid inside of tdi fun into 
to a tree path instance. When the function is executed, instead of
trying to reference the node by a nid, the function will now find
the node referenced using a tree path string. This should avoid
the potential conflicts when the current tree context is changed
after the initial loading of the tdi function.

### DIFF
--- a/tdishr/TdiVar.c
+++ b/tdishr/TdiVar.c
@@ -104,8 +104,8 @@ extern int TdiResetGetRecord();
 typedef struct link {
   struct link *left, *right;	/* binary tree links */
   unsigned short reserved;	/* tree flags */
-  struct descriptor_xd xd;	/* data descriptor */
-  struct descriptor name_dsc;	/* name descriptor */
+  mdsdsc_xd_t xd;	/* data descriptor */
+  mdsdsc_t name_dsc;	/* name descriptor */
   unsigned char name[1];	/* unchanging name */
 } node_type;
 typedef struct {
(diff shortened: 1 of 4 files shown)


zack-vii:
I could not find my way thru the TdiYacc*.c to figure out where I could add the 
```c
++TdiRefZone.l_rel_path // should be placed at 'FUN('
--TdiRefZone.l_rel_path // should be placed at ')' if it terminates a FUN(
```
I added the lines in TdiVar.c instead where it covers all tdi funs read from .fun files. Manually defined functions will still suffer from issue #1785.

zack-vii:
retest this please

tfredian (commented):
Interesting simpler approach but... I don't know if it is important but this only seems to work if when a tdi fun is compiled from a tdi fun file. I tried doing the commands I showed in my similar PR and it did not work if the test function was entered directly in tditest/tdic. Setting a break point on compile_without_resolving_paths I found that when treeopen() is compiled it is called but if public fun test(){return(numeric);} is entered the compile_without_resolving_paths does not get called. As I said it may not be important but it does make for a subtlety that may come back to bite us. So not sure if your approach could be tweaked to work or if we should go with my original PR.

zack-vii:
Never mind, I found the two places where to add the lines. in fact they were meant to be there but have been commented out:
git blame
```
316f9d6f tdishr/TdiYacc.y             (mdsplus           1997-06-27 12:07:37 +0000 417)                                                         /*++TdiRefZone.l_rel_path;*/
316f9d6f tdishr/TdiYacc.y             (mdsplus           1997-06-27 12:07:37 +0000 435)         | fun stmt                              {/*TdiRefZone.l_rel_path--;*/
```

tfredian (approved):
Works!

PR #1791, opened 2019-09-06
Build: added logic to determine if we should make a new maven release

zack-vii:
This added some scripts to determine if the java projects in question have been updated since the last maven release. The 2-maven Task can now run automatically after an 1-alpha-release and will roll roll out a release only if necessary.


zack-vii:
Hi all would be nice if someone could approve this PR. it only changed stuff in deploy/platform/maven

tfredian (approved):

PR #1794, opened 2019-09-11
Fix: reactivate --with-java_target and made java compatible to java 6 

### LINKED ISSUE #1792: Build pipeline does not respect --with-java_target
When running a build with CONFIGURE_PARAMETERS=--with-java_target=6 I would expect the build pipeline to generate jar files compatible to java 1.6. However, this is not the case (loading the jars causes the error "unsupported class file version 52.0" (which indicates a java 1.8 class file).

Digging a bit into the build pipeline, I found the following lines in configure.ac:

```
AC_ARG_WITH([java_target],
	    [AS_HELP_STRING([--with-java_target=version],
                            [specify version of compiled mdsobjects classes (i.e. 6)])],
            [AS_VAR_SET([JAVAVERSION], ["-source $withval -target $withval"])],
            [AS_VAR_SET([JAVAVERSION], [])])
```

However, when looking into the java/mdsobjects/makefile.in I find these lines
```
JAVAVERSION = @JAVAVERSION@

...

echo '$(CLASSPATH_ENV) $(JAVAC) -d $(JAVAROOT) $(AM_JAVACFLAGS) $(JAVACFLAGS) '"$$list2"; \
	  $(CLASSPATH_ENV) $(JAVAC) -d $(JAVAROOT) $(AM_JAVACFLAGS) $(JAVACFLAGS) $$list2; \
```

JAVAVERSION is unreferenced in the compilation command.

I have fixed this locally by making the following two modifications:

Change the javac compilation call to
```
echo '$(CLASSPATH_ENV) $(JAVAC) -d $(JAVAROOT) $(AM_JAVACFLAGS) $(JAVACFLAGS) $(JAVAVERSION) '"$$list2"; \
	  $(CLASSPATH_ENV) $(JAVAC) -d $(JAVAROOT) $(AM_JAVACFLAGS) $(JAVACFLAGS) $(JAVAVERSION) $$list2; \
```

and modifying the directive in configure.ac to only modify the target binary version, but not the source language version

```
AC_ARG_WITH([java_target],
	    [AS_HELP_STRING([--with-java_target=version],
                            [specify version of compiled mdsobjects classes (i.e. 6)])],
            [AS_VAR_SET([JAVAVERSION], ["-source 1.8 -target $withval"])],
            [AS_VAR_SET([JAVAVERSION], [])])
```

Is there an interest to have this patched? I can submit a pull request.

Edit: Formatting

zack-vii:
* Includes JAVAVERSION in JAVACFLAGS to enable flags in default java compiler code generated by autoconf.
* modifies jscop and jtraverser2 source to be backwards compatible with java 6
 - replaced try-resource clause with try-finally-close
this should fix #1792


zack-vii:
retest this please

tfredian:
retest this please

tfredian:
Apparently there is a lot more of the java code which is incompatible with java 6. Is it worth the effort to make the source compatible with 6 or just do what was suggested in the issue, hard coding source=1.8 instead?

zack-vii:
hm i tried leaving -source default to 1.8 but then javac would not compile because 1.6 and 1.8 are not compatible i that sense. seems like -target and -source should be the same.

zack-vii:
but I was not aware that we compile for java6 by default.  we could set the default to java 8

zack-vii:
i deactivated the Javaversion setting for jtaverser2 and mdsplus-api as they are build for java8

tfredian:
I don't fully understand at what point this broke. We used to target Java 6 on our builds (I think not very long ago) and it worked just fine using java 1.8 to build for java 6. I know we targeted Java 6 for compatibility with the default java installed on certain OS's. Perhaps that is no longer required but who knows.

zack-vii:
hm not sure. mdsplus-api andbor jtraverser2 was only working with java8 from tge beginning i think. i played with source 8 target 6 but it complained. now it can build all prijects in 6 but mdsplu-api and jtraverser2. should be good to go.

zack-vii:
retest this please

zack-vii:
retest this please

tfredian (approved):

PR #1800, opened 2019-09-19
Gm improve jscope

GabrieleManduchi:
A set of fixes in jScope for:

1) Fix error reporting that had been broken in the last improvements
2) Fix 2D signal visualization that had been broken by row and col TDI variable settings
3) Other minor fixes


zack-vii (approved):
just a suggestion: if you have a long list of 
```
if (var==a ||var==b||... )
  dothis();
else
  dothat();
```
you may consider a switch instead.
```
switch (var) {
case a:
case b:
...
dothis();break;
default:
dothat();
}
```
but seems ok

Issue #1802, opened 2019-09-24
MDSplus.Range() issue: depending on the delta step, the last element of the original array is included in the final output.


santorofer:
In MDSplus Range() method, depending on the value of the delta step to divide the array, the last element of the array is included in the array output by Range(), which it's not the correct answer for Range().

See the following examples:

```python
>>> Range(0., 9., 1.4999).data()
array([ 0.    ,  1.4999,  2.9998,  4.4997,  5.9996,  7.4995, 8.9994])
>>> Range(0., 9., 1.5001).data()
array([ 0.    ,  1.5001,  3.0002,  4.5003,  6.0004,  7.5005, 9.0006])
```

In other words, if the remainder of the division between the last element of the array and the delta step is less than 0.5, then the array contains the correct elements:
```python
>>> Range(0., 9., 1.1).data()
array([0. , 1.1, 2.2, 3.3, 4.4, 5.5, 6.6, 7.7, 8.8])
```
where the remainder of the division is less that 0.5:
```python
>>> 9/1.1
8.181818181818182
```
On the other hand, if the remainder is 0.5 or greater:
```python
>>> Range(0., 9., 1.2).data()
array([0. , 1.2, 2.4, 3.6, 4.8, 6. , 7.2, 8.4, 9.6])
>>> 9/1.2
7.5

>>> Range(0., 9., 1.3).data()
array([0. , 1.3, 2.6, 3.9, 5.2, 6.5, 7.8, 9.1])
>>> 9/1.3
6.9230769230769225
```

zack-vii:
Just to make sure I understand you wish for range to act exactly like e.g. python's range. 

> which it's not the correct answer for Range().

What definition are you referring to. It seems - based on the source code - the current behavior is intentional. It is caused by TdiNint in tdishr/TdiDtypeRange.c line 174:
```c
  if (STATUS_OK && nelem.pointer->dtype != DTYPE_L) {
    status = TdiNint(&nelem, &nelem MDS_END_ARG);
    if (STATUS_OK && nelem.pointer->dtype != DTYPE_L)
      status = TdiLong(&nelem, &nelem MDS_END_ARG);
  }
```

tfredian:
[...]

PR #1803, opened 2019-09-25
Fix: number of intervals in RANGE should round to lesser integer

zack-vii:
according to #1802 the behaviour of Range is wrong. Well its different from python's range. Given the source code, the difference might have been intentional.


zack-vii:
tditest tests this behaviour and expects Range to exceed end. even with floor rounding this would not resemble pythons range. So far this pr would change our range so that
```
python:  n>=begin and n< end
mdsplus: n>=begin and n<=end         (pr)
mdsplus: n>=begin and n<=end+delta/2 (alpha)
```
for all elemens n produced by range.

zack-vii:
we can use
```tdi
CULL([_s,_e],,_s:_e:_d)
```
to achieve the same think without changes to the codes.

PR #1804, opened 2019-09-25
Gm improve jscope

GabrieleManduchi:
A bunch of fixes and improvements for jScope. (both jScope and ScopeUtilities.c in mdfsmisc).
In particular, event-driven updates for growing segmented signals have been optimized to minimize disk access in update. A memory leak leading to degraded performances after some utilization time has been fixed, too.

A fix has also been added to mstcpip/ProcessMessage.c in order to disconnect from registered events when a communication goes down. Previously, when a connection registered to a MDSplus event and then went down without unregistering the event, the AST routine for that event found connection pointer to null and crashed when using it.


AndreaRigoni (approved):
OK, only one change outside scope.
mdstcpip change has been checked.

PR #1805, opened 2019-09-26
Gm improve jscope

GabrieleManduchi:
A set of fixes to enable optimized data access via jScope MdsDataProvider. Recent changes broke the  mechanism.

C++ methods befinSegmentResamples(), beginSegmentMinMax(), putSegmentesamples(), putSegmentMinMax() have been also added in addition to maheSegmentResample() and makeSegmentMinMax(). These method will be used bu devices producing large signals in order to enable jScope smart resampling.


AndreaRigoni (approved):
OK

PR #1807, opened 2019-09-26
Fix: Added CULL() to get the time dimension with correct segment lengths.

santorofer:
Added CULL() to get the time dimension with the correct array ending for each segment.
So that, there will not be extra points in the dimension array.


tfredian:
Is the print at line 164 intentional or just a leftover debug statement?

tfredian:
also did you want to use a range for the first argument of CULL? I thought an Array worked better.

joshStillerman:
This code is not right.  There are some subtleties.  F and S and I 
worked out the solution

joshStillerman (approved):

PR #1808, opened 2019-09-26
Fix: Fixing python3 compatibility with acq435st

WhoBrokeTheBuild:
Changed `import Queue` to `from queue import Queue`
Changed syntax of relative imports to work in both python 2 and 3

### DIFF
--- a/pydevices/HtsDevices/acq400_hapi/__init__.py
+++ b/pydevices/HtsDevices/acq400_hapi/__init__.py
@@ -1,12 +1,9 @@
-from netclient import Netclient
-from netclient import Siteclient
-from netclient import Logclient
-from acq400 import Acq400, STATE, AcqPorts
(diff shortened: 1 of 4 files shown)


zack-vii:
you might want to consider doing both, one in a try and thebother in the except so you will end up with a Queue for both py2 and py3

joshStillerman:
did you test the  import '.' stuff in both 2 and 3, and in mdstcl ?

joshStillerman:
And mdstcl I assumeOn Sep 27, 2019 1:46 AM, Stephen Lane-Walsh <notifications@github.com> wrote:Yep! This code all works for python 2 and 3, even the queue code.



On Thu, Sep 26, 2019, 4:32 PM Josh Stillerman <notifications@github.com>

wrote:

joshStillerman (approved):

PR #1811, opened 2019-10-01
Feature:display proper NCI info for segmented nodes

GabrieleManduchi:
Now the number of segments is reported in display NCI and additional data such as dtype are displayed only for non segmented nodes


AndreaRigoni (approved):
OK, only jScope changes

Issue #1810, opened 2019-09-30
Performance deteriorating over repeating calls


merlea:
Hello,

I have found an issue with MDSplus releases > 7-46-1 when using the MDSplus backend for IMAS developped by @GabrieleManduchi . It manifests as a deterioration of performance when repeating the same sequence of (reading) calls.

I have tried to reduce the existing IMAS backend to a minimum test case in C++:

```
#include <mdsobjects.h>
#include <cstring>
#include <string>
#include <chrono>

int fill_tree(void)
{

  int shot = 1;

  MDSplus::Tree * modelTree = new MDSplus::Tree("test", -1);
  modelTree->createPulse(shot);
  delete modelTree;
  MDSplus::Tree * tree = new MDSplus::Tree("test", shot);  
  tree->setDefault(tree->getNode("\\TOP"));

  char databuf[] = "DATA(MAKE_RANGE(0D0,1D0,0.001D0))-0.2D0";
  MDSplus::Data * fdata = MDSplus::compile(databuf,tree);
  MDSplus::Data * data = fdata->data();
  MDSplus::deleteData(fdata);

  char timebuf[] = "DATA(MAKE_RANGE(0D0,1000D0,1D0))";
  MDSplus::Data * ftime = MDSplus::compile(timebuf,tree);
  MDSplus::Data * time = ftime->data();
  MDSplus::deleteData(ftime);

  {

    int nSlices = 2000;
    MDSplus::Int32 * nSlicesD = new MDSplus::Int32(nSlices);
    MDSplus::Int32 * nSlicesM1D = new MDSplus::Int32(nSlices-1);

    MDSplus::TreeNode * node = tree->getNode("\\TOP.EQUILIBRIUM:DATA");
    node->deleteData();
    char sbufd[] = "data(GetSegment(build_path(\'\\\\TOP.EQUILIBRIUM:TIME\'),0))[0]";
    MDSplus::Data * startDataD = MDSplus::compile(sbufd,tree);
    char ebufd[] = "DATA(GetSegment(build_path(\'\\\\TOP.EQUILIBRIUM:TIME\'),0))[$1 - GetSegmentInfo(build_path(\'\\\\TOP.EQUILIBRIUM:TIME\'), 0) - 1];";
    MDSplus::Data * endDataD = MDSplus::compileWithArgs(ebufd,tree,1,nSlicesD);
[...]

PR #1813, opened 2019-10-01
Fix: create TdiRefZone.l_zone only if none exists

zack-vii:
should fix issue #1810


zack-vii:
retest this please

GabrieleManduchi (approved):
Seems OK, and the problem disappeared

PR #1815, opened 2019-10-02
Fix: acq2106_423 decimation fixes using MDSplus.Range() and correcting the ends of the range by accounting for decimation values.

santorofer:
The changes are two fold:
1. To Homogenize the way `acq435st.py` and `acq2106_423st.py` deal with the creation of the time dimension segments. We have now corrected the ends of the ranges by accounting for the decimation value. We use MDSplus.Range(). The defined function called truncate() no longer is used.
2. Bring Queue and Empty Python 2 libraries to comply with Python 3 queue library

TODO
- [x] Port changes to `acq435st.py`

### DIFF
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -128,6 +128,9 @@ class _ACQ2106_423ST(MDSplus.Device):
 
             decimator = lcma(self.decim)
 
+            if self.seg_length % decimator:		
+                 self.seg_length = (self.seg_length // decimator + 1) * decimator
+
             self.device_thread.start()
 
             segment = 0
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -119,9 +119,6 @@ class _ACQ2106_423ST(MDSplus.Device):
                     ans = lcm(ans, e)
                 return int(ans)
 
-            def truncate(f, n):
-                return math.floor(f * 10 ** n) / 10 ** n
-
             if self.dev.debug:
                 print("MDSWorker running")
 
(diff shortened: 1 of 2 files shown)


pydevices/HtsDevices/acq2106_423st.py
```diff
@@ -132,34 +128,31 @@ def truncate(f, n):
...
 
                 buffer = np.frombuffer(buf, dtype='int16')
                 i = 0
                 for c in self.chans:
                     if c.on:
                         b = buffer[i::self.nchans*self.decim[i]]
-                        c.makeSegment(self.dims[i].begin, self.dims[i].ending, self.dims[i], b)
-                        self.dims[i] = Range(self.dims[i].begin + truncate(self.seg_length*dt,3), self.dims[i].ending + truncate(self.seg_length*dt,3), dt*self.decim[i])
+                        
+                        dim_limits=[begin, begin + self.seg_length*dt - 1]
```
WhoBrokeTheBuild: You should compute `begin` from `segment` to avoid floating point issues when you accumulate `begin` over time

WhoBrokeTheBuild (changes requested):

zack-vii:
now I see what you planned on doing with cull. why dont you use the window object instead which was meant for just this purpose.
```python
from MDSplus import SUBTRACT,DIVIDE,UNARY_MINUS,Window,Range,Dimension
win  = Window(UNARY_MINUS(self.trigger_pre),SUBTRACT(self.trigger_post,1),self.trigger)
rang = Range(None,None,DIVIDE(1.,self.clock))
dim = Dimension(win,rang).setUnits('s')
```
where the self.xxx are nodes of the device.
with segments you just edit the window pre and post to first and last sample index stored in tha segment, where the index is rekative to the trigger (i.e. negative for pretrigger samples)

santorofer:
Yes, you are right, my mistake. I'll correct it.

joshStillerman:
@WhoBrokeTheBuild  and or @santorofer  
  Can you verify that this latest version gets the right answers ?  The code looks good.

joshStillerman (approved):
Looks great

WhoBrokeTheBuild (approved):

PR #1819, opened 2019-10-05
Removing extra writes to stdout [stable]

merlea:
It seems a previous cleanup missed a few writes to stdout.

### DIFF
--- a/mdsobjects/cpp/mdstreeobjects.cpp
+++ b/mdsobjects/cpp/mdstreeobjects.cpp
@@ -1471,7 +1471,6 @@ void TreeNode::putRow(Data *data, int64_t *time, int size)
 
 TreeNode *TreeNode::getNode(char const * relPath)
 {
-	std::cout << "In C++ TreeNode getNode, relPathS" << std::endl;
 	int newNid;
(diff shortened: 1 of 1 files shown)


GabrieleManduchi:
The same patch has already been reported in alpha, but since those messages area annoying in ITER integrated modeling programs, we can report it directly in stable

GabrieleManduchi:
Tom, how do we proceed? Can you report it directly in stable?

merlea:
Note that the two patches are slightly different, in alpha there is only one such statement removed as the others must have been removed earlier.

tfredian:
Gabriele, I assume your question is related to the: "Can one of the admins verify this patch?"
This message is automatically generated when someone (other than a core MDSplus developer) makes a pull request. Since there is a potential vulnerability with someone putting harmful code into a PR which could be executed during a build and test, a member of the developer team has to simply use a comment on the PR with the phrase "Ok to test". This will cause the PR to be tested and if it passes the test we can approve and merge it.

tfredian:
ok to test

GabrieleManduchi (approved):

PR #1820, opened 2019-10-07
Feature: Added support for acq2106+MGT-DRAM.

seanalsop:
I've adapted one of the other acq400 scripts to work with MGT-DRAM. Let me know if there is anything that needs changed. 
Note: this does need the current acq400_hapi to work correctly, either installed via PIP or pointed to via PYTHONPATH.


tfredian:
ok to test

tfredian:
This looks fine to me. There do seem to be quite a few print statements which are not controlled by the debug attribute. Should they always occur or should they depend on the debug attribute?

seanalsop:
You're right there was some left over print statements, so I got rid of them. I've left some of the other print statements, because I like to know when each step has started/ended and I think it is probably useful for other users too. 

Thanks!

santorofer:
Hi Sean,
I've been testing this python MDSplus device code with acq2106 in our test rack. I noticed that the creation of the INPUT_xxx nodes was not happening because "parts" was not defined in init()  where you were appending the input nodes. I then moved the creation of those nodes up in the code just after parts[ ] is defined in  the class ACQ2106_MGT(). I am now able to have those nodes in my tree, but I don't seem to get data into them. I'm wondering if I'm not understanding how the ACQ is being triggered. There is a mode called "role_default", is that equivalent to a software trigger?

santorofer:
Hi again, I understand now that selecting "soft" or "automatic" trigger actually set the device as a transient recorder, am I correct?

santorofer:
I think I know why the streaming was not started, it's because the MDSWorker tread was not started in the init() method. This is what I see now:

Running init
acq400_hapi.Acq2106_MgtDram8 172.20.240.13
acq400_hapi.Acq2106 172.20.240.13
Netclient.init 172.20.240.13 4220
Netclient.init 172.20.240.13 4221
Netclient.init 172.20.240.13 4222
Netclient.init 172.20.240.13 2235
Netclient.init 172.20.240.13 4233
Netclient.init 172.20.240.13 4232
Netclient.init 172.20.240.13 4234
uut 172.20.240.13 site 11 not populated

But I still don't get data in the input nodes...
What are the numbers in the last column of that output?

seanalsop:
Hi Fernando,

Thanks for taking the time to look at this. As of yesterday Peter and I were
making some quality of life changes and we'd left it in a bad state - so I think
this is why you were having some issues.

To answer your questions:

The creation of the INPUT_xxx nodes was originally being done in init(), but we
were experimenting with moving them.

Automatic or soft are the method of trigger you want to use for
the digitizer. In the init method it will configure the system for whatever
trigger the user has asked for (hard, soft or automatic).

The numbers in the last column of the output you provided are the port numbers
that HAPI is connecting to under the hood. The python establishes these connections
when it is invoked so it can configure the UUT.

I'd like to close this PR for now because we want to add some more features.
I'll reopen in the next day or two once they're done.

Thanks again for taking a look at the code! I'll make sure it is fully functional 
by the next time you use it.

santorofer:
Hi Sean,

No problem at all. We are anxious to use it in our new test rack!.

One of the issue that the code had, specifically in the init() method of the acq2106_mgt() device class was that it was missing the the start of the MDSWorker tread:

        thread = self.MDSWorker(self)
        thread.start()

, so that it was never really trying to collect the data. I am now able to run it, but the issue I'm seeing now is that I'm getting a Queue.Empty exception when trying to get the full buffer (buf = self.full_buffers.get(block=True, timeout=1) also from the init() method.
Yes, let us now how things evolve from. your side.
Cheers.

PR #1821, opened 2019-10-09
Feature: reentrant TdiCompile

zack-vii:
* before TdiCompile had some limitations:
> 	For recursion must pass LEX:
> 	        zone status bol cur end
> 	        tdiyylval tdiyyval
> 	        tdiyytext[YYLMAX] tdiyyleng tdiyymorfg tdiyytchar tdiyyin? tdiyyout?
> 	For recursion must pass YACC also:
> 	        tdiyydebug? tdiyyv[YYMAXDEPTH] tdiyychar tdiyynerrs tdiyyerrflag
> 	Thus no recursion because the tdiyy's are built into LEX and YACC.
> 	IMMEDIATE (`) must never call COMPILE. NEED to prevent this.

this PR  supports (`) on COMPILE

### DIFF
--- a/mdsobjects/cpp/testing/MdsExpressionCompileTest.cpp
+++ b/mdsobjects/cpp/testing/MdsExpressionCompileTest.cpp
@@ -108,7 +108,7 @@ int main(int argc UNUSED_ARGUMENT, char *argv[] UNUSED_ARGUMENT){
     // TODO: adds more tests .. //
     END_TESTING;
 
-    BEGIN_TESTING(GetIdentPutIdent);
+    BEGIN_TESTING(MultiThreadPub);
     MultiThreadTest(PubNCmd,PubCmds);
     END_TESTING;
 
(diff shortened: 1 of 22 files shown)


tfredian (approved):
If this works then great job! It would be even better if you split this kind of PR into two or three separate PR's. More than half of the changes are various renames of variables and type declarations and should probably be done in a different PR. This practice of cleaning up the code as you go while you are trying fix some other issue makes it very difficult to review the changes and understand which changes are relevant to the stated purpose of the pull request.

zack-vii:
Sorry I know. I tend to do a lot of cleanups. This helps me to understand the code better and eventually whats lets is a clear picture of whats actually preventing me from achieving the original goal.

What finally led to the solution was embedding the globals used by TdiYacc and TdiLex into the respective funtions and pass the shared var as argument. Well and the fact the the TDI_STACK_IDX should be controlled by TdiYacc_IMMEDIATE rather than TdiCompile.

I think the code could be threadsafe it it was not for the public vars. But the test in MdsExpressionCompileTest that fails should probably not be run multithreaded.

zack-vii:
retest this please

PR #1823, opened 2019-10-10
Feature: add --server option to tditest to run thru mdsip

zack-vii:
```
tditest --server mdsip-server:8000
```
will now open a tditest session where every command will be executed via mdsvalue on the selected server.
Some bugfixes and performance tweaks for mdsip's ProcessMessage:
* reuse TdiThreadStatic
* fixed error handling (was supposed to return full trace but was broken)
* dont clear DEBUG so client can use it


zack-vii:
retest this please

zack-vii:
this PR builds on changes in #1822

zack-vii:
For discussion. Another implementation would be to exchange the TdiThreadStatic in a way that would hold one per connection. Not sure if this would be better but it would require another API function to discard the buffer when the connection is dropped.

tfredian:
Is there some reason you changed the EMPTYXD macro in mdsdescrip.h??? Normally an xd without any data should have a length field of zero. Since it didn't seem to break anything I assume the code dealing with xd descriptors looks at the pointer before using the length but why change it in the first place.

zack-vii:
There is it seemed the XD methods do not update the length filed (at least not in all cases) whihc left length ==0 at any time. thus the GetErrorText method was always discarding the detailed TRACE and replaced it with the GetMsg of status. it would be much safer to check for pointer!=NULL in all cases.

zack-vii:
but you are right there are a few more locations where it checks for length, maybe instead we should fix teh Xd methods

zack-vii:
hm from 
mdsshr/MdsXdRoutines.c:82 MdsGet1Dx()
```c
    if STATUS_OK {
      dsc_ptr->length = 0;
      dsc_ptr->l_length = *length_ptr;
      dsc_ptr->class = CLASS_XD;
      dsc_ptr->dtype = *dtype_ptr;
    }
```
which is not good or non intuitive at least.

tfredian:
Yep, it would be much better to fix the length handling.

zack-vii:
i reverted the length = sizeof(void*) and added a comment to mdsdecrip.h CLASS_XD about length
in commit 
Fix: GetErrorText should check pointer

tfredian (approved):

Issue #1824, opened 2019-10-14
python post-installer broken


zack-vii:
I am not sure how we are with python on linux and the PYTHONPATH way of integrating MDSplus. In order to support newer versions of debian, the current debian postinstall script should be updated as follows:
* The post-install should check for libreadline6 and if not existent try linking libreadline6 to 7
```sh
      libreadline6=$(find /usr/lib -name libreadline.so.6)
      libreadline7=$(find /usr/lib -name libreadline.so.7)
      if ( test -z $libreadline6 && ! test -z $libreadline7 )
      then ln -sfT ./libreadline.so.7 ${libreadline7/so.7/so.6}
      fi
```
* It should only run 'setup.py install' if the user agrees with 'y'. If it was previously installed that way the script may default to yes in non-interactive mode.
* In case' setup.py install' went wrong. It should skip 'setup.py install' with a warning (non-interactive) or query the user again. 'setup.py install' may fail on debian-buster if 'python3-distutils' is not installed.

zack-vii:
We could start building on jessie which seems to be compatible upto most recent buster.
Eventually there is a libreadline7 for wheezy by now. its just at the end of its lifecycle so its a bit of work to get a working repo from an archive

zack-vii:
can we then eventually disentangle wheezy32 and wheezy in that process.
in Docker repo i would sort it as
```
debian/wheezy/32
debian/wheezy/64
debian/jessie/32
debian/jessie/64
```
and in mdsplus repo, I would like to split the os parameter debian_wheezy into to  {debian7-32,debian7-64,debian8-32,debian8-64}

tfredian:
Whatever you think is appropriate. We don't have any debian usage at the PSFC as far as I know and you might have one of the biggest debian deployment there.

zack-vii:
Thanks, I will set up the jessie docker images and setup deployment accordingly.

zack-vii:
well for compatibility i had to follow thru with the distname but we could think of creating links similar to alpine
old
```
deb http://www.mdsplus.org/dist/DebianWheezy/repo MDSplus alpha
```
new
```
deb http://www.mdsplus.org/dist/debian/wheezy/repo MDSplus alpha
deb http://www.mdsplus.org/dist/debian/jessie/repo MDSplus alpha
deb http://www.mdsplus.org/dist/debian/stretch/repo MDSplus alpha
deb http://www.mdsplus.org/dist/debian/buster/repo MDSplus alpha
[...]

PR #1825, opened 2019-10-14
Fix: debian python post-install

zack-vii:
should fix most of #1824


zack-vii:
works on a fresh installation of buster

tfredian (approved):
Seems ok. Should we add another debian os version support which would remove the need to deal with the libreadline version?

PR #1826, opened 2019-10-15
Build: added debian stretch

zack-vii:
added debian stretch 64bit and split debian wheezy


zack-vii:
the jenkins job should be updated after merge

in job lists replace 'debian_wheezy' with 'debian7-32,debian7-64,debian9-64'

zack-vii:
once both pr are committed we could remove the debian_wheeze opts and job

tfredian (approved):

PR #1827, opened 2019-10-15
Build: added debian stretch stable

zack-vii:
fixes for python post-installer and stretch


zack-vii:
the jenkins job should be updated after merge

in job lists replace 'debian_wheezy' with 'debian7-32,debian7-64,debian9-64'

tfredian (approved):

PR #1830, opened 2019-10-21
Fix: repair python TreeNode.doMethod() method

### LINKED ISSUE #1829: doMethod method of python TreeNode is broken
The doMethod python method of the TreeNode object is broken in a couple different ways.

1) It is referencing _dsc.descriptor_xd() which apparently has been renamed to _dsc.Descriptor_xd()
2) it call _TreeDoMethod with an argument of _ver.MdsEND_ARG which no longer exists.

When the use of MdsEND_ARG was removed from MDSplus this function was not adjusted. There
is now a _TreeDoMethodA function which includes a argument count instead of relying on a variable length argument list.

tfredian:
A change to the python TreeNode implementation broke the doMethod
method. This should fix #1829

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -1648,9 +1648,12 @@ class TreeNode(_dat.TreeRef,_dat.Data): # HINT: TreeNode begin  (maybe subclass 
         xd=_dsc.Descriptor_xd()
         argsobj = [_scr.Int32(self.nid),_scr.String(method)]
         arglist+=list(map(_dat.Data.byref,argsobj))
-        arglist+=[len(args)]
-        argsobj= list(map(_dat.Data,args))
-        arglist+= list(map(_dat.Data.byref,argsobj))
+        arglist.append(len(args))
+        if len(args)>0:
+            argsobj= list(map(_dat.Data,args))
+            arglist.append(_N.array(map(_dat.Data.byref,argsobj)))
+        else:
+            arglist.append(_C.c_void_p(0))
         arglist+=[xd.ref]
         _exc.checkStatus(_TreeShr._TreeDoMethodA(*arglist))
         return xd._setTree(self.tree).value
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -1651,6 +1651,7 @@ class TreeNode(_dat.TreeRef,_dat.Data): # HINT: TreeNode begin  (maybe subclass 
         arglist+=[len(args)]
         argsobj= list(map(_dat.Data,args))
         arglist+= list(map(_dat.Data.byref,argsobj))
+        arglist+=[xd.ref]
         _exc.checkStatus(_TreeShr._TreeDoMethodA(*arglist))
         return xd._setTree(self.tree).value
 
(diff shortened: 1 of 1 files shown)


zack-vii (changes requested):
I have to check tomorrow but xd's ref seems to be missing in the args

tfredian:
Yep, xd.ref was missing from the the arglist. I just added it.

tfredian:
Hmmm,... this needs some testing. I tried doing a part_name method on a device and it segfaulted.

zack-vii:
I fixed it today but did not get the chance to commit. will do so in an hour or so

tfredian:
oops... I was working on the fix at the same time. In trying to test this with `Tree('subtree',-1).A14.INPUT_1.doMethod('part_name')`
I found some other unexpected results. I'm not sure if the original TreeDoMethod worked with things like part_name but the existing one does not. The part_name method for most of the C based devices uses StrCopyDx which doesn't support an xd for the output. It may not be worth fixing this issue. Not sure how the original part name strings are generated for C devices.

zack-vii:
no worries, I just had to do a quick fix so a colleague could use mdsplus to do some test shots. my doMethod

```python
         arglist=[self.ctx]
         xd=_dsc.descriptor_xd()
        argsobj = [_scr.Int32(self.nid),_scr.String(method),None]
         argsobj+= list(map(_dat.Data,args))
         arglist+= list(map(_dat.Data.byref,argsobj))
        arglist[2] = _C.Int32(len(args))
        arglist+= [xd.ref]
        _exc.checkStatus(_TreeShr._TreeDoMethodA(*arglist))
         return xd._setTree(self.tree).value
```
edited

tfredian:
The original part name code is calling _TreeDoMethod and passes a descriptor_d for the answer so that is still working. Not sure we need to change things for a python doMethod() implementation but I'm not entirely sure why the TreeDoMethodA routines stick the status in the output xd since methods should be able to load an answer in the output xd. The routine itself returns the status so if the status is ok the method implementation is responsible for loading the output xd.

zack-vii:
it was so it can be used for amy method. returning mdsplus objects. not only status

zack-vii:
I have to cjeck in _c.Int32 should be _scr.Int32

zack-vii:
I updated the code

zack-vii:
it only converst status into xd for c methods as thatvwa the only thing they returned sobfar. we could change that if required. the xd was maily meant for python methods

tfredian:
You update the code where??? This PR only has commits from me and it seems to work. I only tested it with part_name which uncovered 1) the part name code doesn't work with an xd output and 2) the _TreeDoMethodA implementation overwrites anything put in the xd with the status return of the method call. If I change the macro used by C devices to use MdsCopyDxXd instead of StrCopyDx the contents of the xd in _TreeDoMethodA does contain the string but that immediately gets overwritten with a dtype_long with the status return. I have not tested the handling of args with the _TreeDoMethodA call. Your suggested code in the comment above will not work as _TreeDoMethodA expects to be called with ctx, desc(nid),desc(method),len(nargs),pointer-to-array-descr-ptrs,xd.

zack-vii (approved):

tfredian:
As I said it might not be worth "fixing". I suspect there won't be many new C devices implemented if any and I think the only methods in existing C device that returned data was part_name and GetNci is using _TreeDoMethod and passing a descriptor_d for the output and works.

zack-vii:
we should chsnge the c methods part of TreeDoMethodA to leave xd untouched then. and return status instead if status not ok. basically treat it like py and tdi methods

PR #1831, opened 2019-10-21
Feature: Add python TreeNode.tcl() method

tfredian:
Add tcl method to TreeNode which uses the TreeNode as the current
default node before executing the command.


zack-vii (approved):

tfredian:
thx

PR #1833, opened 2019-10-22
Fix: Allow py doMethod('part_name') on C devices

tfredian:
The  C based devices expected a CLASS_D descriptor for
the output of the "part_name" method. This change will
allow the method to output to either a CLASS_D or a CLASS_XD
descriptor.

With this change one can do things like:

   partname=devicenode.doMethod('part_name')

This change also adds some tests for the doMethod('part_name') and
getnci(node,'original_part_name') with C devices.


zack-vii (approved):
looks ok now

PR #1836, opened 2019-10-24
Build: updated updatepkg script to use same fc version as bootstrap a…

zack-vii:
In case @GabrieleManduchi is not quite done yet with his  MARTe2 streaming PR.


zack-vii:
retest this please

tfredian (approved):

PR #1837, opened 2019-10-24
Fix: plot position when plotting SegmentScale

zack-vii:


santorofer (approved):

santorofer:
Hey Timo,
What does the new paramter to "GraphPanel.newPlot" do? I mean:
JOptionPane.getRootFrame()

zack-vii:
it returns the reference to the parentin frame. to the plotting routine this will mean it opens the window near the caller window instead centered of the currently acrive desktop. basically it prevents the avtive area from jumping around.

PR #1846, opened 2019-11-06
Fix: trunc files when doing new on MDS_IO_OPEN_ONE

zack-vii:
* if you did create pules or edit tree/new, the treefiles would not get truncated but hold the old data if the file existed.


zack-vii:
was justva typo in the test tgat did not matter under linux.

tfredian (approved):

merlea:
Hello,
Thanks for the fix. It would be nice to see this make it in stable too, especially for applications within IMAS.
Best,
Antoine.

PR #1852, opened 2019-11-12
Fix: trunc files when doing new on MDS_IO_OPEN_ONE (#1846)

tfredian:
* Fix: trunc files when doing new on MDS_IO_OPEN_ONE

* Tests: test the O_TRUNC on new


zack-vii (approved):

zack-vii:
I would approve it but self-approved ?

PR #1863, opened 2019-11-27
Fix: TCL DO/METHOD /IF option problem and TCL HELP improvements

tfredian:


zack-vii (approved):
looks good. we should add a test that passes 1 and 2 args to a method.

PR #1870, opened 2019-12-04
Fix:time context in ScopeUtilities

GabrieleManduchi:
Reset time context was called too early, in some cases before actually evaluating the signal


zack-vii (changes requested):
no it should be reset close to the end so it will be reset even if an error occurs.

GabrieleManduchi:
true...it was in the right place....I was misled by other factors, I close the PR

PR #1874, opened 2019-12-13
Fix: python37 initializes threads already during Py_InitializeEx

zack-vii:
This PR deals with the new behaviour of python3.7 which initializes thread during Py_InitializeEx automatically. hence the check is true hence we dont save the ThreadState .. hence calling a python method wil not restore the correct GIL .. hence we run into a deadlock.
this fixes teh init order and conditioning to properly save the threadstate for all python version and both cases.. main is non/python, e.g. tditest, and main is python calling tdi calling python again.

### DIFF
--- a/tdi/dev_support/PyDoMethod.py
+++ b/tdi/dev_support/PyDoMethod.py
@@ -27,22 +27,26 @@ from MDSplus import TreeNode, Device
 from MDSplus import TreeNOMETHOD,MDSplusException,PyUNHANDLED_EXCEPTION
 from sys import stderr,exc_info
 
+
+def domethod(methodobj,args):
+    try:
+        return methodobj(*args)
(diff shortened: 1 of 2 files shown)


joshStillerman:
This looks good, but a quick conversation / explanation would be helpful

zack-vii:
This PR deals with the new behaviour of python3.7 which initializes thread during Py_InitializeEx automatically. hence the check is true hence we dont save the ThreadState .. hence calling a python method wil not restore the correct GIL .. hence we run into a deadlock.
this fixes teh init order and conditioning to properly save the threadstate for all python version and both cases.. main is non/python, e.g. tditest, and main is python calling tdi calling python again.

joshStillerman (approved):

PR #1875, opened 2019-12-14
Fix: nullpointer access in ScopeUtilities if no limits specified

zack-vii:


GabrieleManduchi:
Not sure that the use of int16_t in ScopeUtilities (line 662 old, 631 new)  is correct........

zack-vii:
good find, fixed it

GabrieleManduchi:
Another thing: lines

 *(int *)&retArr[idx] = yLabel->length;	 
    swap4(&retArr[idx]);	

are replaced by 
   SWAP32(&retArr[idx],&yLabel->length);

but  it seems to me that  yLabel->length is a short, and in the replacement it is considered as long

zack-vii:
This is just like in the original code. Since length is unsigned but java can only handle signed length is written out as int. Thus the full int has to be swapped for java is bigendian

GabrieleManduchi:
ylabel->length is a short. I agree that we possibly need swapping but instruction  SWAP32(&retArr[idx],&yLabel->length); copies (possibly swapped) bytes that do not belong to yLabel->length (2 bytes).

Am I missing anything?

GabrieleManduchi:
Seems ok now, but just one last question: why do you mask length field (short) with 0xffff?

zack-vii:
to make sure its unsigned. its also doing the cast to int. i am not sure if this is necessary but it is a more obvious variant of casting to int that tells you: 'this is an unsigned short'.

GabrieleManduchi (approved):

PR #1876, opened 2019-12-14
Fix: race condition in MDS_IO_(UN)LOCK

zack-vii:
* added easier constructor for recursive locks


joshStillerman:
Can we talk about this one also ?

zack-vii:
the lock has to guard the MDS_IO_LOCK call. never seemed to be a problem. but the removal of the tdicompile lock made this flaw apparent

joshStillerman (approved):

PR #1878, opened 2019-12-14
Fix: improved performance of critical function Lib*Vm routines

zack-vii:


joshStillerman (approved):
as discussed in Padova - we should work out CLANG coding standards

PR #1868, opened 2019-12-03
Fix: fixed performance issues

zack-vii:
* performance of tdicompile increased
  - GET_TDITHREADSTATIC_P if costly so we pass the pointer to statics when we can.
  - LibResetVmZone and LibVmGet were slow
  - removed broken old code that is obviously not needed anymore (e.g. use_get_record_fun)
test case
```bash
time for i in `seq 1 100`; do tditest /git/mdsplus/tditest/testing/test-tdishr.tdi >/dev/null; done
```
before
```
real    0m10.142s
user    0m8.580s
sys     0m0.380s
```
after
```
real    0m8.115s
user    0m6.548s
sys     0m0.368s
```

PR #1880, opened 2019-12-15
Build: removed exports and def from source.

zack-vii:
* build them with gendef on-the-fly from dll when needed in mdsobjects/cpp instead


zack-vii:
retest this please

zack-vii:
retest this please

zack-vii:
get rid of the def files that are used for visual studio and generate them during build from the dlls via gendef. this avoids mysterious errors due to outdated def files.

This PR may require a rebase after pr #1868 is merged

joshStillerman (approved):

PR #1881, opened 2019-12-15
Fix: split pthread into logical sections

zack-vii:
pthread_port.h has become pretty big and some of teh stuff is only remotelly pthread related. it stated out as a port for pthread to get windows on board but extended into a general header for all kinds of stuff. this just splits it into its topics, condition, socket stuff pthread-stuff. his requires a rebase if the nci lock is merged

### DIFF
--- a/include/getusername.h
+++ b/include/getusername.h
@@ -1,0 +1,55 @@
+#ifndef GETUSERNAME_H
+#define GETUSERNAME_H
+#ifdef HAVE_UNISTD_H
+ #include <unistd.h>
+#endif
+#ifdef _WIN32
+ #include <io.h>
+#else
+ #include <pwd.h>
+#endif
+#include <pthread_port.h>
+
+#define GETUSERNAME(user_p) GETUSERNAME_BEGIN(user_p);GETUSERNAME_END;
+
+#define GETUSERNAME_BEGIN(user_p) {\
+static pthread_mutex_t username_mutex = PTHREAD_MUTEX_INITIALIZER;\
+pthread_mutex_lock(&username_mutex);\
+if (!user_p) {\
+  user_p = _getUserName()
+
+#define GETUSERNAME_END }\
+pthread_mutex_unlock(&username_mutex);\
+}
+static char* _getUserName(){
+  char *user_p;
+#ifdef _WIN32
+    static char user[128];
+    DWORD bsize = 128;
+    user_p = GetUserName(user, &bsize) ? user : "Windows User";
+#elif __MWERKS__
+    ans.pointer = "Macintosh User";
+#else
+    static char user[256];
+    struct passwd *pwd = getpwuid(geteuid());
+    if (pwd) {
+      strcpy(user,pwd->pw_name);
+      user_p = user;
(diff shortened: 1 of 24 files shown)


joshStillerman (dismissed):

include/getusername.h
```diff
@@ -0,0 +1,55 @@
...
+#ifdef _WIN32
+ #include <io.h>
+#else
+ #include <pwd.h>
+#endif
+#include <pthread_port.h>
+
+#define GETUSERNAME(user_p) GETUSERNAME_BEGIN(user_p);GETUSERNAME_END;
+
+#define GETUSERNAME_BEGIN(user_p) {\
```
WhoBrokeTheBuild, line 15: Why are these split into two macros?

mdsshr/librtl.c
```diff
@@ -58,26 +48,25 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
+ #define setenv(name,value,overwrite) _putenv_s(name,value)
+ #define unsetenv(name)               _putenv_s(name,"")
+ #define localtime_r(time,tm)         localtime_s(tm,time)
+ #define ctime_r(tm,buf)              ctime_s(buf,32,tm)
 #else
-#include <sys/wait.h>
+ #include <sys/wait.h>
 #endif
 
+#include <STATICdef.h>
```
WhoBrokeTheBuild, line 60: Didn't you just remove this?

mdstcpip/ioroutines.h
```diff
@@ -8,54 +8,25 @@
...
-#endif
-#define SEND_BUF_SIZE 32768
-#define RECV_BUF_SIZE 32768
-#ifndef MSG_NOSIGNAL
- #define MSG_NOSIGNAL 0
-#endif
-#define LOAD_INITIALIZESOCKETS
-#include <pthread_port.h>
+
+#include <STATICdef.h>
```
WhoBrokeTheBuild, line 26: Same question as above

mdstcpip/ioroutinesV6.h
```diff
@@ -10,55 +10,24 @@
...
-#endif
-#define SEND_BUF_SIZE 32768
-#define RECV_BUF_SIZE 32768
-#ifndef MSG_NOSIGNAL
- #define MSG_NOSIGNAL 0
-#endif
-#define LOAD_INITIALIZESOCKETS
-#include <pthread_port.h>
+
+#include <STATICdef.h>
```
WhoBrokeTheBuild, line 28: Same question as above

WhoBrokeTheBuild (approved):
LGTM, I question the static def header though.

include/getusername.h
zack-vii: it is used in two parts in one file.. need tro check where. so its a constructor;
then the code can safely do stuff with teh static username
and a cleanup.

zack-vii (commented):
I did but in tdishr.. we should remove them everywhere but this would have blown up this PR as the once before

PR #1884, opened 2019-12-17
Build: add debug information for alpha builds

zack-vii:
added -g to the compiler without breaking the logic of configure simply by using the 
'--enable-debug=info' parameter if branch is alpha. worked locally but the jenkins test are not on branch alpha so.. its hard to test.
nonetheless changing the -O to -Og revealed more uninitialized compiler warnings

But then i had to drop -Og because some OS g++ did not support it.


joshStillerman (dismissed):

zack-vii:
retest this please

joshStillerman (dismissed):

joshStillerman (approved):

PR #1887, opened 2019-12-20
Fix:  Infinite loop in host_list_cleanup

joshStillerman:
treeshr/RemoteAccess.c
    host_list_cleanup  had a loop which checked on existing connections
    in the case that the connection at the top of this list was not
    the one it was looking for, the 'continue' statement caused the loop
    iteration to complete without changing the loop boolean (host).

    added code to move host to next element in list


joshStillerman:
retest this please

zack-vii (approved):
look good

PR #1894, opened 2020-01-04
Tests: enable debian9 memcheck

zack-vii:
since fc26 builds for both 64 and 32 bit memcheck takes quite long with little benefit. Detected extra issues are expected due to different python/java/libc versions and should be suppressed anyway.
Debian9 is a modern os that can test on 64 bit with a recent valgrind version. 
Its much faster, i.e.  <20 min compared to >50 min on fc26.

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -471,9 +471,8 @@ class Tree(object):
         @param mode: Optional mode, one of 'Normal','Edit','New','Readonly'
         @type mode: str
         """
-        if tree is None:
-            self.public = True
-        else:
+        self.public = tree is None
+        if not self.public:
             if path is not None: self.path = path
             self._ctx = _C.c_void_p(0)
             self.tree = tree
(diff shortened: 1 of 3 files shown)


python/MDSplus/tree.py
```diff
@@ -489,7 +488,7 @@ def __enter__(self):
     def __del__(self):
         if not self.public:
             self.__exit__()
-            _TreeShr.TreeFreeDbid(self.ctx)
+            _TreeShr.TreeFreeDbid(self._ctx)
```
WhoBrokeTheBuild, line 491: Why are some `self.ctx` and others `self._ctx`?

zack-vii: ctx is the public access to _ctx and may redirect to TreeCtx() is public is True. _ctx is the private field used if public is False .. if not self.public -> avoid double check

WhoBrokeTheBuild (approved):

PR #1895, opened 2020-01-04
Tests: setup for python3.5 on debian9-64

zack-vii:
#1894 for stable  ... conflict resolved by adding pyhton3.7 suppressions for fc30

### DIFF
--- a/python/MDSplus/tests/_UnitTest.py
+++ b/python/MDSplus/tests/_UnitTest.py
@@ -42,6 +42,7 @@ class logger(object):
 
 class Tests(TestCase):
     debug = False
+    is_win = sys.platform.startswith('win')
     in_valgrind = 'VALGRIND' in os.environ
     inThread = False
     index = 0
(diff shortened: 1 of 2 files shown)


python/MDSplus/tests/_UnitTest.py
```diff
@@ -274,7 +275,7 @@ def tearDownClass(cls):
                 shutil.rmtree(cls.tmpdir)
     def tearDown(self):
         gc.collect()
-        if self.inThread: return
+        if not self.is_win or self.inThread: return
```
WhoBrokeTheBuild, line 278: What's the purpose of this?

zack-vii: under windows the file locking is different and for a proper test traceback we need to make sure we release the tree files even in case of an error. otherwise it may cause an error in a subsequent test of the same module. under linux we dont have to and we should not because python seems to double __del__ for some versions. valgrind would complain but its an issue with pyhton

WhoBrokeTheBuild (approved):

PR #1897, opened 2020-01-07
Fix: mdsip issue with ThreadStatics - MDSplusThreadStatic

zack-vii:
* combine all ThreadStatics in shared lib mdsshr so the thread state can be shared with child threads like in ProcessMessage, i.e. parent waits for Thread to finish before continuing (Thread is used to implements cancelation)


zack-vii:
retest this please

joshStillerman (dismissed):
Looks good

joshStillerman (approved):

PR #1899, opened 2020-01-08
Build: debian build script for py3

zack-vii:


tdishr/TdiExtPython.c
```diff
@@ -110,7 +110,7 @@ inline static void initialize(){
   void *handle;
   char *lib;
   char *envsym = getenv("PyLib");
-  if (!envsym) {
+  if ( !envsym || envsym[0] == '\0' ) {
```
joshStillerman: is your intent  ( !envsym || ( envsym[0] == '\0') ) ? 
is this equivalent to that ?
adding the extra () would make it more clear

zack-vii: it sure is

joshStillerman: Can you add the extra () for clarity. The precedence is confusing.
` ( !envsym || ( envsym[0] == '\0') )` !=  ` ( (!envsym || ( envsym[0]) == '\0')`

joshStillerman (approved):

PR #1912, opened 2020-01-31
Fix:empty segments crashed the new version of XTreeGetRecord

GabrieleManduchi:
The crash occurred in Squish section. A check has been added before in order to pass only non empty segments to Squish function.

### DIFF
--- a/xtreeshr/XTreeGetTimedRecord.c
+++ b/xtreeshr/XTreeGetTimedRecord.c
@@ -121,7 +121,7 @@ EXPORT int XTreeGetTimedRecord(int inNid, mdsdsc_t *startD,
 				mdsdsc_t *endD, mdsdsc_t *minDeltaD,
 				mdsdsc_xd_t  *outSignal){
   int status, nid;
-  int actNumSegments, currSegIdx, numSegments, currIdx, numDimensions;
+  int actNumSegments, currSegIdx, nonEmptySegIdx, numSegments, currIdx, numDimensions;
   int i, nameLen, startIdx, endIdx;
   char resampleFunName[MAX_FUN_NAMELEN], squishFunName[MAX_FUN_NAMELEN];
   char resampleMode[MAX_FUN_NAMELEN];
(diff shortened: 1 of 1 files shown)


xtreeshr/XTreeGetTimedRecord.c
```diff
@@ -341,7 +341,11 @@ EXPORT int XTreeGetTimedRecord(int inNid, mdsdsc_t *startD,
 
     MdsFree1Dx(&dataXds[currSegIdx], 0);
     MdsFree1Dx(&dimensionXds[currSegIdx], 0);
-    signals[currSegIdx] = (mds_signal_t *)resampledXds[currSegIdx].pointer;
+//After resampling some segments may be empty: consider only nonempty segments
+    if(resampledXds[currSegIdx].l_length > 0)
+      signals[nonEmptySegIdx++] = (mds_signal_t *)resampledXds[currSegIdx].pointer;
+    else //Keep track that the segment is not considered
+      signalsApd.arsize -= sizeof(mds_signal_t *);
```
zack-vii: shoukd it not better be using the length field of signalsApd instead of sizeof. best would be though to calculate arsize based on nonEmptySegIdx an length after the for loop.

zack-vii: we should add some tests that tackle those kind of edge cases because xtreeshr is used frequently and its a pitty its not tested well. i may be able to contribute. we should add these tests to treeUnitTest where i test the timecontext stuff. maybe there is a lightweight example that would trigger the issue w/o preexisting tree files.

zack-vii (changes requested):
the arsize should be calculated at the end of the loop.
and we need tests. but the can follow in a different pr

xtreeshr/XTreeGetTimedRecord.c
GabrieleManduchi: Ok, removed length update after the loop

zack-vii (approved):

PR #1928, opened 2020-02-12
Fix: mdsip hostfile mapping

zack-vii:
skip rule if cannot map_to_user
allows
*@server | MAP_TO_USER
*@server | nobody

which maps known users to local accounts and all other to nobody instead of denying immediately.

### DIFF
--- a/mdstcpip/CheckClient.c
+++ b/mdstcpip/CheckClient.c
@@ -114,10 +114,8 @@ static int become_user(const char *remote_user, const char *local_user)
       } else
 	access = ACCESS_GRANTED;
     }
-  } else {
-    fprintf(stderr,"Invalid mapping, cannot find user \"%s\"\n", user);
-    access = ACCESS_DENIED;
-  }
+  } else
+    access = ACCESS_NOMATCH;
   return access;
 }
 #endif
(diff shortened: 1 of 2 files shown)


mdstcpip/CheckClient.c
```diff
@@ -114,10 +114,8 @@ static int become_user(const char *remote_user, const char *local_user)
       } else
 	access = ACCESS_GRANTED;
     }
-  } else {
-    fprintf(stderr,"Invalid mapping, cannot find user \"%s\"\n", user);
-    access = ACCESS_DENIED;
-  }
+  } else
```
WhoBrokeTheBuild, line 117: This should have brackets

mdstcpip/ParseCommand.c
```diff
@@ -140,8 +170,8 @@ void ParseStdArgs(int argc, char **argv, int *extra_argc, char ***extra_argv)
       {"h", "hostfile", 1, 0, 0},
       {"s", "server", 0, 0, 0},
       {"m", "multi", 0, 0, 0},
-      {"c", "compression", 1, 0, 0},
-      {"?", "help", 1, 0, 0},
+      {"c", "compression", 2, 0, 0},
```
WhoBrokeTheBuild, line 173: What do these changes do?

WhoBrokeTheBuild (changes requested):

mdstcpip/ParseCommand.c
zack-vii: 2 allows optional value
1 demands value

WhoBrokeTheBuild (approved):

PR #1931, opened 2020-02-13
Fix: TdiLex failed to match comment

zack-vii:
regexp swallowed * of */ when comment ended on "**"+"/"
i.e. an even number of '*' followed by '/'


joshStillerman (approved):
works like a champ.  THANKS!

PR #1933, opened 2020-02-19
Move Visual Studio files from 2017 to 2019

HerveAncher:


joshStillerman:
Someone else may want to speak to this.... As far as I know we do all of 
our windows builds using cygwin.  If you have it building in visual 
studio, and the files you changed are not ones we use for the standard 
releases, I am fine with updating them.  You may want to add a section 
under 'building from source' on www.mdsplus.org describing how to build 
MDSplus using visual studio.

On a separate track:  I am surprised to see our previous commits in your 
Commit summary.  I am guessing that a rebase instead of a merge would 
have eliminated them.  Does anyone know if we NEED TO REBASE to accept 
this PR ?

-Josh

HerveAncher:
About building MDSplus with VS, yes, we need to first build library with Cygwin or MinGW (I use MinGW) and after generate the compliant library with the script I have modified.
The goal is to have libraries compatibles with MSVC to be able to link directly under VS.

For the PR, you right, if think a rebase would be better. Maybe my forked repo was not up to date.

tfredian:
A couple comments on this PR. This looks good to me but  perhaps Gabriele should take a look at the java Makefile improvements and approve after the PR is updated. The MDSplus C based libraries are indeed built with mingw64 not cygwin. Unfortunately C++ based libraries built with mingw64 are not compatible with applications built using VS so we provide both VS versions and MingW versions of those libraries. Perhaps the easiest way to rebase this is to pull the latest alpha and make another branch and cherry pick your commit from this branch into the new branch and force push the new branch to replace this branch. 

Another important comment is that one of the core development team must respond to the automatic comment about the "ok to txxx". For security reasons the PR will not be automatically tested if it was created by someone not a member of the MDSplus development team. A core developer should review the changes to make sure there are no suspicious security hacks in the code and if all looks good the developer should enter a comment on the PR with the "ok to txxx" phrase in it so the PR will be properly tested.

zack-vii:
@HerveAncher  You should definitely rebase this branch from current alpha so it is clear what actually changed. Also it o would be good to know why you thing this is necessary to add. Whats the purpose of adding vs files to the branch and why is it required to change the makefiles. I assume the vs stuff is a nice to have so users can quickly build extensions with vs but i dont see why that would be necessary, especially since we apparently have 2017 project files already. Could one not use those in vs 2019 and upgrade. users still using 2017 would not be able to use 2019 project files though.
If you wish to ad them anyway we should put them in a dedicated folder that is not touching or touched by the build chain. Please explain what difficulties you try to overcome.
First though rebase, please. It might help understanding.

zack-vii (changes requested):
I doubt we can have this changed as this is used in our build chain. Your version looks more like a manually executed script in a proper Windows environment with very specific layout.

zack-vii:
Hm well you were supposed to fix issues that occured during the rebasing process. that is clearing out all the >>>>.. and <<<<.. stuff. I also have the feeeing you rebased against your local alpha branch. I will try to get it right for you. just explain what you are trying to solve in this pr.

zack-vii:
Ok i cecked it out and test it. we wont be able to use this pr. it contains soecifics to essantial build scripts that wont work for us. 
* removal of dual arch support.
* change of filepath that are required to find build bins on wine

There is at least one errors/warnings in the cpp code aboit unused vars.
For local testing, please use docker and our building tools:
```sh
deploy/build.sh --os=windows --test --release
```
This will have to run thru without any error and of course it is required to produce a complete and working installer for windows.

zack-vii:
For a propperly rebaded branch please reset --hard to 
https://github.com/MDSplus/mdsplus.git HerveAncher-VS2017-2019
or manually rebase 
```sh
# update your upstream
git fetch https://github.com/MDSplus/mdsplus.git # may work
# reset to your original "git reset --hard 90c308d
git rebase alpha_release-7-97-0 -i
# the editor, remove all commits but yours. should be the last 5
```
The force push.

HerveAncher:
Thanks for the steps to properly rebase.

zack-vii (commented):
commit 686e0d1 may be ok, might require the changes of the other commits though

zack-vii:
Since you did not tell us what you are trying to accomplish, I will make a wild guess. You wish to use the event stream under Windows. how about we include the code and wrap it in a macro e.g. if def _HAVE_PTHREAD_H_ or similar. so you can build it in you vs environment. since we do not have pthread in our wine environment yet we wont be able to include it by default.
would that be sufficient. I doubt you are using the batch files to build but if you do. I would like to ask you to store them with your project as they are windows specific and wont work in our wine environment.

zack-vii:
what we can take away from this i gathered in PR #1935

PR #1934, opened 2020-02-19
Update influxhistorian device class

WhoBrokeTheBuild:
- Surround `import influxdb` with try/catch to fail cleanly when the package is not installed
- Change `DATA_COUNT` to 500
- Add `CREDENTIALS` file support
- Cleanup `LAST_READ` and add `START_TIME`
- Generalize `TAG_PATH` to `WHERE` and `FUNCTION` to `SELECT`
- Use proper prameters to have influx return milliseconds, instead of trying to parse timestamps


pydevices/HtsDevices/influxhistorian.py
```diff
@@ -1,39 +1,47 @@
 import MDSplus
 import time
 from datetime import datetime
-from influxdb import InfluxDBClient
+
+try:
+    from influxdb import InfluxDBClient
+except:
+    print("You must install the `influxdb` python package to use the `influxhistorian` device class")
+    exit(1)
```
joshStillerman: Should throw an error not exit

zack-vii: good find. In general if you design a python device driver you should consider loading third party packages when you need them i.e. in the Device methods as a static field of the class. Reason is the PyDevice search engine will preload the modules when looking for any pydevice this may take significantly longer if devices take longer to initially load. 
The device is also loaded when instantiating nodes of the device in python.
You may want to be able to do that even if you dont have the third party packages. That should be fine and silent. it definitely should not exit as this would close you session when attempting to access a device node.

You could add a cached field to your device class.
```python
class MYDEVICE(Device):
    _influxdb = None
    @property
    def influxdb(self):
         if self._influxdb is None:
            self._influxdb = __import__("influxdb",fromlist=["InfluxDBClient"]).InfluxDBClient
         return self._influxdb
```

joshStillerman (approved):

PR #1922, opened 2020-02-10
Fix: use generalized MDSVERSION in each core lib

zack-vii:
* reactivated MdsRelease / MdsReleaseDsc
* unified _version.py
* added for some more libs
```tdi
 MdsShr->MdsRelease:T()
 TreeShr->TreeRelease:T()
 TdiShr->TdiRelease:T()
 XTreeShr->XTreeRelease:T()
 MdsIpShr->MdsIpRelease:T()

 Mdsdcl->MdsdclRelease:T()
 MdsMisc->MdsMiscRelease:T()
```

PR #1935, opened 2020-02-25
Feature: datastream for windows

zack-vii:
extracted from #1933


zack-vii:
retest this please

zack-vii:
retest this please

zack-vii:
requires alpine patch from #1922

HerveAncher:
Dear zack-vii,

Thanks for these adaptations. In the PR #1933 I effectively not specify what was my goal.
So here is it: We use MDSplus to store data as backend in specific project named IMAS.
Currently, I build MDSplus with MinGW and my previous commits in MDS were enough to do that.

Now, I would like to compile MDSplus with the last MSVC compiler, to compile all of my project with MSVC, and I have modified `visual-studio-build.bat` for this purpose.
Perhaps I need to defined some of my var (like VSVER, VCVER and WKVER) to be compliant by default with VS2017, the version you used.

The second point is the output directories (BUILD86 and BUILD64), I cannot use `\workspace\releasebld\` because on my Windows system, this directory does not exist.

The third point is the command `vsbuild` use in Makefile. What is it?
I directly compile under Windows, so I do not use wine. I think if Wine if available, I can use it ans call your command `vsbuild` and in other case call directly `visual-studio-build.bat`

So I propose to make some adaptations by letting the default value for the parameters as there are, and check if some of them are passed to the Batch and use them in this case.

Now I will wait until this PR will accepted and merged and after that I will propose my adaptations as described above.

GabrieleManduchi (approved):

zack-vii:
i think the visual-studio-build.bat scipt should rather go into deploy/packaging/windows as it is only used in for release builds. I might be mistaken, but vsbuild could be a relique from days when we did not use wine yet. since your will you be building all mdsplus modules under vs or just mdsobjects. it might be best to ratger duplicate the batch script and call it *_win.bat where you can run it with arguments. the changes required to make both worlds happy in one file might just be too much. since we have to maintain the file it would be best to keep them simple or they may break when changes come around. if you feel like it, you could share your project files and we can work something out.
cheers
Timo

PR #1936, opened 2020-02-26
Build: run bootstrap during trigger.sh and removed output

zack-vii:
Now bootstrap includes also yacc/lex and gen_messages

We should move any other autogenerated code there as well.

users can run ./bootstrap to create the files w/o docker


zack-vii:
@MDSplus/core :
please have a look at it and vote if you think it is OK not to keep autoconf and automake output in the repo. everything can be build using 
```sh
./bootstrap
```

joshStillerman (approved):
sounds good to me

PR #1940, opened 2020-03-04
Fix: show example overview for wgsi even without trailing /

zack-vii:
before: http://server:80/mdsplus showed 404 error
now: http://server:80/mdsplus is the same as http://server:80/mdsplus/


zack-vii:
@joshStillerman  poke ;)

zack-vii:
retest this please

WhoBrokeTheBuild (approved):

joshStillerman (approved):

PR #1941, opened 2020-03-05
Manage compatible libraries compilation under VS2019

HerveAncher:
Dear all,

After my previous PR #1933 merged by @zack-vii in #1935 I come back with a new PR which is manage better the integration of VS2019 (and also VS2017).

Lot of files were modified because I called ```./bootstrap``` and ```./configure``` at root and this command regenerate several files, like Makefile.

The main modifications are there:

Manage __atribute__ for MSVC (mostly in .c and .h files):
```
#if defined(__GNUC__ ) || defined(__clang__)
#  define UNUSED_ARGUMENT __attribute__((__unused__))
#else
#  define UNUSED_ARGUMENT
#endif
```

Add new arguments to manage Visual Studio to ```configure``` script:
```
--with-visual-studio-ver=VS_VER
--with-visual-cpp-ver=VC_VER
-with-windows-sdk-ver=WK_VER
--with-windows-pthread-dir=WIN_PTHREADDIR
```

Fix asterisks escapement and comments in Lex files.

Normally, with these modifications, you should be able to compile MDSplus safety under Linux and Cygwin (I did it under Ubuntu 18.04 and MinGW-w64).

PR #1942, opened 2020-03-05
Build: added support for other visual studio compiler

zack-vii:
@HerveAncher : please check this one out and add the mdsplus/mdsconfig.h header where you have a conflict with \_\_attribute\_\_ stuff. You may find that mdsplus/mdsconfig.h is already included but only if WINDOWS_H is undefined. In that case, please move it out of this condition as it is handled in mdsconfig.h . just force push you fixed version to your branch in PR #1941. 

FYI: please follow the commit convention. the commit title should start with "Build: "


zack-vii:
branch name too long for version.h header resubmitted in #1943

PR #1949, opened 2020-03-10
Feature: added MDSplusAPI to MATLAB for optional access

zack-vii:
API does not require any setup other than adding the ./matlab folder to MATLAB path.
API comes with hands-on example.
API is portable and does not require any local binaries if remote connection is used.


zack-vii:
retest this please

zack-vii:
> retest this please

joshStillerman (approved):

PR #1950, opened 2020-03-11
Fix: old Matlab api; no javalibpath on modern matlab and python3 bridge

zack-vii:
- java bridge now works out of the box without adding paths to anything but matlab path
```matlab
addpath(fullfile(getenv("MDSPLUS_DIR"),"matlab"));
savepath();
```
- python bridge provides workaround for int64 on python2
- python bridge now works with python3 as well

briges are only tested when activated

mdstest(true)  % tests explicitly the python bridge
mdstest(false)  % tests explicitly the java bridge


zack-vii:
retest this please

joshStillerman (approved):

PR #1967, opened 2020-04-01
Fix: tdi parser to allow keywords for labels

### LINKED ISSUE #1966: Can not evaluate an expression which is a reference to a node with same name as keyword
We have a tree with a node named `\efit_geqdsk:case`.
If I pass this string to the expression evaluator everything proceeds as normal.
If however the current tree default is `\efit_geqdsk` and I ask for the value of the expression `case`, I get an error.
```
IDL> mdsopen,'analysis',1120725012
IDL> help, mdsvalue('\efit_geqdsk:case')
<Expression>    STRING    = Array[87]
IDL> mdssetdefault,'\efit_geqdsk'
IDL> help,mdsvalue('case')
% MDSVALUE: %TDI-E-SYNTAX, Bad punctuation or misspelled word or number
<Expression>    STRING    = '*'
IDL> 
```
The problem is the parser is finding the lex token CASE before seeing if it is a node.  Apparently this used to work.

zack-vii:
is supposed to fix #1966
However, puts back some shift/reduce conflicts.
Ideally one would write down all syntax and regroup but this seems unlikely as not all old code may be compatible with a clean parser.


joshStillerman (approved):
Wow -
  That was fast!  Thanks!  That yacc file is pretty scary! 
When you have a chance can you add a test for a node called 'CASE' ?

zack-vii:
its already in the pr

@zack-vii
Tests: case is supposed to be ok for path
9109e7e

PR #1969, opened 2020-04-05
Feature: tditest added help and quiet mode: added arguments

zack-vii:
you can do shebang
#!/usr/local/mdpslus/bin/tditest -q
and call the tdi file with arguments that will be stored  _$0 _$1 ... idends

### DIFF
--- a/tditest/tditest.c
+++ b/tditest/tditest.c
@@ -213,6 +213,31 @@ static char *getExpression(FILE *f_in) {
   return ans;
 }
 
+static void print_help(int error)
+{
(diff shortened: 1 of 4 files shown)


zack-vii:
this time docker was not running on mfews10.
fixed by logging onto it and executing systemctl start docker
retest this please

joshStillerman (approved):
nice

zack-vii:
the --quiet flag is also veryusefull when debuggin as you done end up debugging the Write(,DEBUG( ))
so you dont need to skip thru all of your breakpoints

PR #1973, opened 2020-04-07
Fix: dropped the concerns about conflicts

zack-vii:


joshStillerman (approved):
I assume this is OK and goodness....

PR #1974, opened 2020-04-08
Fix: makeSegments bug in check_compress_dim

zack-vii:
Resolves TREE-E-INVDTPUSG error on MakeSegment due to bug in check_compress_dim.

### DIFF
--- a/treeshr/TreeSegments.c
+++ b/treeshr/TreeSegments.c
@@ -869,7 +869,8 @@ inline static int check_sinfo(vars_t *vars) {
   return TreeSUCCESS;
 }
 
-inline static int begin_sinfo(vars_t *vars, mdsdsc_a_t *initialValue,  int checkcompress()) {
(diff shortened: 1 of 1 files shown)


joshStillerman (commented):
Timo,
  Can you explain this change?  Where does initialvalue come from ?

zack-vii:
sure: but i found that the method is called twice. need to go back to my desk.
this fixed writing but seems reading is now a problem
I will add some tests once i found the other issue

zack-vii:
strange that it is now covered by our tests

zack-vii:
well now its easier to explain.. begin_sinfo was calling checkcompress with 'mdsdsc_xd_t' but checkcompress expected 'mdsdsc_xd_t *'

zack-vii:
writing and reading works now.

joshStillerman (approved):

PR #1975, opened 2020-04-08
New D-TACQ devices for streaming and transient captures.

seanalsop:
Both acq1001 and acq2106 are included.
Also included is a new copy of acq400_hapi, which (some of) the code in the devices relies on.


joshStillerman:
ok to test

zack-vii:
rebase this please

zack-vii:
also unfortunatly teh branch name can only have 12 chars

zack-vii:
and if you are at it. it would be great if the devices would work with both python2 and 3
e.g.:
```python
if sys.version_info<(3,):
    import Queue as queue
else:
    import queue
```

zack-vii:
PR :  #1976 has these issues addressed and replaces this PR

PR #1976, opened 2020-04-08
New D-TACQ devices for streaming and transient captures. (rebase)

zack-vii:
#1975


zack-vii:
we should not ship https://github.com/petermilne/acq400_hapi
at least not as proper part but as submodule.
I would suggest to throw a warning if module is not found mentioning this url

zack-vii:
still should be pyhton3 ready

zack-vii:
retest this please

zack-vii:
given that acq400_hapi is also used by acq400Devcies it is advised not to ship in at submodule of HTSDevcies.

also packaging need to be updated. we should combine all pure python pydevices into package mdsplus-pydevices.

zack-vii:
fixed acq400Devices so they would work with and without acq400_base also fixed Queue

pydevices/acq400Devices/acq1001_st.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/python
```
joshStillerman: The previous is 'more correct'. The interpreter does not necessarily live a /usr/bin/python

pydevices/acq400Devices/acq1001_tr.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/python
```
joshStillerman: same comment as previous

pydevices/acq400Devices/acq2106_st.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/python
```
joshStillerman: ditto

pydevices/acq400Devices/acq2106_tr.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/python
```
joshStillerman: again

joshStillerman (approved):

PR #1982, opened 2020-04-10
Fix: python Event.stream not using signal name

joshStillerman:
The python entrypoint for Event.stream always used the eventname
'STREAMING'.  Changed to use the input argument 'signal'


joshStillerman:
name to long

PR #1992, opened 2020-04-16
Fix: getShotDB use default_tree_path if <tree>_path not found

zack-vii:


python/MDSplus/tree.py
```diff
@@ -311,7 +311,9 @@ def getShotDB(expt,path=None,lower=None,upper=None):
         expt = str(expt).lower()
         def getTreePath():
             # split path variable into single paths and replaces the ~t treename
-            expt_paths = _mds.getenv(expt+'_path')
+            expt_paths = _mds.getenv('%s_path' % expt)
```
WhoBrokeTheBuild, line 314: Isn't the % format method deprecated in python3?

WhoBrokeTheBuild (changes requested):

python/MDSplus/tree.py
zack-vii: https://realpython.com/python-string-formatting/

given that teh newstyle is not supported in 2.6 but we still support 2.6 this is a legit way of doing it.

WhoBrokeTheBuild: We should drop support for 2.6, it's ancient and out of support, and then start using the new way to be more in line with python3

WhoBrokeTheBuild (approved):

PR #1993, opened 2020-04-16
Gm redpitaya

GabrieleManduchi:
Up-to date RedPitaya ADC support. Single and multiple triggers, level driven triggering, internal, external clock and DIO4 highway clock and events


AndreaRigoni (approved):
OK

PR #1997, opened 2020-04-20
Feature: Add ACQ WRPG device class

### LINKED ISSUE #1055: Support for new D-Tacq models (from Columbia - U)
Hi Josh,
    Our research group is planning to order new D-tAcq hardware for HBT-EP, and I wanted to see if some modules are compatible with input/output control from MDSplus.  Our first intent is to use these in a standalone GPU-based feedback control system that doesn't use MDSplus (based on GPU controller presently used at HBT-EP and DIII-D), but we also would like to know MDSplus compatibility for upcoming input/output needs.  Our CAMAC hardware has been fraught with problems that are getting worse, and I want to scope out a replacement.  The D-tAcq modules from a quote we received are as follows:

- ACQ2106: motherboard/FPGA/chasssis. 1U x 19”, includes 100W external PSU, ZYNQ 7030 FPGA, embedded Linux, gigabit Ethernet.
-- This sounds similar to the housing for our existing ACQ196CPCI units, so I'm guessing MDSplus would have no need to interface with this.
- ACQ424ELF-32 : 32 channel simultaneous analog input, sub 1usec capture latency, 16 bit, +/-10V differential input. Pin-out compatible with ACQ196CPCI.
-- This is the newer modular equivalent of our ACQ196CPCI units.  We have ACQ196's that work with our MDSplus right now, but the ACQ196 units are now becoming obsolete according to D-tAcq.
- AO424ELF-32 : 32 channel simultaneous analog output, 2usec latency in 32 channel mode, 1 usec in 16 channel mode. +/-10V output. Pin-out compatible with AO32CPCI (in 32 channel mode).
-- These can be used to output arbitrary waveforms, but a baseline of just TTL Triggers and Gates in MDSplus would be great.  Our initial control system will be independent of MDSplus, but if we can use duplicates of these with MDSplus, it will save us from lots of downtime that we often experience.
- We're also looking at AO424ELF-16, which is the faster (1usec latency) 16-channel alternative.
- MGT482: SFP fiber optic adapter, 6Gbps.
-- This sounds like a a module that wouldn't interact with MDSplus, and might just be needed for our GPU-based control system.

    We also presently have two AO32CPCI units that run in a feedback loop independent of MDSplus.  If the bundled system above works well, we will probably want to re-purpose the AO32CPCI units to replace the output functionality of our CAMAC J221/222 modules if possible.  It would be great if these work with MDSplus with just triggers and gates, since they could be added seamlessly to our existing machine operation.

Thank you,
Jeff Levesque
Associate Research Scientist
Columbia University Plasma Physics Laboratory
JPL2131@columbia.edu
(212) 851-5937

santorofer:
This PR contains basically one additions and 3 changes:
Addition:
- New MDSplus device class to create and load an STL table into the D-TACQ WRPG. These are in acq2106_WRPG.py.

Changes: 
- Changes to the D-TACQ's new streaming device class. The controls or knobs to the WR clocks were added, as well as the triggerig highway to use WR messaging as triggers.
- Changes to the deploy packaging, Debina and Redhat, adding the new device class.

### DIFF
--- a/pydevices/HtsDevices/acq2106_WRPG.py
+++ b/pydevices/HtsDevices/acq2106_WRPG.py
@@ -129,7 +129,8 @@ class ACQ2106_WRPG(MDSplus.Device):
         rows, cols = (len(t_times), nchan)
         state = [[0 for i in range(cols)] for j in range(rows)]
 
-        # Building the state matrix. For each channel, we traverse all the transition times to find those who are in the particular channel. 
+        # Building the state matrix. For each channel, we traverse all the transition times to find those who are 
+        # in the particular channel. 
         # If the transition time is in the channel, we copied its state into the state[i][j] element. 
         # If a transition time does not appear in that channel, we keep the previous state for, i.e. the state doesn't change.
         for j in range(nchan):
@@ -143,8 +144,10 @@ class ACQ2106_WRPG(MDSplus.Device):
                     state[i][j] = state[i-1][j]
                     
                     # chan_t_states its elements are pairs of [ttimes, state]. e.g [[0.0, 0],[1.0, 1],...]
-                    # chan_t_states[0] are all the first elements of those pairs, i.e the trans. times: e.g [[1D0], [2D0], [3D0], [4D0] ... ]
-                    # chan_t_states[1] are all the second elements of those pairs, the states: .e.g [[0],[1],...]
+                    # chan_t_states[0] are all the first elements of those pairs, i.e the trans. times: 
+                    # e.g [[1D0], [2D0], [3D0], [4D0] ... ]
+                    # chan_t_states[1] are all the second elements of those pairs, i.e. the states: 
+                    # e.g [[0],[1],...]
                     for t in range(len(chan_t_states[0])):
                         #Check if the transition time is one of the times that belongs to this channel:
                         if t_times[i] == chan_t_states[0][t][0]:
--- a/pydevices/HtsDevices/acq2106_WRPG.py
+++ b/pydevices/HtsDevices/acq2106_WRPG.py
@@ -68,7 +68,7 @@ class ACQ2106_WRPG(MDSplus.Device):
         # the digitazer, like d1.
         uut.s0.GPG_ENABLE    ='enable'
         uut.s0.GPG_TRG       ='1'    #external=1, internal=0
-        uut.s0.GPG_TRG_DX    ='d1'   #d1 when testing two WRTTs, WRTT0 or WRTT1. d0 when testing using only one WRTT.
+        uut.s0.GPG_TRG_DX    ='d1'   #d1 for WRTT1. d0 for WRTT0 or EXT.
         uut.s0.GPG_TRG_SENSE ='rising'
         uut.s0.GPG_MODE      ='ONCE'
 
--- a/pydevices/HtsDevices/acq2106_WRPG.py
+++ b/pydevices/HtsDevices/acq2106_WRPG.py
@@ -29,8 +29,6 @@ import time
 import socket
 import math
 import numpy as np
-import csv
-import copy
 
 try:
     acq400_hapi = __import__('acq400_hapi', globals(), level=1)
@@ -41,18 +39,7 @@ class ACQ2106_WRPG(MDSplus.Device):
     """
     D-Tacq ACQ2106 with ACQ423 Digitizers (up to 6)  real time streaming support.
 
-    32 Channels * number of slots
-    Minimum 2Khz Operation
-    24 bits == +-10V
-
-    3 trigger modes:
-      Automatic - starts recording on arm
-      Soft - starts recording on trigger method (reboot / configuration required to switch )
-      Hard - starts recording on hardware trigger input
-
-    Software sample decimation
-
-    Settable segment length in number of samples
+    DIO with 32 channels
 
     debugging() - is debugging enabled.  Controlled by environment variable DEBUG_DEVICES
 
@@ -74,6 +61,9 @@ class ACQ2106_WRPG(MDSplus.Device):
 
     def init(self):
         uut = acq400_hapi.Acq400(self.node.data(), monitor=False)
+        
+        #Number of channels of the DIO482
+        nchans = 32
 
         # #Setting the trigger in the GPG module
         uut.s0.GPG_ENABLE    ='enable'
@@ -83,9 +73,10 @@ class ACQ2106_WRPG(MDSplus.Device):
         uut.s0.GPG_MODE      ='ONCE'
 
         #Create the STL table from a series of transition times and states given in OUTPUT.
+        start_time = time.time()
         print("Building STL: start")
-        self.set_stl()
-        print("Building STL: end")
+        self.set_stl(nchans)
+        print("Building STL: end --- %s seconds ---" % (time.time() - start_time))
 
         #Load the STL into the WRPG hardware: GPG
         traces = False  # True: shows debugging information during loading
@@ -96,21 +87,19 @@ class ACQ2106_WRPG(MDSplus.Device):
 
 
     def load_stl_file(self,traces):
-        example_stl=self.stl_file.data()    
+        stl_table = self.stl_file.data()    
 
-        print('Path to State Table: ', example_stl)
+        print('Path to State Table: {}'.format(stl_table))
         uut = acq400_hapi.Acq400(self.node.data(), monitor=False)
         uut.s0.trace = traces
         
         print('Loading STL table into WRPG')
-        with open(example_stl, 'r') as fp:
+        with open(stl_table, 'r') as fp:
             uut.load_wrpg(fp.read(), uut.s0.trace)
 
-    def set_stl(self):
-        nchan = 32
 
-        states_hex    = []
-        states_bits   = []
+    def set_stl(self, nchan):
+
         all_t_times   = []
         all_t_times_states = []
 
@@ -120,7 +109,7 @@ class ACQ2106_WRPG(MDSplus.Device):
             # Pair of (transition time, state) for each channel:
             chan_t_states = chan_t_times.data()
 
-            # Creation of an array that contains as EVERY OTHER element all the transition times in it, adding them
+            # Creation of an array that contains, as EVERY OTHER element, all the transition times in it, appending them
             # for each channel:
             for x in np.nditer(chan_t_states):
                 all_t_times_states.append(x) #Appends arrays made of one element,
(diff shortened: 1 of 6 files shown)


zack-vii:
fixup in PR #1998

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,189 @@
+#
+# Copyright (c) 2017, Massachusetts Institute of Technology All rights reserved.
```
WhoBrokeTheBuild: Change this to Copyright 2020

santorofer: Done. Thanks. I also changed it in Sean's code.

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,179 @@
...
+        import acq400_hapi
+        uut = acq400_hapi.Acq400(self.node.data(), monitor=False)
+        
+        #Number of channels of the DIO482
+        nchans = 32
+
+        # #Setting the trigger in the GPG module
+        uut.s0.GPG_ENABLE    ='enable'
+        uut.s0.GPG_TRG       ='1'    #external=1, internal=0
+        uut.s0.GPG_TRG_DX    ='d0'
```
WhoBrokeTheBuild: Maybe an extra comment to help this make sense

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,179 @@
...
+
+    parts=[
+        {'path':':NODE',        'type':'text',                     'options':('no_write_shot',)},
+        {'path':':COMMENT',     'type':'text',                     'options':('no_write_shot',)},
+        {'path':':TRIG_TIME',   'type':'numeric',                  'options':('write_shot',)},
+        {'path':':RUNNING',     'type':'numeric',                  'options':('no_write_model',)},
+        {'path':':LOG_OUTPUT',  'type':'text',   'options':('no_write_model', 'write_once', 'write_shot',)},
+        {'path':':INIT_ACTION', 'type':'action', 'valueExpr':"Action(Dispatch('CAMAC_SERVER','INIT',50,None),Method(None,'INIT',head))",'options':('no_write_shot',)},
+        {'path':':STOP_ACTION', 'type':'action', 'valueExpr':"Action(Dispatch('CAMAC_SERVER','STORE',50,None),Method(None,'STOP',head))",      'options':('no_write_shot',)},
+        {'path':':STL_FILE',    'type':'TEXT'},
```
WhoBrokeTheBuild: Why is the STL_FILE a node? Isn't it generated and loaded by the device? We should have dt-acq modify the HAPI so that it can load STL's without files, and remove this completely, we should hopefully never be using the filesystem to store a temp file during a shot.

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,193 @@
...
+        {'path':':RUNNING',     'type':'numeric',                  'options':('no_write_model',)},
+        {'path':':LOG_OUTPUT',  'type':'text',   'options':('no_write_model', 'write_once', 'write_shot',)},
+        {'path':':INIT_ACTION', 'type':'action', 'valueExpr':"Action(Dispatch('CAMAC_SERVER','INIT',50,None),Method(None,'INIT',head))",'options':('no_write_shot',)},
+        {'path':':STOP_ACTION', 'type':'action', 'valueExpr':"Action(Dispatch('CAMAC_SERVER','STORE',50,None),Method(None,'STOP',head))",      'options':('no_write_shot',)},
+        # {'path':':STL_FILE',    'type':'TEXT'},
+    ]
+
+
+    for j in range(32):
+        parts.append({'path':':OUTPUT_%3.3d' % (j+1,), 'type':'NUMERIC', 'options':('no_write_shot',)})
```
WhoBrokeTheBuild, line 57: `(j+1,)` should be `(j+1)`

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def init(self):
+        import acq400_hapi
+        uut = acq400_hapi.Acq400(self.node.data(), monitor=False)
+        
+        #Number of channels of the DIO482
+        nchans = 32
```
WhoBrokeTheBuild: This should be stored as a class constant

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # uut.s0.SIG_SRC_TRG_0 ='WRTT0'
+        # Setting the trigger in the GPG module. These settings depends very much on what is the
+        # configuration of the experiment. For example, when using one WRTT timing highway, the we can use d0, which will be
+        # the same used by the digitazer module. Otherwise, we can choose a different one, to be a independent highway from
+        # the digitazer, like d1.
+        uut.s0.SIG_EVENT_SRC_0 = 'GPG'
+        uut.s0.GPG_ENABLE    ='enable'
+        uut.s0.GPG_TRG       ='1'    #external=1, internal=0
```
WhoBrokeTheBuild, line 73: Are these supposed to be configurable by changing nodes in the device?

pydevices/HtsDevices/acq2106_WRPG.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Converting the original units of the transtion times in seconds, to micro-seconts:
+        times_usecs = []
+        for elements in t_times:
+            times_usecs.append(int(elements * 1E6)) #in micro-seconds
+        # Building a pair between the t_times and bin states:
+        stl_list = zip(times_usecs, binrows)
+
+        # Creating the STL file with the path and file name given in self.stl_file.data()
+        fd, temp_path = mkstemp(prefix='acq2106-dio482-', suffix='.stl')
```
WhoBrokeTheBuild: We *really* should not be using a temp file for this, did you ever reach out to dtacq to get them to add a non-file version of this function to their library?

pydevices/acq400Devices/acq400_base.py
```diff
@@ -92,7 +88,51 @@ def init(self):
...
+        uut.s1.CLKDIV = "{}".format(clockdiv)
+
+        acq_sample_freq = uut.s0.SIG_CLK_S1_FREQ
+        
+        self.dprint(1, "After setting the sample rate the value in the ACQ is%r Hz", acq_sample_freq)
+
+        #The following is what the ACQ script called "/mnt/local/set_clk_WR" does to set the WR clock to 20MHz
+        uut.s0.SYS_CLK_FPMUX     = 'ZCLK'
+        uut.s0.SIG_ZCLK_SRC      = 'WR31M25'
+        uut.s0.set_si5326_bypass = 'si5326_31M25-20M.txt'
```
WhoBrokeTheBuild, line 115: What is this? A text file?

pydevices/acq400Devices/acq400_base.py
```diff
@@ -119,6 +159,17 @@ def arm(self):
         thread.start()
     ARM=arm
 
+    def trig(self, msg=''):
+        message = str(msg)
+        uut = acq400_hapi.Acq400(self.node.data())
+        self.dprint(1, "The trigger message for streamming is: %s", message)
+
+        wrtdtx = '1 --tx_id=' + message
```
WhoBrokeTheBuild: We should make sure `message` is escaped, or otherwise won't break when we add it to that command

pydevices/acq400Devices/acq400_base.py
```diff
@@ -284,18 +349,65 @@ class _ACQ400_TR_BASE(_ACQ400_BASE):
...
+
+        trstate = acq400_hapi.acq400.STATE.str(_status[0])
+        
+        print("TR state: {}".format(trstate))
+        print("TR status: {}".format(_status))
+    STATE=state
+    
+    def stop(self):
+        uut = acq400_hapi.Acq400(self.node.data())
+        print("PRE {}, POST {} and ELAPSED {}".format(uut.pre_samples(), uut.post_samples(), uut.elapsed_samples()))
```
WhoBrokeTheBuild: Change this to use the `%` formatting

pydevices/acq400Devices/acq400_base.py
```diff
@@ -311,12 +423,147 @@ def store(self):
...
+                #     print(line)
+
+        return "\n".join(stl_literal_lines)
+
+
+    def init(self):
+        # args.uuts = [ acq400_hapi.Acq2106(u, has_comms=False) for u in args.uut ]
+        uut = acq400_hapi.Acq2106(self.node.data())
+        # master = args.uuts[0]
+        self.STL.putData("/home/dt100/PROJECTS/acq400_hapi/user_apps/STL/acq2106_test10.stl")
```
WhoBrokeTheBuild, line 554: Hardcoded path?

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_WRPG.py
zack-vii: no (j+1,) is good as well % expects a tuple

pydevices/HtsDevices/acq2106_WRPG.py
zack-vii: no this is better as acq400_hapi is a third party package that we do not want to ship and maintain and may not be installed. compiling the class es must work on all default python installations + what is used by MDSplus anyway (numpy)

please check https://www.mdsplus.org/index.php?title=pydevice_dev

pydevices/HtsDevices/acq2106_WRPG.py
zack-vii: 👍

pydevices/acq400Devices/acq400_base.py
```diff
@@ -23,26 +23,28 @@
...
 
 
-import sys
 import threading
 import socket
 import time
-import inspect
-
-import numpy as np
+from queue import Queue, Empty
```
zack-vii: pythton2 incompatible.. and import full modules
https://www.mdsplus.org/index.php?title=pydevice_dev

pydevices/acq400Devices/acq400_base.py
```diff
@@ -23,26 +23,28 @@
...
+import numpy
 import MDSplus
 
-if MDSplus.version.ispy3:
-    from queue import Queue, Empty
-else:
-    from Queue import Queue, Empty
+try:
+    acq400_hapi = __import__('acq400_hapi', globals(), level=1)
```
zack-vii: dont import third party packages on module level
https://www.mdsplus.org/index.php?title=pydevice_dev

pydevices/acq400Devices/acq400_base.py
zack-vii: i think its a defined file on the dtacq device. that hopefully comes with the firmware

pydevices/acq400Devices/acq400_base.py
zack-vii: though out.. after rebasing this should trigger a fail if it is a problem some .format() syntax is ok but this one is not

pydevices/acq400Devices/acq400_base.py
zack-vii: shoule be set by node. or better file content should be content of a node.

pydevices/HtsDevices/acq2106_WRPG.py
santorofer: I think there is a way to ask the acq for the number of channels, instead of hardcoding it. I will double check that.

pydevices/HtsDevices/acq2106_WRPG.py
santorofer: Are you thinking that nodes of the device should have the values used to configure the knobs of the ACQ? That can be done, of course.

pydevices/HtsDevices/acq2106_WRPG.py
santorofer: No, I haven't yet. I was thinking of bring up this question, since I can see that they are doing something similar in their class for the _ACQ400_MR_BASE with the hardcoded STL file.

pydevices/acq400Devices/acq400_base.py
santorofer: Yes, exactly. It's in here: 
/etc/si5326.d/si5326_31M25-20M.txt

pydevices/acq400Devices/acq400_base.py
santorofer: This is something I noticed and I wanted to bring to their attention. I'm not sure why this was done. I wonder if they were just testing. We are not using Multi Rate devices yet. But for sure I need to remind them about this.

pydevices/acq400Devices/acq400_base.py
santorofer: Yes, you are correct about this.

pydevices/HtsDevices/acq2106_WRPG.py
santorofer: Yes, there is a way to ask the ACQ site 6 how many channels it has:
get.site 6 NCHAN
I should use that instead of hardcode it.

pydevices/HtsDevices/acq2106_WRPG.py
santorofer: Ok, I have added a new tree node so that we can choose in the model between d0 or d1 for the triggering signal.

pydevices/acq400Devices/acq400_base.py
santorofer: Now the message string is escaped.

pydevices/acq400Devices/acq400_base.py
santorofer: Done

pydevices/acq400Devices/acq400_base.py
santorofer: This is now done.

pydevices/HtsDevices/acq2106_WRPG.py
santorofer: Yes, I still need to ask them about this.

PR #2000, opened 2020-04-23
Fix: jtraverser2 when built with maven was missing icons

zack-vii:
fixes the maven scripts to build when maven projects changed.
fixed --release mkdir RELEASEDIR
fixed missin resources in jtraverser2


GabrieleManduchi (approved):

GabrieleManduchi:
Timo, will I merge, too?

PR #2012, opened 2020-04-30
Fix: RfxDevices for pyhton2 and python3

### LINKED ISSUE #2011: Error accessing a  python device twice
When a Python device is accessed in python, e.g 

>>>t=Tree('my_tree', -1)
>>>n=t.getNode('path of a python device')  #first time works
>>>n=t.getNode('path of a python device')  #second time fails

  File "<stdin>", line 1, in <module>
  File "/opt/anaconda3/lib/python3.7/site-packages/MDSplus/tree.py", line 800, in getNode
    return TreeNode(int(n.value),self)
  File "/opt/anaconda3/lib/python3.7/site-packages/MDSplus/tree.py", line 1092, in __new__
    return node.record.getDevice(node,head=0)
  File "/opt/anaconda3/lib/python3.7/site-packages/MDSplus/compound.py", line 265, in getDevice
    if issubclass(cls,(_tre.Device,)):
TypeError: issubclass() arg 1 must be a class

Looking into it it seems that this is due to an apperently wrong management of module.\_\_dict\_\_ in compound.py. The first time module.\_\_dict\_\_[device class name] is not defined and instruction at line 264 is correctly executed. The second time, module.\_\_dict\_\_[device class name] is defined, but it does not contain a class object, rather a module object and it causes the error

zack-vii:
major style fix and common badied replaced with goodies
* class generation
* class inheritance and field access
* submodule import
* tabs and indentation
* third party dependencies during class generation
* parts assembly


zack-vii:
@GabrieleManduchi I am not sure if "Feature :" and "Fix :" will be recognized correctly please omit the " " (space) between the keyword and the ':'.
You can do this via
```sh
git fetch
git rebase origin/alpha -i
# replace pick with r (for reword)
# remove the space
git push -f
```
now that you added pid please integrate it in MARTE2_PYTHON_PID instead of having an extra file. same or the others. gempy can be part of the superclass file

zack-vii:
@GabrieleManduchi : you kind of undid my changes though... you need to reset --hard you local branch otherwise you local version will undo all teh fixes
i will rebase and add the last two commits of yours

GabrieleManduchi:
how could I undo your changes? I pulled your version and made the fixes required to make it run, and then pushed the changes. What is wrong in this?

GabrieleManduchi:
pygam.py and pid.py must be separate from MARTE2_PYTHON.py and MARTE2_PYTHON_PID.py because pid and pygam are used also  by MARTe2 that knows nothing about MDSplus devices.

zack-vii:
hmm strange then well when you pulled thats doing a merge.. anyway i did not read this in time.. I integrated pid in the device.
if the pdi and pygem are external files then they should not be included in MDSplus as that would clash when versions change. 
In that case the FALLBACK BUILDER should suffice.

zack-vii:
I fixed the BUILDER so it should work with your external pid and pygem files. please checkout 
branch zck_marte2_pid for the changes i made to make pid and pygem safe to use ... did not test it though.. just so you get an idea what would be necessary and how it could be done.

zack-vii:
Oh and one important thing, please consider wrapping the checks in initialize in a if debug or make it an extra function check or test so you can run it during development but not every time the device is loaded.. this is for performance

GabrieleManduchi:
ok, thanks, I will checkout zck_marte2_pid and verify if it works

zack-vii:
no i meamt use this pr's branch for testing and merge but get inspiration from the zck_marte2_pid to eventually refactor pid and multiply. just make sure this pr works as expected when you have pid and multiply available.

GabrieleManduchi:
ok, I will look into it shortly

GabrieleManduchi:
Checked gm_marte2_devices, removed pygam.py and pid.py: thanks to BUILDER_FALLBACK they do not harm when working on different devices and are not reachable from PYTHONPATH. Of course for using MARTE2 devices they must be available in PYTHONPATH but this makes sense. (pid.py and pygam.py now belong to MARTe2 distribution)

GabrieleManduchi:
Now, before pushing gm-marte2.devices take note that I did the following actions

1) pulled the latest origin version
2) Solved a few merge conflicts: all the conflicts made sense
3) commited the removal of pid.py an pygam.py
As you can see this is the normal git usage, I am now ready to push it. Anything against it?

zack-vii:
there should not be any merge conflict. there should not be any merge. you cannot pull as this would try to reapply the fixed commits onto the local broken ones.
you pull approach only works if we have a clean history but i needed to rebase your branch and change commit messages etc. so git would not know what to do.
you need to 
```sh
git reset --hard origin/gm-marte2-devices
```
then test and fix if needed and commit

zack-vii:
if you added stuff after my takeover push it to a different branch first. you may cherrypick the commits after. resetting. use 
```
git reflog
git log new_branch
```
to select the hash of the commit and then 
```
git cherry-pick <hash>
```

GabrieleManduchi:
Ok, I did a reset --hard. It turns out to be equivalent to my version when merging a few conflicts. Anyway it is all ok now. There is only an updated RfxDevices/device not related to MARTE2 (and not developed by me, but they asked me to report in alpha)

zack-vii:
cool =) then you may push the changes if any and we can finish it up

GabrieleManduchi:
moving to jTraverser stuff now

zack-vii:
well i fixed the CEAN device as well. what did you add as feature? problably this is waht i found when i did diff as it reverted my fixes. in particular indentation return 0 instead of raising exceptions. print instead of dprint and parts assembly

zack-vii:
@GabrieleManduchi can you approve this PR when you think we are done here. I will build the import string fix on the updated alpha then

GabrieleManduchi (approved):

GabrieleManduchi:
Approved, but the reason why merging is blocked is unclear to me.....

GabrieleManduchi:
Never used commit signing before

zack-vii:
i activated it to see what it does. well seems like we wont be using it ;)

GabrieleManduchi:
thanks!

PR #2013, opened 2020-05-04
Feature: add LIB->ROUTINE:C that frees the returned pointer to TdiCall

### LINKED ISSUE #1865: LIB->ROUTINE:T needs a LIB->ROUTINE:C that frees the returned pointer after copying into mds descriptor
We could reduce memory leaks if we allow another TdiCall Type specifier that will trigger a CLASS_D return value. API methods like TreeGetMinimumPath, TreeGetPath, TreeGetTags suffer from this issue.

zack-vii:
* first commit adds test that should fail with memchecks.
* second commit adds feature and alters test to use :C instead of :T
fixes issue #1865


joshStillerman (approved):
one good turn deserves another.  Sorry for the delay

zack-vii:
I updated the wiki 
http://www.mdsplus.org/index.php/Documentation:Reference:TDI_C1#KIND
but i have the feeling this is not the only place that needs to be updated

PR #2018, opened 2020-05-06
Fix: consider all entries in CASE() and fixes switch and case when used in function form

zack-vii:
Allows multiple ranges in case
```tdi
switch(5) {
  case default {
    _ = $FALSE;
    break;
  }
  case(0:2, 4:6, 8:10) {
    _ = $TRUE;
    break;
  }
};_
1BU
```
as well as its function form
```tdi
switch(5, statement(\
 default(statement(_ = $FALSE, break())),\
 case((0:2, 4:6, 8:10), statement(_ = $TRUE, break())),\
));_
1BU
```
before multiple ranges in case were allowed but only the last one was tests against because they are combined in a OPC_COMMA which was evaluated first. Similar COMMA combination prevented evaluation of function form switch. This may apply to FOR UNTIL etc. as well. but will be considered in a separate PR


zack-vii:
So can I merge this at some point ;)

joshStillerman (approved):

PR #2019, opened 2020-05-06
Fix: use opcode_t as default pointer type for function desc; fixup length in mitdevices

zack-vii:


zack-vii:
opcode_t is an enum. it is the expected pointer type for mds_function_t. Exceptions are typically due to manually constructed broken descriptors (c.f. mitdevice fixes).
Using it opcode_t as default allows gdb to displays the name of the enum value (OPC_<name>) instead of a broken string or uint8 chunk of the uint16 value that represents the opcode_t. Which if a great help.

WhoBrokeTheBuild (dismissed):

Issue #2025, opened 2020-05-22
Core dump in TdiCompile when compiling <no-node>


GabrieleManduchi:
TdiCompile crashes when expression <no-node> is passed. This is an incorrect expression, so TdiCompile should report a bad status, not crash.  

The issue can be easily replicated in tdic
```
TDI>compile('<no-node>')
```
It crashes regardless a tree has been open before.

Unfortunaltely <no-node> is the output of TdiDecompile for a non existing tree node, so it can happen that this unfortunate string is then passed to TdiCompile.

zack-vii:
Please always state the version? I get
```tdic
TDI> MdsShr->MdsRelease:T()
"alpha_release_7.105.0"
TDI> compile("")
*
TDI>
```

zack-vii:
Side node "" should not be the deco of a nonexisting node.. Is it a nid?.. it should probably decompile like DTYPE_L and if it is a PATH it should just convert to DTYPE_T, no?

zack-vii:
oh i did not mean to close it

zack-vii:
can you run the two commands and post the response please
```tdi
MdsShr->MdsRelease:T()
compile('')
```

GabrieleManduchi:
Sorry, my example was not correct (GitHub canceled the special characters).
The crash occurs when you passes the string no-node within angular brackets

zack-vii:
oh i see

GabrieleManduchi:
"\<no-node\>"

GabrieleManduchi:
No way, I am not able to write in github...

zack-vii:
you need to excape the \< and \> with \\ as \\\<

GabrieleManduchi:
ok, thanks

zack-vii:
hm well this is a bad decompile.. but also it is bad that it crashes

zack-vii:
if you wrap your code in 
\`\`\` 
"\<no-node\>"
\`\`\`

it becomes 
```
"<no-node>"
```
without the need of the escaping

zack-vii:
I will see what i can do

possibly we should
1. change the decompile to and comments /\*no-node\*/ which would process as * aka $MISSING (deco string must be length 12 due to the nature of the implementation)
2. fix the compile of \<verb-verb\> which would be a syntax error?

GabrieleManduchi:
agree

zack-vii:
Hi @GabrieleManduchi 

can you test/approve the PR, please?

[...]

PR #2027, opened 2020-05-25
Fix: deco of no-node should be compilable

zack-vii:
* Decompiles nids as \\NODE::NOT.FOUND which is likely to evaluate again to TreeNNF error but can compile.
* Fixes compile crash of e.g. "<verb"
 proper fix: remove symbolic operators from label.
 failsafe: explicidly NULL rptr.
*  MULS -> only used for MOD (rename to MOD), MUL only used for '/' and '%', '*' special due use as MISSING.

fixes issue #2025


GabrieleManduchi:
Seems ok to me, shall I approve?

zack-vii:
now it should be good. when it succeeds the test, yes please review and approve, thanks.

GabrieleManduchi (approved):

PR #2032, opened 2020-05-28
Fix: PyDoMethod exception handling missing model and method

### LINKED ISSUE #2031: python do method hides exceptions from called device method
In the current version of tdi/dev_support/pyDoMethod.py  (line 34), if the called device method issues a python exception, this exception is trapped possibly printing a message. This has two incovenients:

1) The message makes reference to non existing python variables (model, method) so that a new exception is generated referring to this problem

2) In any case trapping such an exception is not a good idea for me since it hides the original problem in the device method making it very hard to debug it (indeed I had to comment the exception management in order to discover the problem in my device method)

zack-vii:
fixed #2031


GabrieleManduchi (approved):

zack-vii:
retest this please

PR #2035, opened 2020-06-02
Feature: jDispatcher great bug fix, fixed also bugs in jScope. structured classes

zack-vii:
* using manifest and java -jar
* use personalized packaging our packages should start with "mds."
* first attempt to split jscope in two layers (still inter-dependency during compilation but critical functionality of WaveDisplay available w/o jScope)
* unified java projects Makefile.am layout
* fixed jdispatcher thread managements, several typos and overly complicated/unsafe loops, typos in varnames, missing cleanup of resources.
* jDispatchMonitor allows to start/stop local jServers
* jDispatchMonitor can spawn a local jDispatcherIp if none found.
* jDispatcher resolves Dispatch and Task of a node if it is a node ref
* use execute (USING(TCL("do "//$),,$,$), nodepath, shot expt) which works with treerefs in action
* jDispatcher looks in more folders for jDispatcher_<expt>.properties
* JDispatchMonitor could provides menu to send predefined TCL commands to jDispatcherIp server.
* updated default jDispatcher.property with a general example
TODO:
* JUnit Test for the non-GUI components (low priority)


GabrieleManduchi (dismissed):

GabrieleManduchi:
Timo, note that jdispatcher and jserver are fully documented in MDSplus tutorial

GabrieleManduchi:
The reason for 1.6 was due to matlab, but maybe now it is no more necessary

GabrieleManduchi:
jServer has a main,  the startup script is correct (maybe you were confused by Server.java, that is a component of jDispatcher)

GabrieleManduchi:
not sure there is yet any user of WaveDisplay..........

zack-vii:
oh if there is a jServer that might be.. nothing showed up when i started it

GabrieleManduchi:
yes, indeed it waits for incoming actions to be executed

zack-vii:
@GabrieleManduchi : were you still using the jdispatcher at RFX? I when through the code and found several bugs. I cleaned it up pretty good. and it works for far i can tell. well i would need a proper test tree. currently has a test tree with one DEMOADC device but was running on windows so the lib was not available. it did try INIT and STORE though.

zack-vii:
Hi @GabrieleManduchi I know its a lot of changes but the seemed necessary. the jDispatcher is pretty smooth now and has great features for doing test measurements and debugging.

zack-vii:
Windows installer will be available here once the test is complete
https://www.mdsplus.org/dist/release/Windows/zck_jar_manifest/MDSplus-zck_jar_manifest-7.106-0.exe

GabrieleManduchi (approved):

GabrieleManduchi:
Good if you revise jDispatcher code, it is pretty old (20 years). Nevertheless we have always used it and we are still using it in our facility (and plan to use in the new RFX machine). I believe it is much better in functionality than the C dispatcher and server version, In our RFX application we use around 40 jServers and in these years it behaved pretty well.

GabrieleManduchi:
When you have finished it, I can try it in a dry run on our NBTF test facility. What is important is that the protocol between jdispatcher and jservers is not changed, otherwise it would be very difficult to upgrade the system in our system with tens of actor nodes.

zack-vii:
the protocol jdispatcher -> jServer should be as before. Feature wise its done i think. next would be the unit test.

PR #2036, opened 2020-06-03
Fix:update .h include for TreeNode

GabrieleManduchi:
setNciFlags and getNciFlags were already implemented in mdsobjects.c, but I did not update (i.e. generate via javah) the include file. Now the new one has been generated

### DIFF
--- a/java/mdsobjects/src/main/java/MDSplus/TreeNode.java
+++ b/java/mdsobjects/src/main/java/MDSplus/TreeNode.java
@@ -135,9 +135,9 @@ public class TreeNode extends Data
 	private static native long getNciLong(int nid, long ctx, int nciType) throws MdsException;
 	private static native java.lang.String getNciString(int nid, long ctx, int nciType) throws MdsException;
 	private static native void setNciFlag(int nid, long ctx, int flagType, boolean flag) throws MdsException;
-	//private static native void setNciFlags(int nid, long ctx, int flags) throws MdsException;
+	private static native void setNciFlags(int nid, long ctx, int flags) throws MdsException;
 	private static native boolean getNciFlag(int nid, long ctx, int flagType) throws MdsException;
-	//private static native int getNciFlags(int nid, long ctx) throws MdsException;
+	private static native int getNciFlags(int nid, long ctx) throws MdsException;
 	private static native int[]getNciNids(int nid, long ctx, int nciNumType, int nciType) throws MdsException;
 	private static native void turnOn(int nid, long ctx, boolean on) throws MdsException;
 	private static native boolean isOn(int nid, long ctx) throws MdsException;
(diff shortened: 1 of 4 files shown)


java/mdsobjects/src/main/java/MDSplus/TreeNode.java
```diff
@@ -599,7 +599,7 @@ public int getNciFlags() throws MdsException
 				flags += flag;
 		return flags;
 	}
-
+/*
```
zack-vii: HINT: We use git for versioning. you should not comment old code when there is no lesson to learn.
You need the function signature to wrap around the native entry points. the nid argument is a field of TreeNode instance.

set/getNciFlags should be 
{
resolveNid();
this.setNciFlags(nid, ctxTree.getCtx(), flags); / return this.getNciFlags(nid, ctxTree.getCtx())
}
and keep  this.getNciFlag

javamds/MDSplus_TreeNode.h
```diff
@@ -307,6 +325,14 @@ JNIEXPORT void JNICALL Java_MDSplus_TreeNode_turnOn
 JNIEXPORT jboolean JNICALL Java_MDSplus_TreeNode_isOn
   (JNIEnv *, jclass, jint, jlong);
 
+/*
```
zack-vii, line 328: Are you sure you want to implement isParentOff? since you already have a isOn, the parent version should be changed to isParentOn returning the inverse.

You need to change this where you used it... if you are using it already.

zack-vii (changes requested):

GabrieleManduchi:
Sorry...I have been too quick...

GabrieleManduchi:
Parent off is not the same of !parentNode.isOn, and indeed TreeIsOn returns either TreeON, TreeOFF and TreePARENT_OFF(i.e. the node has not turned off, but one ancestor has been turned off)

zack-vii:
i am aware of the concept of parent_state just saying if you have a isOn intuitively it pairs better with parentOn

GabrieleManduchi:
Right, probably I followed (more than 10 years ago...) more strictly the original treeshr  naming approach (TreePARENT_OFF)

java/mdsobjects/src/main/java/MDSplus/TreeNode.java
```diff
@@ -590,12 +590,17 @@ public void setWriteOnce(boolean writeOnce) throws MdsException
 		resolveNid();
 		setNciFlag(nid, ctxTree.getCtx(), NciM_WRITE_ONCE, writeOnce);
 	}
+	public boolean getNciFlag(int flagMask)
```
zack-vii: It was using flagId before if i am not mistaken, so it should now too
```
flagMask = 1<<flagId
```

zack-vii (changes requested):

java/mdsobjects/src/main/java/MDSplus/TreeNode.java
zack-vii: also this is recursive
 you meant
```
public boolean getNciFlag(int flagId)
{
    resolveNid();
    return this.getNciFlag(nid, ctxTree.getCtx(), 1<<flagId);
}

zack-vii (changes requested):

GabrieleManduchi:
No,  in this case the passed mask may contain more than one bit set (it is up to the caller to build the approprate mask)

GabrieleManduchi:
wait....I must look into it

GabrieleManduchi:
yes, you are  right

zack-vii:
@GabrieleManduchi I ll just do it real quick.

GabrieleManduchi:
Hmmm....the constants NciMxxx passed to getNciFlag already contain the bit shifted, so it is not correct to make a further shift in getNciFlags

zack-vii:
but they do call the native getNciFlag

zack-vii:
look at the current diff and you will see this it how it was before

zack-vii (dismissed):

GabrieleManduchi:
ok, seems so, anyway I will test it afterwards

zack-vii:
hmm but based on Database
11:49:25 /home/jenkins/slave/workspace/1-pull-request-test/java/jtraverser/src/main/java/Database.java:251: error: method getNciFlag in class TreeNode cannot be applied to given types;
11:49:25 		return node.getNciFlag(flagMask);
you where using it as mask.. so there is another inconsistency i will see if there are more hint on how it was implemented before

zack-vii:
looks like originally you implemented it as FlagId which is teh right way to do as the return type is boolean and i assume there is no logic to consolidate mask with more than one bit set. Database is using it wrong though, same with setFlag

zack-vii (approved):
getNciFlag is unused but pair with set/clearNciFlag so it should use the same kind of input flagId

zack-vii:
hmm well the method is unused so it does not matter.. i renamed the argument in Database to it is not missleading

zack-vii:
@GabrieleManduchi :  I approved it. Can you merge it if you agree with my modifications?

zack-vii:
can be tested with 
https://www.mdsplus.org/dist/release/Windows/gm-fix-mdsobjects/MDSplus-gm-fix-mdsobjects-7.105-4.exe

GabrieleManduchi:
Everything seems ok, thanks

PR #2039, opened 2020-06-03
Fix: some bugs in devicebeans

zack-vii:
* wrong while condition prevented components from being investigated
* broken getUsage in jtraverser2 prevented DeviceDispatcher to show actions
* wrong scope of num_actions prevented actions to be found
* jtraverser2 would return "*" on getDataExpr if no data, Interface expects Exception


GabrieleManduchi (dismissed):

zack-vii:
even though there are a few bugs. the interface works quite well. The Interface probable should get some documentation so it is clearer what is expected (Exceptions, string format, etc.)

PR #2040, opened 2020-06-04
Gm fix jscope drag

GabrieleManduchi:
Now the information passed is the full path including experiment name, change jScope Transferable accordingly.


AndreaRigoni (approved):
OK

PR #2044, opened 2020-06-05
Fix: use doAction and Data.getMdsMsg

zack-vii:
This restore old bahaviour but handles exception if GetLastError is not found.
For full funktionality we need mdsobjects doAction fixed #2045

### DIFF
--- a/java/jdispatcher/src/main/java/mds/jdispatcher/jDispatchMonitor.java
+++ b/java/jdispatcher/src/main/java/mds/jdispatcher/jDispatchMonitor.java
@@ -795,7 +795,7 @@ public class jDispatchMonitor extends JFrame implements MdsServerListener, Conne
 				{
 					final Object[] options =
 					{ "DISCONNECT", "CANCEL" };
-					final int val = JOptionPane.showOptionDialog(null,
+					final int val = JOptionPane.showOptionDialog(this,
 							"Dispatch monitor is already connected to " + mds_server.getProvider(), "Warning",
 							JOptionPane.DEFAULT_OPTION, JOptionPane.WARNING_MESSAGE, null, options, options[0]);
 					if (val == 0)
--- a/java/jdispatcher/src/main/java/mds/jdispatcher/jServer.java
+++ b/java/jdispatcher/src/main/java/mds/jdispatcher/jServer.java
@@ -533,7 +533,6 @@ public class jServer extends MdsIp
 				if (mdsTree != null)
 					mdsTree.close();
 				mdsTree = new MDSplus.Tree(tree, shot);
-				// mdsTree.open();
 				lastTree = tree;
 				lastShot = shot;
 			}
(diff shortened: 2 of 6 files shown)


zack-vii:
this doAction should work, but i would like to test it as it seemed too easy 
Well the magic happens in CTXCALLN, which is great.
where magic is turning java ctx into treectx and calling the _ variant of the method

zack-vii:
@GabrieleManduchi 
the DEBUG stuff can go away now.. before it was a tdi function GetLastError() which was not defined.. what does it do.. should it no be in ./tdi/dev_support?

zack-vii:
ready

zack-vii:
retest this lease

zack-vii:
retest this please

zack-vii:
@GabrieleManduchi Can you review this please. and on that note, can you send me a copy of you getLastError tdi function. It should probably follow the naming schema Dev* and we should eventually provide a generic version of it in ./tdi/dev_support

GabrieleManduchi:
getLastError.fun is in tdiRfxDevices, we have no chenged version in respect of that

I wil now check your branch on a simple demo for jDispatcer, jServer jDispatchMonitor. If it works I will approve the PR, but be aware that there may be problems when running on a larger system (deadlocks, etc as we experienced in the original development). However I cannot now test on a real world system (our production system). Later this summer there will be a long shutdown and in that occasion I will test thoroughly the dispatcher stuff (but let it be n alpha now anyway)

zack-vii:
Ah i see.  hm we could then change it to 
```
PUBLIC FUN DevGetLastError(IN _expt, IN _shot, IN _action_path)
{
  RETURN(IF_ERROR(TEXT(PUBLIC _last_device_error), ""));
}
```
and
```
PUBLIC FUN DevClearLastError(IN _expt, IN _shot, IN _action_path)
{
  PUBLIC _last_device_error=*;
}
```
and document it. This way users would be able to override it if they have a different scheme to investigate issues.

GabrieleManduchi:
I am missing something......how can I checkout  zack-vii:zck-jdispatch-debug??

zack-vii:
next to Squash and merge there is "view command line instructions"
```
Step 1: From your project repository, check out a new branch and test the changes.

git checkout -b zack-vii-zck-jdispatch-debug alpha
git pull https://github.com/zack-vii/mdsplus.git zck-jdispatch-debug
```

zack-vii:
I have a commit that would rename getLastError and move it to dev_support, also it adds a DevClearLastError that resets it in case the error is not handled in action so it wont print the wrong message

zack-vii:
once this is merged i can make it an new commit?

zack-vii:
or you can try it altogether already 
```sh
git pull https://github.com/zack-vii/mdsplus.git  zck-get-last-error
```
I tried it on windows with a dummy device task 
```
public _last_device_error="MYERROR";0
```
 and it works as i would expect it.
```
Server:  Mon Jun 08 11:55:07 CEST 2020, Failed \TEST::TOP:PHANTOM:ACTIONSERVER:SOFT_TRIGGER in TEST shot 1: %SS-W-SUCCESS, Success; MYERROR
```

GabrieleManduchi:
ok, I will try later today, I have an urgent task now.....

zack-vii:
Hm I will close this PR in favor of #2049 , I would like to merge that one before alpha releases.

PR #2046, opened 2020-06-05
Fix:Missing tree context in fixup nid in copydx

GabrieleManduchi:
TreePutRecord TreeFixupNid needs also tree context otherwise TreeGetPath() fails whan getting the path of a node in a different subtree. (see issue 2043).  MdsCopyDxXdOutZ has been changed to account for an additional parameter to FixupNid. Subsequent changes around derive from this

### DIFF
--- a/include/mdsshr.h
+++ b/include/mdsshr.h
@@ -72,7 +72,7 @@ extern int MdsCompareXd(const mdsdsc_t *const dsc1_ptr, const mdsdsc_t *const ds
 extern int MdsDecompress(const mdsdsc_r_t *const rec_ptr, mdsdsc_xd_t *const out_ptr);
 extern int MdsCopyDxXd(const mdsdsc_t *const in, mdsdsc_xd_t *const out);
 extern int MdsCopyDxXdZ(const mdsdsc_t *const in, mdsdsc_xd_t *const out, void **const zone,
-	int (*const fixup_nid)(), void *const fixup_nid_arg,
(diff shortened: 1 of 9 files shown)


GabrieleManduchi:
Timo, have a look also to the change I had to do in ServerBuildDispatchTable. The previous usage of fixup_nid there was wrong as it declared an int argument, while a pointer argument was passedby CopyDxXdOutZ

GabrieleManduchi:
Check also the change in TdiGetDbi. Again, the argument for fixup_nid was wrongly declared as int

zack-vii:
i already left for the weekend. will check it on sunday when i am back home.

zack-vii:
I dont think we can make that api change, those methods are EXPORTED function so they may be used in user code.  instead we would need to somhow pass the second arg. as part of the first. e.g. is teh arg is usually a descriptor it could be a DTYPE_LIST with ctx and the usual arg1. better would be to have an _fixupd_nid that takes a array of two pointers. and pass the required args in that array

zack-vii (changes requested):
We connot do an API change

zack-vii (changes requested):
Can you add a test that fails on current alpha. here in the comments. I will modify the TreeFixupNid to take the required arglist as one arg so the api can be used as is.

zack-vii:
@GabrieleManduchi : i made an alternative PR the should have the same fix/ but also for Segment and without breaking the API in #2047 . Can you check if this fixes your problem, review it, and close this one, please?

PR #2047, opened 2020-06-07
Fix: fixed tree_fixup_nid to accept dbid

zack-vii:
this should fix #2046 also for Segments and without breaking the API


zack-vii:
retest this please

GabrieleManduchi (approved):
Ok, I tested it  and it works, better than mine as the mdsCopyDx API is not changed

PR #2048, opened 2020-06-08
Fix:wrong package name in data providers

GabrieleManduchi:
the old jScope package name has been replaced by mds.jscope, otherwise Data Providers are not activated by jScope


zack-vii (approved):
ah yes i forgot about that

PR #2049, opened 2020-06-08
Fix: use DevClearLastError and DevGetLastError instead of getLastError

zack-vii:
builds on #2044 
requires rebase once #2044 is merged


zack-vii:
@GabrieleManduchi : Can i merge this if we do a full test late on anyway ? maybe just do a quick stylistic review. are you ok with generalising the GetLastError fun and move into kernel so its available even without RfxDevices

zack-vii:
https://www.mdsplus.org/dist/release/Windows/zck-get-last-error/MDSplus-zck-get-last-error-7.107-0.exe

GabrieleManduchi:
Ok, I will do anyway a quick test today

PR #2053, opened 2020-06-09
Fix: repackage jScope: fixing dataprovider

### LINKED ISSUE #2037: Whats the purpose of WaveDisplay.jar?
WaveDisplay should be independent of jScope and may become its one project in ./java/wavedisplay.
jScope could then simple add WaveDisplay.jar as DEPENDENCY.
I assume WaveDisplay is used by jTraverser to plot date provided by jTraverser.. no need for MdsDataProvider.

most of this would be modifying access private/package/protected/public or the introduction of methods e.g.:
((Signal)src).copy(Signal dst) or getter/setter methods

### LINKED ISSUE #2052: jScope broken
Timo,
   I pulled the current alpha, but jScope is in an inconsistent state. Indeed jScope only starts with default configuration, when loading ay different data providers it breaks. Several classes make reference to package mds.connection that does not exists. Moreover the distinction between packages (mds.jscope, mds.Wavedisplay) does not make sense in my view. 

If we really want to separate packages, what makes sense in my view is to separate data providers from the remaining jScope classes. But this is no high priority in my view

zack-vii:
* fixed nullpointer on disconnect
* fixed dataprovider
* made jScope selfsustained again by including mds.connection
* mdsobjects is only required for MdsDataProviderStreaming
* all DataProviders are under mds.provider


GabrieleManduchi (dismissed):

GabrieleManduchi:
Ok as soon as it is merged, I will check it again

GabrieleManduchi (dismissed):

PR #2054, opened 2020-06-09
Build: dropped javabootclass

zack-vii:
rt6 might not be required. Does anyone recall the necessity of this classes. Were they only required when we were building on each platform (due to different jdk versions)?

Is java 8 ok for device beans?
The maven artifacts are built in java 8 and can be found here
http://jenkins2.mdsplus.org/job/maven/984/artifact/maven/
they work with 7.107.x alpha binaries.

If OK i would like to switch to java 8 for all but maybe mdsobjects (to support older matlab)

### DIFF
--- a/java/Makefile.inc.am
+++ b/java/Makefile.inc.am
@@ -1,7 +1,5 @@
 # common makefile for java packages
 
-RT6JAR = $(top_srcdir)/java/rt6.jar
-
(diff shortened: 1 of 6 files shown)


zack-vii:
retest this please

GabrieleManduchi (approved):
java 8 is ok for devicebeans

PR #2055, opened 2020-06-09
Fix: cleanup DataProvider: private methods and removed unused

zack-vii:
looks like most DataProvider where copied from MdsDataProvider but only implement DataProvider which resulted in a lot of unused methods.


GabrieleManduchi:
Well, most of them are old providers no more used (e.g. TwuDataProvider for Textor no more existing). Sooner or later I will get rid of them.

GabrieleManduchi:
Ok, it works, but the current package splitting does not make much sense.
What makes sense the division in providers and jScope, because the only interface between code of the provider objects and that of jScope is the DataProvider interface.  jScope, if we really want, may be split in waveform and jscope, i.e. the code for plotting waves and that for assembilng waveforms in jscope.  I my work on it, but I have currently more urgent tasks

GabrieleManduchi:
To be more specific: mds.provider.access and mds.connection are not meaningful, they gather classes that have no shared functionality.

I see now that you have already split mds.jscope and mds.wave, which makes sense. 

Anyway, let's keep the current organization, when I find some time I will remove old classes no more used, and the organization possibly becomes clearer.

zack-vii:
the split of mds.connection as it was the only  thing that was needed for jdispatcher. moving it to mdsobjects droped its dependency of jscope entirely. the other packaging helps a lot in terms of public private package access rights. ideally you weoikd want everything as private a possible so you would know whats what. therfore it helps to understand side effects when working on a feature/bugfix.
I am glad you are ok with it.
next step would be for me to maven it.
and then to merge it with jscopew7x. which features a.o. rgba images, and we need switch or logic that controls the meaning of int64 times, ms or ns.

PR #2056, opened 2020-06-10
Fix: add mdsobjects to maven and a minor improvement for mdsplus-api

zack-vii:
* added pom for mdsobjects to be build for maven
* jscope was to follow but mindterm is not available.. maybe use jsch instead


GabrieleManduchi:
We can get rid of JetDataProvider and consequently of mindterm. The refer to the old RDA interface provided more than 15 years ago at Jet, but now Jet exports data directly in MDSplus format.
So feel free to remove them.

PR #2057, opened 2020-06-10
Feature: jScope to use jsch for ssh and added to maven

zack-vii:
@GabrieleManduchi 
Could you test the ssh functionality, i.e. the ssh tunnel for MdsDataProvider. I switched to jsch as it is available on maven. what it does is as before.. you forward a remote port to local. but it can also interpret .ssh/config, i.e. i follows ProxyJump definitions.

in particular: Please double check it teh notations are correct:
```
argument = server:11111
tunnelport = 22222
```
will create a tunnel according to 
```
ssh -L 22222:127.0.0.1:11111 server
```
So tunnelport is expected to be a int .. is that correct?


zack-vii:
full jar files for testing can be found here
http://jenkins2.mdsplus.org/job/maven/984/artifact/maven/
they work with 7.107.x alpha binaries

GabrieleManduchi:
Good for the tunneling, even if we use it very seldomly

GabrieleManduchi (approved):

PR #2058, opened 2020-06-11
Fix: remove jet and fixup of interfaces

zack-vii:
* I removed the JetDataprovider and replaced it with the former JetMdsDataProvider
* I cleaned up the interfaces


GabrieleManduchi:
thanks for doing a job I always wanted to do.....

zack-vii:
Check this out =)
https://search.maven.org/search?q=g:org.mdsplus

GabrieleManduchi (approved):

PR #2059, opened 2020-06-11
Fix: last cleanup before maven release

zack-vii:
As you suggested, I removed the DataAccess stuff and added a mechanism thats should populate the Provider list dynamically


GabrieleManduchi (dismissed):

zack-vii:
done :)
Can you merge it, please? so i wont be tempted to add more ;)

PR #2061, opened 2020-06-13
Fix: prevent LOCK_FAILURE on windows

zack-vii:
Real issue is that some method lock and  call TreePutNci which locks as well.
the lock is not reentrant and does not keep track of number of locks.
```
LOCK = locked
protected access
{
  LOCK = locked
  protected access
  UNLOCK = unlocked
}
unprotected access - assumed to be protected
UNLOCK = fail
```


zack-vii:
enabling debug revealed that there is at least one more nested io_lock. 

edit: fixed using the locked tracker

zack-vii:
new implementation used locked counter to keep track of nci lock. will only lock first time and unlock last time.

joshStillerman (approved):

PR #2065, opened 2020-06-17
Fix: fixup filters hide hidden files and __pycache__; print shows in …

zack-vii:
…mdsip, err does not

### DIFF
--- a/pydevices/HtsDevices/cryocon18i.py
+++ b/pydevices/HtsDevices/cryocon18i.py
@@ -242,8 +242,8 @@ class CRYOCON18I_SHOT(CRYOCON18I):
 
         print('Writing data into shot node')
         for i in self.inputs:
-            trend_temp     = trend_dev.__getattr__('input_%c'.format(c))
-            trend_resis    = trend_dev.__getattr__('input_%c_resistence'.format(c))
+            trend_temp     = trend_dev.__getattr__('input_%c' % (i,))
+            trend_resis    = trend_dev.__getattr__('input_%c_resistence' % (i,))
 
             times    = trend_temp.dim_of().data()
             temps    = trend_temp.data()
--- a/pydevices/HtsDevices/cryocon18i.py
+++ b/pydevices/HtsDevices/cryocon18i.py
@@ -24,9 +24,7 @@
 #
 import MDSplus
 import threading
-import copy
 import socket
-import string
 import time
 import datetime
 import numpy as np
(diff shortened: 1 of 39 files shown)


zack-vii:
This the last time I attempt to fix mitdevices. it just systax correction mostly tabs, print, and Exception, one wrong import

Now that the findPydevices if fixed we see how broken mitdevices are

zack-vii:
can you review this so we can rebase #2066 onto this fix so we can identify potential issues with the changes

pydevices/HtsDevices/cryocon18i.py
```diff
@@ -241,30 +236,29 @@ def store(self):
...
             temps    = trend_temp.data()
             resists  = trend_resis.data()
 
             start_time = times[0]
             for j in range(len(times)):
                 times[j] -= start_time
-                times[j] = float(times[j]) / 1000.    
+                times[j] = float(times[j]) / 1000.
+
+            shot_temp     = self.__getattr__('input_%c' % (i,))
```
joshStillerman, line 257: Why use format above but not here ?  I don't understand the  tradeoffs between various format syntaxes

pydevices/HtsDevices/cryocon24c.py
```diff
@@ -29,31 +27,27 @@ class CRYOCON24C(MDSplus.Device):
...
+    for i in inputs.upper():
+        parts.extend([
+            {'path': ':INPUT_%c' % (i,), 'type': 'signal', 'options': ('no_write_model', 'write_once',)},
+            {'path': ':INPUT_%c:SERIAL_NO' % (i,), 'type': 'TEXT', 'options': ('no_write_shot')},
+            {'path': ':INPUT_%c:CALIBRATION' % (i,), 'type': 'TEXT', 'options': ('no_write_model', 'write_once',)},
+            {'path': ':INPUT_%c:RESISTENCE' % (i,), 'type': 'signal', 'options': ('no_write_model', 'write_once',)},
+        ])
+    for i in loops:
+        parts.extend([
+            {'path': ':LOOP_%d' % (i,), 'type': 'signal', 'options': ('no_write_model', 'write_once',)},
```
joshStillerman, line 48: Same question as above

pydevices/MitDevices/dt132.py
```diff
@@ -40,7 +40,7 @@ class DT132(Device):
     Init(arg) - initialize the DTAO32 device
                 write setup parameters and waveforms to the device
     Store(arg) - store the data acquired by the device
-    Help(arg) - Print this message
+    Help(arg) - print(this message
```
joshStillerman: extra '('

pydevices/acq400Devices/acq400_base.py
```diff
@@ -81,18 +81,18 @@ def init(self):
 
         # The default case is to use the trigger set by sync_role.
         if self.trig_mode.data() == 'role_default':
-            uut.s0.sync_role = "{} {}".format(self.role.data(), self.freq.data())
+            uut.s0.sync_role = "%s %s" % (self.role.data(), self.freq.data())
```
joshStillerman, line 84: same questions about .format

pydevices/HtsDevices/cryocon18i.py
```diff
@@ -82,12 +77,12 @@ def debugging(self):
 
 class CRYOCON18I_TREND(CRYOCON18I):
```
santorofer, line 79: Timo, I think we should leave this section unchanged. I remember that if we didn't do a shallow copy of the CRYOCON18I.parts it wouldn't work correctly. What will be the main reason for this purposed change?
Thanks.

pydevices/HtsDevices/cryocon24c.py
```diff
@@ -303,15 +301,15 @@ def stop(self):
     STOP = stop
 
 class CRYOCON24C_SHOT(CRYOCON24C):
```
santorofer, line 303: The same comment as in CRYOCON18I. I think we should leave the shallow copy in place.
Cheers.

pydevices/acq400Devices/acq400_base.py
zack-vii: .format does not work on rhel6.. i.e. old python2.. % is the most compatible variant

pydevices/HtsDevices/cryocon18i.py
zack-vii: the + [ does the copy for you

pydevices/HtsDevices/cryocon24c.py
zack-vii: same here list + list makes a copy

pydevices/HtsDevices/cryocon18i.py
zack-vii: never use .format

pydevices/HtsDevices/cryocon24c.py
zack-vii: never use .format but use % for compatibility

pydevices/HtsDevices/cryocon18i.py
zack-vii: real issue was the c instead of i

pydevices/HtsDevices/cryocon18i.py
zack-vii: extra use of copy package .. unnecessary. list(parts would accomplish the same) and in this case using list + list is better for allocation

pydevices/HtsDevices/cryocon24c.py
zack-vii: and you are right.. you have to make a copy if you intend to append or extend as they reuse the object and would add to the base_parts as well

santorofer: Ok, I understand. Then it should be fine. Thank you.

PR #2070, opened 2020-06-22
Fix: added phases to phase table

zack-vii:
provide a more versatile phase table as default. 

leave room for fill-ins by incrementing defualt phase id by 10 (commodore style)

added pre, post, trigger, store2 as aliases.

### DIFF
--- a/tdi/dispatch/phases/phase_table.fun
+++ b/tdi/dispatch/phases/phase_table.fun
@@ -1,19 +1,19 @@
-/*	 PHASE_TABLE.FUN
-; PURPOSE:	Site customizable phase table
-;               for use with MIT Dispatch/Serve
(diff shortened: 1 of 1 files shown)


zack-vii:
looks like this triggered an issue in TdiSort.c
http://jenkins2.mdsplus.org/job/fc27/5996/artifact/tests/64-san-thread/mdsobjects/cpp/testing/MdsExpressionCompileTest.log/*view*/

this issue is addressed in #2072

zack-vii:
anything against this PR?

Idea was to use more general names for phases.. this way we can make more universal device defaults settings.

GabrieleManduchi:
from my side nothing against, as we use jDispatcher.......

joshStillerman (approved):

PR #2073, opened 2020-06-25
Fix:Wrong shell files

GabrieleManduchi:
Shell files CompileTree and DecompileTree did not make correct reference to java classes (now in mds.jtraverser package)

On many linux systems xinetd /etc/rc.d/init.d/xinetd  does not exist anymore and service command must be called instead.

post_install_script failed if symbolic links already /etc/profile.d/mdsplus.csh and /etc/profile.d/mdsplus.sh already present


AndreaRigoni (approved):
Not all distros provide the command "service" so that would be a further try to fire the service up.
I approve it though.

PR #2074, opened 2020-06-26
Gm fix new jtraverser

GabrieleManduchi:
1) An unmanaged exception was generated when open tree failed, now an error message is displayed

2) DeviceChoice returned incorrect text field (missing quotes)


zack-vii (dismissed):
i squashed the two commits. turns out the second commit reverts most of the first one.

zack-vii (approved):
i sqashed the two commit. turns out thge second commit reverts most of the first one.

GabrieleManduchi:
thanks Timo

PR #2083, opened 2020-07-20
Fix: System.arraycopy argument order in TdiShr.java

wuhao-thu:
Fix a bug which could cause a NullPointerException when calling tdiCompile because of the misuse of arraycopy, where it should copy the elements of "args1" to "args", instead of the other way round.


zack-vii (approved):
Ah yes thanks.. well in c its usually the other way round. hence my confusion. hm it might be an issue elsewhere too.

zack-vii:
I quickly checked the occurrences of arraycopy in mdsplus-api and the rest seemed correct, puh.

zack-vii:
ok to test.

PR #2090, opened 2020-07-30
Fix: use gperf to generate TdiHash

### LINKED ISSUE #2089: use gperf to builf hash function for tdi
https://www.gnu.org/software/gperf/

zack-vii:
May make tdi even faster as the hashfun is optimized for set.  e.g. not all chars must be considered to get unique hash values. no more filling of the hash table at runtime. 

addresses #2089


zack-vii:
anyone wants to drop a comment on this one?

joshStillerman (approved):

PR #2096, opened 2020-08-17
Fix: move imports in tdi/dev_support/*.py, so they may survive module…

zack-vii:
… reloads and co.

not trivial with original layout when it is loaded from libpython withing python. MDSplus objects unloaded but reload was not triggered as import was outside of function definition. this is due to the way we import and cache pythonic tdi funs.


zack-vii:
anyone,

PR #2097, opened 2020-08-19
Gm java streams

GabrieleManduchi:
Let jScope optimize a larger class of expressions (including nid references to nodes containing in turn other expressions including segmented nodes)


AndreaRigoni (approved):
OK !

PR #2099, opened 2020-08-20
Fix: device setup

### LINKED ISSUE #2091: Device setup issues
When trying to bring up the form for a DIO2 module I find:
jTraverser 
   - chokes with internal java error and core dumps
jTraverser2 
  - sometimes works fine, and sometimes only displays 'reset' and 'Cancel' buttons.
  - always prints out a ton of messages about not being able to read from empty nodes.
traverser
  - complains: 
```
Error: /usr/local/mdsplus/lib/libJavaMds.so: undefined symbol: deviceSetup
```

MDSplus version:
```
jas@tokzonemit:~$ mdstcl
TCL> sho version


MDSplus version: 7.109.6
----------------------
  Release:  alpha_release_7.109.6
  Browse:   https://github.com/MDSplus/mdsplus/tree/alpha_release_7.109.6
  Download: https://github.com/MDSplus/mdsplus/archive/alpha_release_7.109.6.tar.gz


TCL> 
```

zack-vii:
to call it from within MDSPLUS_DIR/java/classes
```sh
java -cp "jTraverser2.jar;Devicebeans.jar;jDevices.jar;jScope.jar" mds.devices.Device expt shot path
```

### DIFF
--- a/deploy/packaging/debian/java_bin.amd64
+++ b/deploy/packaging/debian/java_bin.amd64
@@ -6,4 +6,5 @@
 ./usr/local/mdsplus/bin/jServer
 ./usr/local/mdsplus/bin/jTraverser
 ./usr/local/mdsplus/bin/jTraverser2
+./usr/local/mdsplus/bin/setupDevice
 ./usr/local/mdsplus/lib/libJavaMds.so
--- a/deploy/packaging/debian/java_bin.armhf
+++ b/deploy/packaging/debian/java_bin.armhf
@@ -6,4 +6,5 @@
 ./usr/local/mdsplus/bin/jServer
 ./usr/local/mdsplus/bin/jTraverser
 ./usr/local/mdsplus/bin/jTraverser2
+./usr/local/mdsplus/bin/setupDevice
 ./usr/local/mdsplus/lib/libJavaMds.so
--- a/deploy/packaging/debian/java_bin.i386
+++ b/deploy/packaging/debian/java_bin.i386
@@ -6,4 +6,5 @@
 ./usr/local/mdsplus/bin/jServer
 ./usr/local/mdsplus/bin/jTraverser
 ./usr/local/mdsplus/bin/jTraverser2
+./usr/local/mdsplus/bin/setupDevice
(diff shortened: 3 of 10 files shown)


santorofer:
Timo, I was trying to test it, to see what I get if I run your java command. So I cloned your repo and checkout your  "zck-device-setup".

When I build it, using "bootstrap" and then configure, I'm not getting in the end the directory structure that you are expecting. 

For example, I don't have "java/classes", and also I don't see "java/jtraverser2/src/main/java/mds/devices/Device.java". What I get is "java/jtraverser2/classes/mds/devices"

Am I building it wrong?

zack-vii:
hm well i nad is a short version but in the end in an installed fs there should be a java/classes folder in $MDSPLUS_DIR where you find all .jar files.
Device.java is here
https://github.com/MDSplus/mdsplus/blob/alpha/java/jtraverser2/src/main/java/mds/devices/Device.java
which will be compiled and packed into jtraverser2.jar. tge other jars are requred to open the setup,

as an easy way use the jtraverser2 bin script and replave the entrypoint with mds.devices.Device . that should do it.

zack-vii:
i added the executable scripts setupDevice.template/bat in java/jtraverser2
e.g. \TEST::TOP:DEVICE is DIO2 device
```sh
setupDevice test -1 DEVICE
```

zack-vii:
retest this please

failed because /mdsplus was not mounted on all slaves

zack-vii:
TODO: tdi calls should use something like:
```tdi
execute("spawn('setupDevice $ $ $')", $EXPT, $SHOT, getnci(_nid, "FULLPATH"))
```
python could use os.system() or subprocess.POpen or similar

edit: on that note if the tdi functions are converted to python scripts they could use a shared python method to do the system call

santorofer:
Hi Timo,
Just tested your setupDevice executable with one of our test trees that had a DIO2.
It worked well, it seems.
Cheers!

zack-vii:
great thanks for testing. how do we go from here. i can implement the puthon api. should it be a separate pr? or should we merge this first

GabrieleManduchi (approved):

PR #2101, opened 2020-08-26
Fix: jtraverser crash

GabrieleManduchi:
Handle empty nodes in DeviceChoice avoiding jTraverser crash

### DIFF
--- a/java/devicebeans/src/main/java/DeviceChoice.java
+++ b/java/devicebeans/src/main/java/DeviceChoice.java
@@ -122,13 +122,18 @@ public class DeviceChoice extends DeviceComponent
 		else
 			try
 			{
-				final String data_value = subtree.getString(data);
-				for (int i = 0; i < choiceItems.length; i++)
+				if(data == null)
+					comboB.setSelectedIndex(-1);
+				else
 				{
-					if (choiceItems[i].equals(data_value))
+					final String data_value = subtree.getString(data);
+					for (int i = 0; i < choiceItems.length; i++)
 					{
-						comboB.setSelectedIndex(i);
-						break;
+						if (choiceItems[i].equals(data_value))
+						{
+							comboB.setSelectedIndex(i);
+							break;
+						}
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


zack-vii:
```
$ git checkout gm-fix-jtraverser
Branch gm-fix-jtraverser set up to track remote branch gm-fix-jtraverser from upstream.
Switched to a new branch 'gm-fix-jtraverser'
$ git rebase upstream/alpha -i
Successfully rebased and updated refs/heads/gm-fix-jtraverser.
$ git push -f
```

zack-vii (dismissed):
definitely a bug fixed - should not have cause a crash though.. eventually you could look into why passing null to  subtree.getString(null) causes a crash and handle it properly on the lowest level possible.

zack-vii (changes requested):
1.) my concern is that you fix merely glosses over the real issue that subtree.getString() crashes instead of throwing an error when passed null.
2.) your exception handling to chose index 0 selects the first one in the list which may mislead the user that everything is in check when in reality no data is recorded. the field should be empty. is it index -1?

GabrieleManduchi:
A better management is now provided:
- An exception is issued by Database.getString() and Database.getStringArray() rather than crashing
- DeviceChoice displays no choice if the item is missing

zack-vii (changes requested):
should the handling not be on a lower level e.g. in tdicompile or if possible even lower in the jni (c) code.
point is java is not supposed to crash so jni should honor that and do best efforts to prevent it using exception handling

zack-vii (changes requested):
now that i see teh code .. you could actually condense the 4 methods into two by using Data... args instead of Data args[].. this would automatically render args to be new Data[0] if no args provided

also turns out tdiExecute and tdiCompile cannot throw Exception as it changes the signature
maybe make them return $Missing instead

GabrieleManduchi:
retest this please

zack-vii:
retest this please

zack-vii (approved):

PR #2112, opened 2020-09-04
Fix: Improve HW filtering (this also fixes a bad merge)

santorofer:
Just a small changes to the HW filtering algorithm. Also, this PR will fix a bad merge/build of the version of the acq435st.py that is currently in "Release:  alpha_release_7.111.1"

1- there a check has been added to be sure that if the acq435 returns a HW filter = 0, that this is set to 1 instead, which. means: no hardware decimation.
2- the bad merge code that was in the INIT() is now the correct one, where the HW_filter is set.
3- The hardware filter tree node  now is not a power of 2. Instead it is the actual hardware sample decimation. It is in the range of [1,32] where 1 (and 0) means "decimation disabled")
4- We read the value set in the INIT() function directly from the box when using it in the MDSworker.

### DIFF
--- a/pydevices/HtsDevices/acq435st.py
+++ b/pydevices/HtsDevices/acq435st.py
@@ -147,7 +147,7 @@ class ACQ435ST(MDSplus.Device):
 
             trig = self.dev.trigger.data()
             
-            # Retrive the actual value of NACC (samples) set in the ACQ box
+            # Retrive the actual value of NACC (samples) already set in the ACQ box
             nacc_str   = uut.s1.get_knob('nacc')
             if nacc_str == '0,0,0':
                 nacc_sample = 1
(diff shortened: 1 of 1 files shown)


pydevices/HtsDevices/acq435st.py
```diff
@@ -147,15 +147,19 @@ def lcma(arr):
...
+                nacc_tuple  = ast.literal_eval(nacc_str)
+                nacc_sample = nacc_tuple[0]
+                
             if self.dev.debug:
-                print("The ACQ NACC value was set to {}".format(nacc))
+                print("The ACQ NACC sample value was set to {}".format(nacc_sample))
 
-            # nacc values are always between 1 and 16, set in the ACQ box by the INIT() function , therefore:
-            dt = 1./self.dev.freq.data() * nacc
+            # nacc_sample values are always between 1 and 16 (ie 2^0 to 2^4, see nacc in INIT), set in the ACQ box by the INIT() function , therefore:
```
joshStillerman: 1 and 32.  2^0 .. 2^5

joshStillerman (dismissed):

WhoBrokeTheBuild (dismissed):

WhoBrokeTheBuild (approved):

PR #2117, opened 2020-09-15
Fix:let no args tdiExecute and tdiCompile in Tree.java

GabrieleManduchi:
Variable args do not handle the case no args

### DIFF
--- a/java/mdsobjects/src/main/java/MDSplus/Tree.java
+++ b/java/mdsobjects/src/main/java/MDSplus/Tree.java
@@ -531,11 +531,28 @@ public class Tree
 		return retData;
 	}
 
+	public Data tdiCompile(java.lang.String expr)
+	{
+		if (expr == null || expr.isEmpty())
+			return new Data();
+		final Data retData = compile(ctx, expr, new Data[0]);
+		retData.setCtxTree(this);
+		return retData;
+	}
+
 	public Data tdiExecute(java.lang.String expr, Data... args)
 	{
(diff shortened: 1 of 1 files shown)


zack-vii (commented):
these functions should not be required as they are covered by the Data... data methods.
calling tree.tdiExceute("$SHOT") should leave data be an empty array of Data just like new Data[0]

GabrieleManduchi:
Unfortunately they are not covered by the Data... arguments. I discovered this just calling tdiCompile with no args.

GabrieleManduchi:
retest this please

zack-vii:
could you check in debug. i use it that way a lot in mdsplus-api. and otherwise this question on SO does not make sense

https://stackoverflow.com/questions/12751746/requiring-at-least-one-element-in-java-variable-argument-list

zack-vii:
calling tdiCompile with no args is an issue as you need at least the String expr.

zack-vii:
Also the unit ttest covers it in
java/mdsobjects/src/test/java/MDSplus/MdsTreeTest.java
DataTest line 36-39

zack-vii (changes requested):
I am not convinced this code duplication is necessary.

GabrieleManduchi:
Really strange, I repeat the tests on my side.........

GabrieleManduchi:
You are right, I discard the PR  I don't know why I got those compilation errors ....

PR #2126, opened 2020-09-30
Fix: IDL bug in evaluate causing dimension truncation in mdsvalue

### LINKED ISSUE #2125: mdsvalue truncates data with dimensions of size 1
When uploading data to mdsplus using mdsput or mdsvalue using a VM in idl (without a license), mdsvalue removes degenerate dimensions (dimensions of size 1) before uploading them to mdsplus. The result is that the dimensions of the data get truncated in idl and the data is uploaded with the degenerate dimensions removed. The bug was narrowed down to be in idl/mdsvalue starting at like 67. When the statements below are not run in a VM (with full idl license) the idl "execute" function is used and the data retains its dimensions and is uploaded correctly. However, when run in a VM, the "evaluate" function is used instead and the data has its degenerates removed by this function before being uploaded by mdsplus.
The following code below can be used to reproduce the bug in idl:

Code:
 psin = [1,2,3]
    psin = REFORM(psin,3,1,/OVERWRITE)
    help,psin
    print,size(psin)
    result = MdsValue('TreePut($,$,$)','\PSIN','BUILD_SIGNAL(BUILD_WITH_UNITS($,\PSIN:UNITS),*,*)',psin,/quiet,status=stat)
    help,psin
    print,size(psin)
Output when using VM:
PSIN            INT       = Array[3, 1]
           2           3           1           2           3

PSIN            INT       = Array[3]
           1           3           2           3
Note: The data uploaded to mdsplus server also has the same truncated dimensions above.

Output when using full idl license:
PSIN            INT       = Array[3, 1]
           2           3           1           2           3

PSIN            INT       = Array[3, 1]
           2           3           1           2           3

Opened branch and pull request with proposed solution that is confirmed to work:
link: https://github.com/MDSplus/mdsplus/pull/2126

orozda:
These changes were made to fix bug brought up in issue 2125, more info in issue.
Link: https://github.com/MDSplus/mdsplus/issues/2125

in file idl/mdsvalue.pro, starting at line 27, mdsvalue uses either idl function execute or custom function "evaluate" depending on whether it is being called from VM. The issue was narrowed down to be in "evaluate" function. The logic added would remove the need to call execute or evaluate, since they are only used on a string and must fit into one of the case statements.

### DIFF
--- a/idl/mdsvalue.pro
+++ b/idl/mdsvalue.pro
@@ -64,11 +64,27 @@ function MdsValue,expression,arg1,arg2,arg3,arg4,arg5,arg6,arg7,arg8,arg9,arg10,
     if not status then return,0
     n = n_params()
     Mds$SendArg,sock,n,0,expression
-    if (vm_being_used) then begin
-      for i=1,n-1 do x = evaluate('Mds$SendArg,sock,n,i,arg'+strtrim(i,2))
-    endif else begin
-      for i=1,n-1 do x = execute('Mds$SendArg,sock,n,i,arg'+strtrim(i,2))
-    endelse
+    for i=1,n-1 do begin 
+          case (i) of ;This case statement replaces the logic to use idl function execute or evaluate
+          ; replacing this logic with a case statement, removes the dependence on an idl license and evaluate (removing bugs in evaluate)
+              1:Mds$SendArg,sock,n,i,arg1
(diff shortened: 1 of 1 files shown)


zack-vii (approved):
I assume you tested it.. i like what you promised in the comment

joshStillerman:
retest this please

PR #2127, opened 2020-10-06
fix: python wrong comparison

AndreaRigoni:
You should use **== and !=** to compare variables **values** and use the Python **is** and **is not** operators only when you want to **check whether two variables point to the same memory address**.

taken from: [https://realpython.com/python-is-identity-vs-equality/](https://realpython.com/python-is-identity-vs-equality/)


zack-vii (approved):
Strings in python are reused so 'is' should work as well and is faster. but if that has changed or is not valid for all versions this should work too.

PR #2134, opened 2020-10-16
Build: there are no globus 32bit RHEL8 rpms

joshStillerman:
Remove the -with-gsi 32 bit tests for RHEL8 since there are no
32 bit globus RPMS


zack-vii (approved):
should do the trick

PR #2135, opened 2020-10-19
Acq400s minimum sample rate requirement

santorofer:
A simple error check has been added to both 423ELF and 435ELF. The minimum recommended value from D-Tacq to execute this devices is 10KHz.

From email conversations with Peter Milne, from D-Tacq:

"For lower rate output, the ACQ435 logic does include a simple
accumulate and decimate filter:

http://www.d-tacq.com/resources/d-tacq-4G-acq4xx-UserGuide-r33.pdf
13.1

set.site 1 NACC 8
decimate by 8, 10kHz becomes 1.25kHz.

It's best to keep to powers of 2, then voltage scaling is preserved.
If that is too much of a restriction, ACQ42x does have a filter that preserves scaling for any value of NACC.

The NACC filter means that the ADC logic can run at acceptable speeds, but since it does reduce DMA data rate, if your plan is to run at very low rates, then it's a good idea to reduce the DMA buffer length. It's best to run with > 10 DMA buffers/second, both for responsiveness, and to avoid DMA timeouts.
To set buffer len: 4GUG#28.2.7"

### DIFF
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -258,13 +258,14 @@ class _ACQ2106_423ST(MDSplus.Device):
         except:
             pass
         for card in range(self.sites):
-            coeffs =  map(float, slots[card].AI_CAL_ESLO.split(" ")[3:] )
+            coeffs  =  map(float, slots[card].AI_CAL_ESLO.split(" ")[3:] )
             offsets =  map(float, uut.s1.AI_CAL_EOFF.split(" ")[3:] )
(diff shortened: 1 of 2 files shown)


joshStillerman (commented):
I think you should raise 
`DevBAD_PARAMETER`
like:
`raise DevBAD_PARAMETER("PiCam - Could not find camera %d"%serial_no)`

WhoBrokeTheBuild (dismissed):

WhoBrokeTheBuild (approved):

PR #2138, opened 2020-10-21
423ELF full use of sync_role() function

santorofer:
The code now uses solely the sync_role() functions to set clocks and triggers.


joshStillerman (approved):
At some point we should support non 'master' modes.  

I thought we had a specific error DevUNSUPPORTED, but I can not find it.  At some point we should, or at least have DevCONFIGURATION_ERROR or something so it can be checked for.

Regardless,
I approve as is.
-Josh

PR #2145, opened 2020-10-29
Fix: Treeshr can not open thick to very old server

joshStillerman:
The current stable fails to open a tree using thick client when
the server is sufficently old.

In particular a server from
```
4352e8d4978c4d2f375b2d0094078295e826bbbb
```
Replacing RemoteAccess.c with the current alpha version appear to
fix this.

### DIFF
--- a/treeshr/RemoteAccess.c
+++ b/treeshr/RemoteAccess.c
@@ -53,7 +53,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #ifdef DEBUG
 # define DBG(...) fprintf(stderr,__VA_ARGS__)
 #else
-# define DBG(...) do{}while(0)
+# define DBG(...) {}
 #endif
 
 static inline char *replaceBackslashes(char *filename) {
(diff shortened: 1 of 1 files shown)


joshStillerman:
retest this please

joshStillerman:
retest this please - I increased the timeout on the windoz job

joshStillerman:
retest this please - I have disabled some of the slower builders in an attempt to get it not to timeout

joshStillerman:
This was timing out because the windoz test suite is hanging.  After poking around in the running docker I can not see why.  The only change in the PR was to replace RemoteAccess.c. 

Can anyone help with this ?

I think the next step is to only replace the one routine that was checking for the old version and reporting UNSUPPORTED instead of the whole file.

Thoughts ?

zack-vii:
windows sometimes times out.. if it happens more frequently maybe the RemoteAccess.c relies on other changes.?

zack-vii (commented):
This does only change the indentation, no? this should not have any effect.

joshStillerman:
retest this please - I think this time it did run out of time vs getting stuck

joshStillerman:
now I am confused.... it is true that: https://github.com/MDSplus/mdsplus/pull/2145/files/8101bb84cc69e148ed3275f723ee37bbd4d9eece 
is indentation only - though that was not my intent.  The link above your comment: 
https://github.com/MDSplus/mdsplus/pull/2145/commits/8101bb84cc69e148ed3275f723ee37bbd4d9eece
shows more significant change.

Basically what should be there now is the stable version of RemoteAccess.c with the one routine that was returning UNSUP replaced with the alpha.

zack-vii (approved):
If it works it works... well work for alpha anyway so should be good enough

zack-vii:
I should add the Ubuntu18 fix as well

joshStillerman:
can someone  approve this PR?

joshStillerman:
or does 'merged' above mean that it is already there?

PR #2146, opened 2020-10-30
Build: fixed Ubuntu18 reprepro by adding explicit export

zack-vii:
should fix the reprepro issue on ubuntu18 .. newer version refuse to export implicitly and return a non-zero errorcode.. causing publish to fail.

copied from alpha -  where it seems to work

@joshStillerman : if you could merge this before the next build.. otherwise the ubuntu18 repo needs manual fix again.


zack-vii:
ok looks like alpha may fix stable since its cleaning it up when it runs but we should still merge this to be safe

joshStillerman (approved):

PR #2150, opened 2020-11-02
Build: remove dependencies on /usr/bin/python

joshStillerman:
RPM seems to search the code and decide we need /usr/bin/python which
does not exist on rhel8

### DIFF
--- a/mdstcpip/mdsip-client-http
+++ b/mdstcpip/mdsip-client-http
@@ -1,4 +1,4 @@
-#!/usr/bin/python -u
+#!/usr/bin/env python -u
 import struct
 import os
 import sys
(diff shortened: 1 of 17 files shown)


mdstcpip/mdsip-client-http
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python -u
+#!/usr/bin/env python -u
```
WhoBrokeTheBuild, line 1: So why did you change the first one from `/usr/bin/env python` to `/usr/bin/python2` but the rest of them you changed to have `/usr/bin/env`?

WhoBrokeTheBuild (approved):
Approved, with a question

PR #2155, opened 2020-11-09
Jas alpha debian requirements

joshStillerman:
builds failing because the tests for java require python and numpy


zack-vii (approved):
look clean

PR #2159, opened 2020-11-12
Gm fix jtraverser

GabrieleManduchi:
A new device+DW setup (EL_PROBES_SCG)
A fix for jTraverser node rename
A fix to avoid OutOfBounds Exception in jScope when giving wrong dimensions
A FUN for time conversion improved


AndreaRigoni (approved):
Ok

PR #2160, opened 2020-11-15
Fix: thread-safe TreeMakeSegments and co

zack-vii:
This should fix the issue #2156 .. added test that was fixed by this.. so looks promissing.


GabrieleManduchi (approved):
Ok, this makes sense. We will later remove the mutex we used in our application code for the same problem and see what happens.

PR #2163, opened 2020-11-16
Fix: use open file descriptor locks which are thread safe

### LINKED ISSUE #2156: Race condition in TreeMakeSegment ?
In rare cases data written with TreeMakeSegment writes bad data into the segment rendering the segment unreadable.  In the case investigated closely the bytes returned when reading the dimension of the segment were not a valid descriptor.  This has happened 4 times over the course of 100s of hours of streaming of 100s of channels.

None of the simple test cases were able to consistently reproduce this problem. 

No error is returned by TreePutSegment.

It is possible that `_TreeXNciMakeSegment` is not threadsafe ?  This code is now a set of macros wrapping in-line functions.  Can someone (Timo) look at it?

We are attempting to verify that if the writers of the streams are in separate processes as opposed to just separate threads the problem does not occur.  **We do not have this confirmation yet**

zack-vii:
may not help for windows. i added the test to make for verification

### DIFF
--- a/testing/Makefile.am
+++ b/testing/Makefile.am
@@ -4,7 +4,7 @@ include @top_builddir@/Makefile.inc
 
 AM_CFLAGS = -I${top_srcdir}/include $(TARGET_ARCH) $(WARNFLAGS)
 AM_CXXFLAGS = -I${top_srcdir}/include @CXXFLAGS@ $(TARGET_ARCH) $(WARNFLAGS)
-           
+
 #
 # Static Base library
 #
@@ -47,7 +47,7 @@ JAVA_TEST_DIRS = \
 	java/mdsplus-api/tests \
 	java/mdsobjects/tests
 else
-JAVA_TEST_DIRS = 
+JAVA_TEST_DIRS =
 endif
 
 PYTHON_TEST_DIRS = \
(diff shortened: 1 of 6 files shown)


deploy/platform/redhat/redhat_build_rpms.py
```diff
@@ -243,8 +243,11 @@ def buildRpms():
             if child.status != 0:
                 sys.stdout.flush()
                 raise Exception("Error signing rpms. status=%d" % child.status)
-        except Exception as e:
```
joshStillerman: This is fixed by:

https://github.com/MDSplus/mdsplus/pull/2158

joshStillerman (commented):
Don't see the comment I stuck in the code....

The change to the exception in the build script is not needed.  Fixed by: https://github.com/MDSplus/mdsplus/pull/2158

deploy/platform/redhat/redhat_build_rpms.py
zack-vii: OK I skipped the commit

zack-vii:
Test should be good now.. this solution works way better than #2160. 
Through this PR threads are treated as foreign processes and will block untill teh file lock is released even if held by thread of the same process.  Windows seems to have done thread unique locks natively. 
it may make the pydevice workaround with threads self.dev = dev.copy() obsolete.. but this should be tested.

joshStillerman (approved):

PR #2167, opened 2020-11-17
Fix: Allow thick to open from OLD servers

joshStillerman:
The current alpha can open trees served by very old servers via
thick client.  In particular ones served by mdsplus built from:
```
4352e8d4978c4d2f375b2d0094078295e826bbbb
```
This ports alpha's routines involved in this in TreeOpen and
RemoteAccess into stable.

Note:  once the tree is opened, not all operations work. This is
true for the current alpha also.
```
$mdstcl

TCL> set tree test
TCL> dir
Error message was: %TDI-E-UNKNOWN_VAR, Unknown/undefined variable name
mdsdcl: --> failed on line 'dir'

Total of 0 nodes.
```


santorofer (approved):
The changes make sense to me. They look good.

PR #2168, opened 2020-11-17
Fix:handle xz(x) 2D visualization mode

GabrieleManduchi:
That 2D visualization mode was not correclty managed in Waveform


AndreaRigoni (approved):
Ok

PR #2172, opened 2020-11-19
Build: windows mdsip_service

zack-vii:
a new version of wine complains. return type is void

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -28,6 +28,8 @@ then
 	exit $abort
 fi
 
+conf/update_submodules
+
 aclocal -I ./m4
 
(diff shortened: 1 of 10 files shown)


zack-vii:
same here 
when i updated the docker image a newer gcc was complaining. nor sure it it actually changes the binaries now.. whihc is why i decided it is more a Build pr than a Fix.. but it avoids a -Werror under the new compiler

zack-vii:
fixes a race, and potential overflow and finished docker-srcdir implementation for trigger.

zack-vii:
retest this please

PR #2173, opened 2020-11-19
Build: windows mdsip_service

zack-vii:


joshStillerman (dismissed):
What is this about ?  the comment on the commit does not say anything useful?

zack-vii:
when i updated the docker image a newer gcc was complaining. nor sure it it actually changes the binaries now.. whihc is why i decided it is more a Build pr than a Fix.. but it avoids a -Werror under the new compiler

PR #2178, opened 2020-11-27
Fix:porting to java mdsobjects

GabrieleManduchi:
Use TreeNode method getNid() in place of old interface getInt()


AndreaRigoni (approved):
OK

PR #2179, opened 2020-11-30
Fix: move acq2106_435 from python 2 to python3

joshStillerman:
The current version of acq400_hapi from d-tacq is not python2
compatible.  Switching DAQ processes from python2 to python3
causes the error:
```
map object is not subscriptable
```
This is because the map is a generator.  The fix is to wrap the
2 map s in the code in list()


santorofer (approved):
Looks good.

PR #2182, opened 2020-12-02
Fix: Remove bug from super().__init__() in DeviceWorker class

santorofer:
The code in the DeviceWorker class had a missing class name in the call to:

super(_ACQ2106_423ST.MDSWorker.DeviceWorker, self).__init__()

i.e. 'MDSWorker' was missing in that call.


zack-vii (approved):
oh yes should be the full class path

santorofer:
Thank you Timo!

PR #2190, opened 2020-12-15
Feature: New ACQ2106 WRTD device

santorofer:
This is a new WRTD device for ACQ2106.

The device:
1- Sets the global WR trigger message, or WR_ID
2- Defines the WR RX matches: i.e. the messages the receiver will respond to.
3- Sets the WR receiver (RX) and and transmitter (TX)
4- Sets the WR trigger source, i.e. FPTRG or HDMI

Usage:
TCL> do /meth wrtd init
For hardware trigger (FPTRG or HDMI):
TCL> do /meth wrtd trig
For immediate transmit of specific message:
TCL> do /meth wrtd trig /ARG="""shared_all"""

### DIFF
--- a/pydevices/HtsDevices/acq2106_WRTD.py
+++ b/pydevices/HtsDevices/acq2106_WRTD.py
@@ -47,7 +47,7 @@ class ACQ2106_WRTD(MDSplus.Device):
         # If DIO-PG is present:
         {'path':':DIO_SITE',    'type':'numeric', 'value': 0.,       'options':('write_shot',)},
         # A DIO-PG source can be one of [d0...d5, TRGIN, WRTT]
-        {'path':':DIO_TRIG_SRC','type':'text',    'value': 'dx', 'options':('write_shot',)},
+        {'path':':PG_TRIG_SRC','type':'text',    'value': 'dx', 'options':('write_shot',)},
         
         {'path':':WR_INIT',     'type':'text',   'options':('write_shot',)},
             # ns per tick. (tick size in ns). uut.cC.WRTD_TICKNS: For SR=20MHz. It's the Si5326 tick at 20MHz ..
@@ -145,7 +145,7 @@ class ACQ2106_WRTD(MDSplus.Device):
 
         if not message.strip():
             # Set WRTD_ID: message to be transmitted from this device if FTTRG or HDMI is used to trigger.
-            print("WRTD_ID={} will be used. Waiting for external trigger ({})...".format(str(self.wr_init_wrtd_id.data()), str(self.trig_src.data())))
+            print("WRTD_ID={} will be used. Waiting for external trigger ({})...".format(str(self.wr_init_wrtd_id_glb.data()), str(self.trig_src.data())))
             uut.cC.WRTD_ID = str(self.wr_init_wrtd_id_glb.data())
             # Set trigger input on (FTTRG or HDMI)
             uut.s0.WR_TRG    = 1
--- a/pydevices/HtsDevices/acq2106_WRTD.py
+++ b/pydevices/HtsDevices/acq2106_WRTD.py
@@ -89,6 +89,9 @@ class ACQ2106_WRTD(MDSplus.Device):
         print("WRTD_ID {}".format(str(self.wr_init_wrtd_id.data())))
         uut.cC.WRTD_ID = str(self.wr_init_wrtd_id.data())
 
+        # Sets WR "safe time for broadcasts" the message
+        uut.cC.WRTD_DELTA_NS=self.wr_init_wrtd_dns.data()
+
         # Receiver:
         # Turn on RX
         uut.cC.WRTD_RX = int(self.wr_init_wrtd_rx.data())
--- a/pydevices/HtsDevices/acq2106_WRTD.py
+++ b/pydevices/HtsDevices/acq2106_WRTD.py
@@ -75,15 +75,6 @@ class ACQ2106_WRTD(MDSplus.Device):
         import acq400_hapi
         uut = acq400_hapi.Acq2106(self.node.data(), has_wr=True)
 
-        # WR TRIGGER SOURCE:
-        print("WR TRG Source {}".format(str(self.trig_src.data())))
-        uut.s0.WR_TRG_DX = str(self.trig_src.data())
-
-        # Global WR settings:
-        # Set WRTD_ID: message to be transmitted from this device if FTTRG or HDMI triggered.
-        print("WRTD_ID {}".format(str(self.wr_init_wrtd_id.data())))
-        uut.s11.WRTD_ID = str(self.wr_init_wrtd_id.data())
-
         # Sets WR "safe time for broadcasts" the message, i.e. WRTT_TAI = TAI_TIME_NOW + WRTD_DELTA_NS
         uut.s11.WRTD_DELTA_NS = self.wr_init_wrtd_dns.data()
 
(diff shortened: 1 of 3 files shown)


pydevices/HtsDevices/acq2106_WRTD.py
```diff
@@ -0,0 +1,148 @@
...
+
+    parts=[
+        {'path':':NODE',        'type':'text',    'options':('no_write_shot',)},
+        {'path':':HOSTNAME',    'type':'text',    'options':('no_write_shot',)},
+        {'path':':COMMENT',     'type':'text',    'options':('no_write_shot',)},
+        {'path':':TRIG_MSG',    'type':'text',    'options':('write_shot',)},
+        {'path':':TRIG_SRC',    'type':'text',    'value': 'FPTRG', 'options':('write_shot',)},
+        {'path':':TRIG_TIME',   'type':'numeric', 'value': 0.,       'options':('write_shot',)},
+        {'path':':T0',          'type':'numeric', 'value': 0.,       'options':('write_shot',)},
+        {'path':':WR_INIT',     'type':'text',   'options':('write_shot',)},
```
joshStillerman, line 52: Is it ever not 50 ?  If it is always 50 then lets make it a constant in the code

pydevices/HtsDevices/acq2106_WRTD.py
```diff
@@ -0,0 +1,148 @@
...
+            {'path':':WR_INIT:WRTD_RX_DTP', 'type':'text', 'value': "acq2106_999", 'options':('write_shot',)},
+            {'path':':WR_INIT:WRTD_DELAY',  'type':'numeric', 'value': 5000000, 'options':('write_shot',)},   # From uut.cC.WRTD_DELAY01
+            {'path':':WR_INIT:WRTD_ID',     'type':'text', 'value': "acq2106_999", 'options':('write_shot',)},
+            {'path':':WR_INIT:WRTD_TX',     'type':'numeric', 'value': 0, 'options':('write_shot',)},
+            {'path':':WR_INIT:WRTD_RX',     'type':'numeric', 'value': 0, 'options':('write_shot',)},
+
+        {'path':':RUNNING',     'type':'numeric', 'options':('no_write_model',)},
+        {'path':':LOG_OUTPUT',  'type':'text',    'options':('no_write_model', 'write_once', 'write_shot',)},
+        {'path':':INIT_ACTION', 'type':'action',  'valueExpr':"Action(Dispatch('CAMAC_SERVER','INIT',50,None),Method(None,'INIT',head))", 'options':('no_write_shot',)},
+        {'path':':STOP_ACTION', 'type':'action',  'valueExpr':"Action(Dispatch('CAMAC_SERVER','STORE',50,None),Method(None,'STOP',head))",'options':('no_write_shot',)},
```
joshStillerman: Stop?  Seems like it has an init method (which it makes sense to dispatch) and a trigger method that could be dispatched to fire the shot (but probably does not need an action as part of the device)

pydevices/HtsDevices/acq2106_WRTD.py
```diff
@@ -0,0 +1,148 @@
...
+        ]
+
+    def init(self, newmsg=''):
+        import acq400_hapi
+        uut = acq400_hapi.Acq2106(self.node.data(), has_wr=True)
+
+        if ":" in newmsg:
```
joshStillerman: what re the d0 and d1 msgs ?  Can you not only have one and not the other ?

pydevices/HtsDevices/acq2106_WRTD.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Transmiter:
+        # Turn on TX
+        uut.cC.WRTD_TX = int(self.wr_init_wrtd_tx.data())
+        #Commit the changes for WRTD TX
+        uut.cC.wrtd_commit_tx = 1
+
+    INIT=init
+
+    def trig(self, msg=''):
```
joshStillerman, line 113: Can you explain the trig method ?  Is this 'listen for trig'?  If so. do we want a repeat ? (probably don't need it) do we want a cancel ?

pydevices/HtsDevices/acq2106_WRTD.py
```diff
@@ -0,0 +1,148 @@
...
+# DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+# FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+# DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+# SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+# OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+import MDSplus
+import time
```
joshStillerman: Time is not used

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: Yes, I agree, stop action shouldn't be there. And it wouldn't be used in any case. I have forgotten to remove that node.

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: I left it open since I'm not sure if the value could change, due to a Peter's comments on one of the original versions of the default settings of the system. But it's true that it has been always 50 even when the TIGA system has a different clock.

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: I think this can be improved, yes. It's true that the two messages are independent of each other, since they use different signal paths. I can have one and not the other.

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: If you execute the trig method and a system is listening to that message, it will trigger. I thought that it could be good to have a way to trigger a system with a specific message, sent immediately, but maybe this will never be the case in a shot cycle?

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: For example, if we only set one of the signals, e.g d0, we can call it:
TCL> do /meth wrtd_161 init /ARG="""shared_all:"""
And the code will work. d1 will now have a string of length zero, which is used in line 100. Where the uut.cC (i.e. uut.11) will ignore it. But since in this case signal d1 is not used, this is ok. But I should probably improve this.

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: You are right. I removed it.

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: Josh, I was thinking that maybe the best will be to have two nodes for each of the set of messages that each dx can have. So that we don't need arguments to the method. I don't think we necessary needs to pass arguments.

santorofer:
Latest changes:
1- Now INIT initializes WRTD from nodes of the tree. No arguments to the function are needed. I.e. messages to be sent or received are set in the nodes .
2- TRIG() with no arguments sets the WRTD to be waiting for an external trigger
2- TRIG() with an argument sets it to send an immediate WR message. If the a system has the RX variable set with the same  message, it will trigger.
3- TRIG method no longer in an independent thread.

pydevices/HtsDevices/acq2106_WRTD.py
santorofer: Josh, so trigger has two ways of using it: with or without an argument.
Without arguments, sets the trigger so that it can be hardware triggered or trigger by a particular message in the WR network.
With an argument, a message is sent immediately. If the systems are armed and WR configuration ready, then the message will trigger them.

joshStillerman (approved):

PR #2191, opened 2020-12-16
Feature: New Generic WRPG device for acq2106

santorofer:
This is a new WRPG generic device for acq2106.

The device:
1- distinguishes between systems that contains:
    - GPG in site 0 (32 Channels DIO)
    - PG in site 1 to 6 (4 or 32 Channels DIO)
2- Sets the GPG/PG unit, e.g. enable the module, set the trigger line, etc.
3- Creates the STL table from output nodes
4- Load the STL table into the GPG/PG module

The GPG/PG module can then be triggered with a WR trigger message using the TRIG method. Or can be triggered by hard or soft trigger as well.


joshStillerman:
retest this please

joshStillerman (commented):
I am confused by the file: pydevices/acq400Devices/acq2106_pg.py.  This Python main is for debugging purposes ?

santorofer:
Since we have two different kinds of PG devices: 
4 Channels (TIGA) and 32 Channels (non-TIGA)

First the acq2106_pg.py file allows you to create a class for 4 channels or for 32 channels. Similar to the way the acq2106_435st.py allows you to say how many sites the system has.

For example, from mdstcl you could say:
TCL>  add node wrpg_161 /model=acq2106_pg_32

And the device will then be a 32 channels GPG.
I guess I could have this done all from the same file instead of having to files.
But the idea is to be able to say how many channels the WRPG will have.

santorofer:
Just to clarify, in the following commit:

"Add CH to the name of the classes for each channe"

I'm using our way of creating classes for the different number of channels, i.e classes for 4 and 32 channels.
On the other hand, in the following commit:

"Change the way the creation of the classes for different number of channels is done. Closer to D-Tacq design."

The way to create those classes is closer to what D-Taqc had in mind. The number of channels are part of a list of channel available, ie. [4,32]

Both ways are working fine to create a device that in the end has 4 or 32 channels.

joshStillerman (approved):

PR #2193, opened 2020-12-17
Feature:add sw trigger on MDSplus event

GabrieleManduchi:


GabrieleManduchi:
retest this please

GabrieleManduchi:
retest this please

joshStillerman:
retest this please

joshStillerman:
retest this please

GabrieleManduchi:
Hi Josh,
    I worked a bit with Nicolò, in order to figure out (remember) what is 
the correct type of the output in python GAM

It must be a tuple (this comes from the fact that python C interface 
exchanges only tuples) where the first element is np.ndarray with the same 
shape as the number of outputs. So the correct execute code for PyReset is

def execute(x):
   global reset
   if not reset:
     y = np.float64(0.)
   else:
     y = np.float64(1.)
     reset = False

   return np.ndarray((1,),y),   #note the final comma, it is a tuple whose 
only element is a numpy array of one element

the passed argument x is a numpy array with as many elements as the number 
of inputs

On my system it works, let me know what happens on yours.

                          Gabriele

GabrieleManduchi:
retest this please

GabrieleManduchi:
retest this please

AndreaRigoni (approved):
Great !

PR #2198, opened 2020-12-28
Fix: fallback to old locking schema if new not available

### LINKED ISSUE #2189: Arguments passed to fcntl() in io_lock_local  in treeshr/RemoteAccess.c make MDSplus non portable accross linux versions
The system routine fcntl() for locking a file in io_lock_local  (treeshr/RemoteAccessc) uses F_OFD_SETLK and  F_OFT_SETLKW as arguments, where their value is manually set as:
#define F_OFD_GETLK  36 
#define F_OFD_SETLK  37 
#define F_OFD_SETLKW  38

This is not portable and indeed it fails on Red Hat 7.2

The problem can be fixed by adopting the portable solution, i.e. using F_GETLKW, F_SETLK and F_SETLWK in place of the non portable F_OFD_SETLK and F_OFD_SETLKW values.

zack-vii:
not using a mutex yet.

file locking is kind of broken in the sense of how rwlocks work.. 

How old file locks work in threads
```
Thread1: lock file for write - get file lock
Thread2: lock file for read- no operation as process already has lock of higher quality 
Thread2: release lock - process releases lock
Thread1: may still be writing in now unprotected file section
Thread1: release lock - unlock error (suppressed)
```

Only way to fix this is to have a mutex making the file exclusive;
Using one shared mutex would make all file io exclusive though.

Any other way would basically require us to implement the OFD_LOCKS ourselves just to support it for OSes that we will drop in the future.

### DIFF
--- a/treeshr/RemoteAccess.c
+++ b/treeshr/RemoteAccess.c
@@ -35,9 +35,10 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  *
  * workaround: define flock so linux/fcntl.h will define lx_flock instead
  */
-# define flock lx_flock
-# include <linux/fcntl.h>
-# undef flock
+//// alpine does not linux/fcntl.h
+// # define flock lx_flock
+// # include <linux/fcntl.h>
+// # undef flock
 # ifndef F_OFD_SETLK
 #  define F_OFD_GETLK   36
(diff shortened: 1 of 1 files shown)


GabrieleManduchi (approved):
Ok, of course this shall not guarantee inter thread mutual file locking in old systems, but at least it is not worse than it was before. And hopefully newer systems will replace the old ones in the next years.....

PR #2211, opened 2021-01-12
Fix: python3 sockets take bytes not strings

### LINKED ISSUE #2212: cryocon18i code is not python3 compatible
The device support for the cryocon18i is not compatible with python3.  

In python3 the socket API expects and generates bytes not strings.

joshStillerman:
The HtsDevices are now run using Python3.  The socket calls in the
crycon18i device needs to send bytes, and receive bytes instead of
strings

### DIFF
--- a/pydevices/HtsDevices/cryocon18i.py
+++ b/pydevices/HtsDevices/cryocon18i.py
@@ -85,12 +85,14 @@ class CRYOCON18I_TREND(CRYOCON18I):
     ]
 
     def sendCommand(self,s,cmd):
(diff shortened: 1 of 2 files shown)


zack-vii (changes requested):
nope, you should fix this so it would work for both python2 and 3

e.g. use tobytes from MDSplus.version
https://www.mdsplus.org/index.php/Pydevice_dev#Support_python_2.7_and_python_3.x

zack-vii (approved):

PR #2216, opened 2021-01-13
Feature: Format all of treeshr using default clang

joshStillerman:
This is an attempt to apply clang-format to all of the code in
treeshr.

Feel free to reject this if you believe we should have more
customization of the format.


zack-vii:
What did you do to get this.. if you wish to keep it in this form, it would make sense to provider a command tool in e.g. conf that will redo the formating... one thing though.. the imports are sometimes sensitive to order, e.g. winsock2.h must be imported before windows.h
```
/usr/x86_64-w64-mingw32/sys-root/mingw/include/winsock2.h:15:2: error: #warning Please include winsock2.h before windows.h
```

zack-vii:
I rebased to alpha as it seems the windows tests was deactivated (my bad) and i fixed the order.. would be good to have a script that would do the formatting, that acknowledges the exceptions (import order).

joshStillerman:
I used clang-format -i filename.c

and -i filename.h

We I wonder if there are hooks on commit, or we have to do it in a push hook

-Josh

joshStillerman:
Did it reorder the includes ?  I am surprised

joshStillerman:
Timo -
  Here is an article about it... suggesting that we do this with our editors.
https://github.com/andrewseidl/githook-clang-format#:~:text=Once%20the%20pre%2Dcommit%20hook,repo%2C%20or%20set%20the%20hooks.

zack-vii:
Hm that could be the way to go. well we need to be able to tell it to keep the include order for some edge cases.. looks like it sorted them alphabetically. also I would like to add a manual script that would do it as well.. on the jenkins we could then run that script and reject the PR if the code does not comply with clang.

zack-vii:
We can add this and make clang a milestone with which we associate PRs that convert other libs.

zack-vii (approved):

joshStillerman:
Timo -

   I don't know anything about  / never heard of milestones.  I think 
all the initial formatting PRs should be 'semantics free'. There may be 
an option that says 'do not reorder includes', I will check.  The web 
page seems to say - program your tools to maintain this formatting, so 
that formatting does not get mixed with semantics in PRs.  Which makes 
sense to me.

I propose:

1 - look into include ordering

2 - look into failure modes

3 - maybe have a job that checks if the formatter is going to make any 
changes, and if so, does it, and makes a PR.  We could do this in a 
scheduled way.

4 - do the same for .py files.  (I don't think we want the default 8 
spaces!)

5 - add any formatting customization, and editor customization scripts 
to our git repo

Thoughts ?

-Josh

zack-vii:
Well i think it can reorder unless its critical.. like for the windows.h and winsock2.h so one would put a config file for the priority order or something. We would already benefit if we do in manually once and then stick with the formats. not sure if we would be able to agree on an IDE. I personally use a mix of nano and recently VS Code.. but for python sometimes Spyder. I think knowing what we want to got for is one thing. How we maintain the code as much in that style without breaking things is another.

A Milestone is something in git you can create and then you can assign PRs and issues to it so its easier to keep track of.

zack-vii:
e.g. We had a milestone "Improve code quality" to which i added this PR just now.

As for treeshr this PR is already quite good i would say .. who we tell a formater to keep the order is on a different page.

joshStillerman:
I think there are configs for all of the potential editors.

vi

vscode

atom

emacs

PR #2222, opened 2021-01-19
Feature: format all python code default pep8

joshStillerman:


zack-vii (approved):
looks like only white space changes.. the tests worked, so it did not seem to have introduced syntax errors. should be good

PR #2223, opened 2021-01-19
Fix: try2 to format cpp code

joshStillerman:


joshStillerman:
Timo -
  This builds fine in my directory but fails on tdi_hash undefined symbol.  Can't see why, could you look ?

joshStillerman:
Ohh I think I know

joshStillerman:
retest this please

joshStillerman:
retest this please

zack-vii (approved):
Sorry, I missed teh comment bout the TdiHash .. but yea, those generator files like 
opcbuiltins.h should not be formated. there should be more like this i think.. maybe at some point we should change the extension from .h to something else to emphasize that.

looks good now?

zack-vii:
hm i just realized doing stuff like 

```c
if STATUS_OK {
  this();
} else {
  that();
}
```

was a bad idea.

i should have kept it with ()

```c
if (STATUS_OK) {
  this();
} else {
  that();
}
```
clang converted it like this

```c
if 
  STATUS_OK {
    this();
  }
else {
  that();
}
```

on that note. do you have a command or script or profile now that one can use to format the files the same way as these commits?

PR #2228, opened 2021-01-27
Fix: mdsip with compression fails to decode data

### LINKED ISSUE #2227: MdsIpShr client cannot read back array data from mdsip in session mode with compression on
Under certain circumstances an MdsIpShr client will not receive the result of an evaluated TDI expression from the mdsip server. The LSB of the status returned by `GetAnswerInfo()` / `GetAnswerInfoTS()` / `GetAnswerInfoTO()` / `MdsValue()` remains 0, despite sending a perfectly valid TDI expression.

Necessary conditions for the problem appear to be:
1. The result has a certain degree of complexity: E.g. a result array of up to 8 integers works, one with more than 8 integers fails.
2. mdsip in session mode (neither "-m" nor "-s" options selected), i.e. mdsip is run through xinetd and stdin/stdout. No error found when mdsip is run in "multi mode" ("-m") and listens to a port itself.
3. mdsip allows compression, i.e. called with "-c (level)" option, level = 1..9. No error if the "-c" option is not given.

The issue appeared when I upgraded servers from older versions to 7.96 (stable) binary packages on Linux amd64, both with Ubuntu 20.04 LTS or CentOS7. It persists after I upgraded to 7.125.1 (alpha) on both systems.
The client version must be recent (>=7.96) for the problem to appear. I have found old clients (7.49 and some really ancient ones) to work fine with "buggy" servers. 
The issue propagates into python MDSplus, as `Connection.__getAnswer__()` throws a "SUCCESS" exception (sic!). Java clients seem not to be affected.

The C code below (which binds to libMdsIpShr) demonstrates the problem for a short integer array (which always works) and a long integer array (which does not work with a buggy server).

``` 
/* test of mdsplus client-server connection */

#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <ipdesc.h>

#ifndef GetAnswerInfoTO
int GetAnswerInfoTO(int id, char *dtype, short *length, char *ndims, int *dims,
                    int *numbytes, void **dptr, void **mout, int timeout_msec);
#endif

long mdsplus_socket = 0;
struct descrip mdsplus_param,
               mdsplus_result;
void print_data_item(struct descrip *d)
{
  int i, totlen;
  void *p;
  if (!d)
    return;
  printf("dtype: %d  length: %d  ndims: %d (", d->dtype, d->length, d->ndims);
  totlen = 1;
  for (i=0; i < d->ndims; ++i)
    {
      printf("%d ", d->dims[i]);
      totlen *= d->dims[i];
    }
  printf(") num items: %d \nData: ", totlen);
  p = d->ptr;
  for (i=0; i < totlen; ++i)
    {
      if (i>0) printf(",");
      switch(d->dtype)
	{
	case DTYPE_UCHAR:
	  printf("%c", ((char *) p)[i]);
	  break;
	case DTYPE_SHORT:
	  printf("%d", ((short *) p)[i]);
	  break;
	case DTYPE_LONG:
	  printf("%d", ((int *) p)[i]);
	  break;
	case DTYPE_LONGLONG:
	  printf("%ld", ((long int *) p)[i]);
	  break;
	case DTYPE_FLOAT:
	  printf("%f", ((float *) p)[i]);
	  break;
	case DTYPE_DOUBLE:
	  printf("%lf", ((double *) p)[i]);
	  break;
	}
    }
  printf("\n");
}

void test1(char *expr)
{
  int stat;
  printf("\nEXPRESSION: %s\n", expr);
  stat = MdsValue(mdsplus_socket, expr, &mdsplus_result, 0);
  printf("STATUS (MdsValue): %d\n", stat);  
  printf("RESULT: \n");
  print_data_item (&mdsplus_result);
}

int main()
{
  char *server = "mdsplus:8000";
  if ((mdsplus_socket = ConnectToMds(server)) == -1)
    {
      printf("Cannot connect to MDSplus server");
      exit(-1);
    }
  test1("0:7");    /* works */
  test1("0:10");   /* will not work with "mdsip -c 9" in session mode */
  if (DisconnectFromMds(mdsplus_socket) == -1)
      exit(-1);
}
```

joshStillerman:
fixes MDSplus/mdsplus #2227

The code was using the dlen from the message header instead of the
one copied out of the first four bytes of the message body
#closes #2227

### DIFF
--- a/mdstcpip/GetMdsMsg.c
+++ b/mdstcpip/GetMdsMsg.c
@@ -108,6 +108,7 @@ Message *GetMdsMsgTOC(Connection *c, int *status, int to_msec) {
       if (IS_OK(*status) && IsCompressed(header.client_type)) {
         Message *m;
         memcpy(&msglen, msg->bytes, 4);
+        dlen = msglen - sizeof(MsgHdr);
         if (Endian(header.client_type) != Endian(ClientType()))
           FlipBytes(4, (char *)&msglen);
         m = malloc(msglen);
--- a/mdstcpip/GetMdsMsg.c
+++ b/mdstcpip/GetMdsMsg.c
@@ -101,7 +101,7 @@ Message *GetMdsMsgTOC(Connection *c, int *status, int to_msec) {
         *status = SsINTERNAL;
         return NULL;
       }
-      unsigned long dlen = msglen - sizeof(MsgHdr);
+      unsigned long dlen;
       msg = malloc(msglen);
       msg->h = header;
       *status = GetBytesTO(c, msg->bytes, msglen - sizeof(MsgHdr), 1000);


joshStillerman:
retest this please

WhoBrokeTheBuild (approved):

PR #2234, opened 2021-02-12
Feature: Add Multi-sampling, i.e. re-sampling, of the data streaming

### LINKED ISSUE #2197: makeSegmentResampled from python
The current code for makeSegmentResampled and makeSegmentMinMax are in the cpp mdsobjects code. All of the current python code is implemented by calls into treeshr.

We need a way to call makeSegmentResampled from python.
We could write a C glue and call the C from python (not optimal)
We could move the C++ code for these functions to treeshr and then call them from C++ for the C++ API
  - Why does the C++ objects make its own entry points for all of the treeshr entry points it needs.  Wouldn't it be cleaner to only do this if there are C++ ish things it needs to do?

santorofer:
1- The new make segments with resampling options has been added to the device streaming:

makeSegmentResampled()

2- Two new extra nodes:

- RES_FACTOR
- INPUT_xxx : RESAMPLED

3- A function to get from the UUT the active sites:

getSlots()

TODO: 
- Add WR capability.

### DIFF
--- a/pydevices/HtsDevices/acq2106_435st.py
+++ b/pydevices/HtsDevices/acq2106_435st.py
@@ -57,45 +57,132 @@ class _ACQ2106_435ST(MDSplus.Device):
     """
 
     carrier_parts = [
-        {'path': ':NODE', 'type': 'text', 'value': '192.168.0.254',
-            'options': ('no_write_shot',)},
-        {'path': ':SITE', 'type': 'numeric',
-            'value': 1, 'options': ('no_write_shot',)},
-        {'path': ':COMMENT', 'type': 'text', 'options': ('no_write_shot',)},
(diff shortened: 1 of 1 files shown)


santorofer:
Hello Gabriele,

I have a quick question.
I have made this changes to the acq2106_435st.py device to make use of:

makeSegmentResampled()

But, should I be using:
makeSegmentMinMax()

 instead?
I can see the differences in the C code, but I'm not sure which of the two will be better for our streaming device. Will it depend on the actual signal that is being recorded.?

joshStillerman (commented):
I worry that we have acq2106_435st devices in existing trees and changing the node structure will break things.

joshStillerman:
retest this please

PR #2238, opened 2021-02-21
Gm grafana

GabrieleManduchi:
The instructions for installing the grafana plugin should work for the latest released plugin


AndreaRigoni (approved):
ok

PR #2241, opened 2021-02-24
Feature: New Acq2106 423ELF transient device

### LINKED ISSUE #2243: D-Tacq ACQ2106 435ELF and 423ELF transient support
Both D-Tacq devices, ACQ2106 with 435ELF and 423ELF needs to have MDSplus devices that support transient capture.

1- set pre-samples and post-samples
2- arm the digitizer
3- trigger pre-to-post transition. This is done by setting the trigger source for pre-sample to be a D-Tacq 'dx' signal and for post-samples an 'event0' signal.
4- store samples

santorofer:
New acq2106_423tr transient device:

1- Add to the acq2106_423st superclass:

pre-sample and post-sample tree nodes
trigger source and event0 source tree nodes

2- calibration expression written in the INPUT_xxx nodes
3- add a new parameter to the super-class init() method to be able to select between arming the digitizer when using the streaming device, or the transient device.

### DIFF
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -341,9 +341,11 @@ class _ACQ2106_423ST(MDSplus.Device):
         'fp_sync',      # Front Panel SYNC
         'wrtt1'         # White Rabbit Trigger
     ]
-    
-    def init(self, is_arm=1):
+
+    def init(self, armed_by_transient = False):
         uut = self.getUUT()
+
+        uut = acq400_hapi.Acq400(self.node.data(), monitor=False)
         uut.s0.set_knob('set_abort', '1')
 
         if self.ext_clock.length > 0:
@@ -417,13 +419,13 @@ class _ACQ2106_423ST(MDSplus.Device):
                 ch.COEFFICIENT.putData(float(coeffs[ic]))
 
         self.running.on = True
-        if is_arm:
-            # The following will arm the ACQ by this super class
+
+        if armed_by_transient is False:
+            # Then, the following will armed by this super-class
             thread = self.MDSWorker(self)
             thread.start()
         else:
-            if self.debug:
-                print('Skipping streaming from MDSWorker thread. ACQ will be armed by a sub-class')
+            print('Skip streaming from MDSWorker thread. ACQ will be armed by the transient sub-class device')
     INIT = init
 
     def stop(self):
@@ -465,8 +467,9 @@ def assemble(cls):
                 'type': 'NUMERIC',
                 'options': ('no_write_model', 'write_once',)
             },
+
             {
-                'path': ':INPUT_%3.3d:OFFSET' % (i+1,),     
+                'path': ':INPUT_%3.3d:OFFSET' % (i+1,),    
                 'type': 'NUMERIC',
                 'options': ('no_write_model', 'write_once',)
             },
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -420,7 +420,7 @@ class _ACQ2106_423ST(MDSplus.Device):
 
         self.running.on = True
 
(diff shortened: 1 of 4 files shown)


santorofer:
retest this please

santorofer:
retest this please

joshStillerman:
retest this please

pydevices/HtsDevices/acq2106_423tr.py
```diff
@@ -0,0 +1,275 @@
...
+
+    INIT=init
+
+
+    def arm(self):
+        import acq400_hapi
+        uut = acq400_hapi.Acq400(self.node.data())
+
+        shot_controller = acq400_hapi.ShotController([uut])
+        print("Using HAPI ShotController to run the shot.")
```
WhoBrokeTheBuild: These print statements should be wrapped in `if debugging()`

pydevices/HtsDevices/acq2106_423st.py
```diff
@@ -255,7 +255,7 @@ def run(self):
                         else:
                             self.full_buffers.put(buf)
 
-    def init(self):
+    def init(self, is_arm=1):
```
WhoBrokeTheBuild: Does `is_arm` mean it is already armed, or that we want to arm it?

pydevices/HtsDevices/acq2106_423tr.py
```diff
@@ -0,0 +1,275 @@
...
+
+        print("{}".format(uut.s0.transient))
+
+        # Initializing Sources to NONE:
+        # D0 signal:
+        uut.s0.SIG_SRC_TRG_0   = 'NONE'
+        # D1 signal:
+        uut.s0.SIG_SRC_TRG_1   = 'NONE'
+
+        #Trigger sources choices:
```
WhoBrokeTheBuild: Please put spaces between the `#` and first letter of your comments

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_423st.py
santorofer: Maybe the name of the variable is confusing. What it means is that:

a) if the value is 1, then, the super-class will arm it, i.e. by starting the MDSWorker() thread.
b) if the value is 0. then, the super-class will not arm it. I will be armed by the transient device instead. Which in this case is the code acq2106_423tr.py.

I needed to make the distinction, otherwise, when we try to do a transient recording, the super-class will arm the ACQ before we can start the transient recording.

santorofer: Maybe it should be:

0: if we are not using a transient recording, and so it will be armed, as usual, by the super-class device.
1 : if we want to arm it when only starting a transient recording

Could this be less confusing values for the variable?

pydevices/HtsDevices/acq2106_423st.py
```diff
@@ -313,8 +375,13 @@ def init(self):
                 ch.COEFFICIENT.putData(float(coeffs[ic]))
 
         self.running.on = True
-        thread = self.MDSWorker(self)
-        thread.start()
+
+        if armed_by_transient is False:
```
WhoBrokeTheBuild: `if not armed_by_transient`

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_423st.py
santorofer: Alright.

zack-vii:
is this still a thing?

santorofer:
Hi Timo,
Yes, this device will be the transient recorder for the D-Tacq 2106 423. It just need one more review and then be merge. Also, there is another device similar to this one, but for the 435 (that was the other PR that was open).
I will be sure that this one is merge and then re-open the other one and do the same.
Cheers!
Fernando

WhoBrokeTheBuild (approved):

PR #2262, opened 2021-03-19
update darwin to use @rpath loading

dgarnier:
This fixes most dependencies on DYLD_LIBRARY_PATH.  Other compiler flags changed to match linux/gcc.  Works on Big Sur with after bootstrap and from a fresh directory  (for openmotif installled with macports)

../mdsplus/configure --x-includes=/opt/local/include --x-libraries=/opt/local/lib --without-labview --exec-prefix=/usr/local/mdsplus --prefix=/usr/local/mdsplus

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -528,38 +528,32 @@ case "$host" in
 *apple-darwin*) echo "Configuring for MacOS X";
        THREAD=""
        MACOSX="macosx"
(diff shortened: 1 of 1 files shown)


joshStillerman:
test this please

joshStillerman (approved):

PR #2274, opened 2021-04-09
Fix:  bootstrap should always use python3

joshStillerman:
centos does not have default for python command.


zack-vii:
might be better to define an alias for centos8, python3 is not available on older os.

joshStillerman:
Which OSes?On Apr 13, 2021 7:06 AM, Timo Schroeder ***@***.***> wrote:
might be better to define an alias for centos8, python3 is not available on older os.

joshStillerman:
replaced by PR that chooses best python

Issue #2278, opened 2021-04-12
Potential error in treeshr/TreeRemoveNodesTags.c


merlea:
https://github.com/MDSplus/mdsplus/blob/e77417d944207cb06d93dcdbc0063b2f495f91fa/treeshr/TreeRemoveNodesTags.c#L141-L143

I came across this code in version `stable_release-7-0-X` which is in use at TCV (SPC-EPFL) and it seems to be still present in the latest versions. I think there is an error which was causing the containing loop to continue indefinitely.

If `swapint32(&info_ptr->tag_link) == 0`, then `dblist->tree_info->tag_info + swapint32(&info_ptr->tag_link) - 1` will be equal to `dblist->tree_info->tag_info - 1` which reaches before the `dblist->tree_info->tag_info` pointer and is probably not desired.

A possible solution is to change the condition in L141 to `swapint32(&info_ptr->tag_link) > 0`, as I think `tag_link=0` corresponds to the situation where no tag is associated to the node.

merlea:
I suppose this error should not be encountered for a healthy tree as the linked list of tag links starting from the node will contain the tag being removed. I do think this should still be fixed. Perhaps an error indicating that something is wrong about the tree should be thrown (I don't know if there's a good status already existing for this).

joshStillerman:
Do you have a 'broken' tree that this happens on?  And a scenario we can test to cause it?  The code for the tags is complicated and I would like to step through it before making a change.

merlea:
Here is a minimal example that replicates the situation. The original trees were modified using version 7-0-42 (or 7-0-X) which had a `memcpy` call with overlapping areas in `_RemoveTagIdx` which could on occasion cause corruption of the tag_info when removing a tag.

Note that I needed to include `treeshrp.h` to modify the tree information directly.

```
#include "treeshr.h"
#include "treeshrp.h"

int main()
{
  int status;
  int nid;

  status = TreeOpenNew("test",1) & 1;

  status = status && (TreeAddNode("node_1",&nid,0) & 1);
  status = status && (TreeAddTag(nid,"tag_1") & 1);

  status = status && (TreeAddNode("node_2",&nid,0) & 1);
  status = status && (TreeAddTag(nid,"tag_2") & 1);

  status = status && (TreeAddNode("node_3",&nid,0) & 1);
  status = status && (TreeAddTag(nid,"tag_3") & 1);

  if (!status) return 1;

[...]

PR #2287, opened 2021-04-15
Fix: fix exit condition in loop over tag links in _RemoveTagIdx

merlea:
Fix for #2278.

As mentioned in the issue, this loop exit condition should never be reached for a healthy tree. I thought about returning an error in case we do reach it but the function has a `void` return type.


joshStillerman:
test his please

joshStillerman (approved):
Thank you for your contribution

zack-vii (approved):

zack-vii:
test this please

PR #2288, opened 2021-04-15
Feature: MdsIpThreadStatic ConnectionList; should fix deadlock and be faster

### LINKED ISSUE #2285: race condition in remote IO
While working with MARTe2 which potentially has multiple threads trying to do calls to make_segment there are consistent lockups when using remote trees.  Especially thick client but also to a lesser extent distributed client.  

I think it comes from the fact that the remote locks are not operated atomically.   There is code in mds_io_request which takes out a local lock.  If it can not get it, then the enclosing calls never tell the remote side to give up the lock.  This is happening on the write lock of the data file when use_ofd_locks is 1 in the server code.

I tried to reproduce this using a variant of the segment test code from the test suite ,but could not get it to hang so I don't have a simple test for it.

It could be that a program that has 2 event AST routines attached to the same event that both try to call make_segment will do it.

zack-vii:
Added MdsIpThreadStatic to ensure threads use their own connections: prevents race conditions on server and deadlocks


zack-vii:
the python connection test fails.. might be because the mdsip is a single connection server (-s) server

joshStillerman:
Let me know when this is ready for review

zack-vii:
well the mdsip cleanup one is first.. this is still in progress. i had it almost but then it was not possible to share one python Connection object between threads

zack-vii:
retest this please

zack-vii:
@joshStillerman : its a lot to process but this should enable your parallel fast remote write issue. The changes i made to mdstcpip are deep but in the process several hidden issues were discovered and fixed.

You usecase is should be covered by the python test connection_write_test.py.

Going through each commit may be a bit complicated as i had to adapt the implementation several times, sometimes discarding changes i made earlier. I think the best would be to either trust our tests and observe its performance in a real live setup.

joshStillerman:
Wow -  This was a lot of work.  Can we go over it so that you are not the only one who knows what is here ?

zack-vii:
sure

python/MDSplus/connection.py
```diff
@@ -175,39 +182,80 @@ def __getAnswer__(self, to_msec=-1):
...
+        return self._get_answer(timeout)
+
+
+class Connection(object):
+    """Implements an MDSip connection to an MDSplus server"""
+
+    @property
+    def local(self):
+        try:
+            local = self._local.local
```
WhoBrokeTheBuild: We could consider calling this `self._local.conn` or something similar

WhoBrokeTheBuild (dismissed):

python/MDSplus/connection.py
zack-vii: like so ?

WhoBrokeTheBuild (approved):

PR #2289, opened 2021-04-19
Feature: cleanup Mdsip

zack-vii:
Eliminate MdsIpShr: mdsip needs a cleanup.. before i can efficiently tackler #2288


mdstcpip/mdsip_connections.h
```diff
@@ -247,6 +246,7 @@ EXPORT char ClientType(void);
 /// the server "MdsIpSrvShr" library.
 ///
 EXPORT int CloseConnection(int conid);
+int CloseConnectionC(Connection *coonection);
```
joshStillerman: connection?

mdstcpip/mdsipshr/Connections.c
```diff
@@ -272,8 +270,7 @@ Connection *NewConnectionC(char *protocol) {
     connection->readfd = -1;
     connection->message_id = -1;
     connection->protocol = strdup(protocol);
-    connection->id = INVALID_CONNECTION_ID;
-    connection->state = CON_IDLE;
+    connection->id = INVALID_CONNECTION_ID;    connection->state = CON_IDLE;
```
joshStillerman: why one line? '\n's are cheap

mdstcpip/io_routines/ioroutinestcp.h
```diff
@@ -400,7 +399,8 @@ static int io_listen(int argc, char **argv) {
             // add socket to active sockets //
             FD_SET(sock, &fdactive);
           }
-      } else {
+      }
```
joshStillerman, line 436: Can you give a bit of explanation here of what the code is doing ?  It used to have an else, now it always does both parts ?

zack-vii: This will force the server to also cycle thru the clients even if at the same time the server got a new connection

mdstcpip/mdsip_connections.h
zack-vii: its for internal use

joshStillerman: why not spell it connection instead fo coonection

zack-vii:
based on #2293: removes timeout feature, that seems to potentially cause more damage that it is useful.
cleanup of mdstcpip so it can be reimplemented to multi thread connections which would prevent dead-lock during fast remote write operations

zack-vii:
retest this please

joshStillerman:
Let me know when this is ready for review

joshStillerman (approved):

joshStillerman:
Is this PR closed ?

PR #2291, opened 2021-04-20
Fix: change boostrap to look for python

joshStillerman:
Since Timo did not like the previous solution:
  'assume python3'

Script now looks to see which python is installed.
Tries:
  python3
  python2
  python

### DIFF
--- a/bootstrap
+++ b/bootstrap
@@ -15,11 +15,22 @@ then
 fi
 }
 
+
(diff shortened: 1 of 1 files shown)


zack-vii (changes requested):
Please check my corrections in #2292 for the proper syntax

zack-vii:
merged as part of #2293

PR #2293, opened 2021-04-20
Build: used standard clang-format tools with style file

zack-vii:
This PR is the base of #2289


zack-vii:
added TreeFindNodeWild.c to yylex
added .clang-format style
added ./clang-format.sh script
added clang-format exclusions and fixups

* Standard C03 prevent "> >" to ">>" in cpp scripts 
* SED to unwrap multiline comments
*

joshStillerman (approved):

PR #2294, opened 2021-04-21
Gm jscope perf

GabrieleManduchi:
Under some conditions jScope was evaluating the entire waveform when not required, thus slowing display


GabrieleManduchi:
retest this please

joshStillerman (approved):

PR #2301, opened 2021-04-28
Fix:avoid division by zero foir empty segment

GabrieleManduchi:


GabrieleManduchi:
retest this please

zack-vii (dismissed):
Hi Gabrielle, can you please use clang-format to get the format checked out. 
You can do this by running the ./clang-format,sh script which does the required fixups as well.
```sh
./clang-format.sh xtreeshr
```

GabrieleManduchi:
Hi Timo, after installing clang, I get a lot of errors about unknown key 'ReflowComments'

GabrieleManduchi:
it is better you do it on a system where clang works properly

zack-vii:
i did but should the line not be
```c
int itemSize = (numData == 0) ? 0 : dataD->arsize / numData;
```

zack-vii:
Furthermore you should add a test with empty segments to check it this is a clean solution (memory safe).
numData == 0 is if numDims <= 1 : line 234
```c
  int numTimebase, numData = (numDims <= 1)
                                 ? (int)(dataD->arsize / dataD->length)
                                 : dims[numDims - 1];
```
Looks like it 'numData == 0' -> numTimebase = 0: line 246
```c
  // Check data array too short
  if (numData < numTimebase)
    numTimebase = numData;
```
which will lead to a memory violation around line 267
```
  if (endD)
  {
    status = XTreeConvertToDouble(endD, &end);
    if (STATUS_NOT_OK)
      return status;
    if (end > fulltimebase[numTimebase - 1])
      end = fulltimebase[numTimebase - 1];
  }
  else
    end = fulltimebase[numTimebase - 1];
```

zack-vii:
to me it looks like the easiest way to avoid downstream issues is to short circuit the arsize == 0 case.
something like 
```
if (arsize == 0)
{
copy in to out;
}
else
{
do normal resampling
}
form answer;
```

GabrieleManduchi:
right!
I have changed things to handle empty segments explicitly.
Moreover I discovered another error: at line 241, the input signal was wrongly passed as pointer to pointer

AndreaRigoni (approved):
Ok !

PR #2304, opened 2021-05-04
Build: platform builds on empty/new release folder and format

zack-vii:
Should automatically heal/setup /release and /publish folder for new oses


zack-vii:
also fixed signing redhat packages for python3 dists like fc32

PR #2305, opened 2021-05-04
Fix: Add validation for the user's selection of the frequencies for the ACQ 2106 435 SC device.

santorofer:
The class ACQ2106_435SC has now a series of user's input validation for the allowed frequency choices that the acq2106 435SC has.

The allowed frequencies are:

40KHz (boot-time default)
20KHz
10KHz

Otherwise, an exception is raised.

### DIFF
--- a/pydevices/HtsDevices/acq2106_435sc.py
+++ b/pydevices/HtsDevices/acq2106_435sc.py
@@ -85,7 +85,7 @@ class _ACQ2106_435SC(acq2106_435st._ACQ2106_435ST):
             clock_div = 4
         
         # For now, the frequency must be a clean division of 40000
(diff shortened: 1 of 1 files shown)


santorofer:
I have done 3 tests for those 3 choices and all 3 tests have passed.

santorofer:
I have done 3 tests for those 3 choices and everything is set correctly.

santorofer:
This PR is closed because D-Tacq has changed sync_role() so that certain SR frequencies can be used. Therefore, the corresponding clock plan is automatically chosen.
See, also related closed PR, "Fix: explicitly setting the WRTD_TICKNS from tree node value #2323"

PR #2309, opened 2021-05-10
Build: ub20 does not have python_numpy

joshStillerman:
The dependecies in linux.xml  Only distinguish on 'platform'.
UB20 has different dependencies than UB<20.  Added new section to
linux.xml for specifying DIST dependencies and logic to
debian_build_debs.py to find and use this.

Eventually this should be refactored so that the 'DIST' section is
inside of the 'PLATFORM' section.

the RHEL8 hack (its own XML) can probably leverage this same thing.


zack-vii (commented):
I am willing to patch the scripts can you refresh my memory ,, what was so special about centos8 ?
is it the python vs python2 ?

zack-vii:
@joshStillerman  This should do it. I changed the jenkins jobs to build packages for centos8 and ubuntu20 on pull-requests

zack-vii:
Failed on alpine-3.6-armhf because it got stuck in java tests, unrelated though.

zack-vii (approved):

PR #2312, opened 2021-05-14
Gm beans

GabrieleManduchi:
new java beans support for DW setup of MARTe2 Simulink devices
new MARTE2_SIMULINK devices 
structure changed of MARTE2 devices
RedPitaya support


AndreaRigoni (approved):
ok

PR #2315, opened 2021-05-27
Feature: jDevices setup form for WRTD

santorofer:
A device setup form was built.
It contains the necessary menus to create a WRTD device for ACQ digitizers as well as for PG/TIGA systems.

Example:
Screen Shot 2021-06-23 at 1 38 58 PM


zack-vii:
hm, would it not make sense to layout the components of the form to better fit a screen. also one tab for one lonely text field seems odd. maybe you can fit the ones from the top in there too and rename the tab.

santorofer:
Hi Timo, yes, I see what you mean. This is my first pass to the actual design of the form. What it's good is that the form works well when using it.
By creating those tabs I thought it will be clear what is what, instead of put them all in one page. But I understand whaat you mean. I will re-think it.

santorofer:
GUI's test passed.

santorofer:
I will merge this PR so that you can try out  this WRTD  Java form in alpha.

To test the form, one can create a test tree with one device: 
- TCL> add node wrtd /model=acq2106_WRTD
- Open tree with jTraverser, then use "Setup Device"

PR #2318, opened 2021-06-01
Fix: mdstcpip memory issue with jscope

zack-vii:
There was a double free causing the crash. also SupportsCompression is not a flag of client_type but according to jScope a bit in the status filed. .. jScope seems to be the only client that cares about compression but does not support DTYPE DSC .. I assume the code that i tried to revive there was commented out for that reason. I removed it, whoever in two commit in case we want this back some day.
Fixes #2318


GabrieleManduchi (approved):
Thanks!

Issue #2319, opened 2021-06-09
xlate_logical_name not external in camshr


joshStillerman:
Brian Nelson writes:
Hi All,

I'm trying to get a CAMAC physical name from a CTS logical name in TDI.  (If there is a slicker way to do it, I'm quite open.)

These TDI lines used to work for an older Ubuntu install to get the CAMAC physical name from the CTS logical name:

_CamKey = zero(4, 0b);

CamShr->xlate_logicalname(_LogicalName, ref(_CamKey));

where _LogicalName is a string of the CTS logical name, and I can parse the _CamKey structure for the physical name.  However now we get an error:

Error: /usr/local/mdsplus/lib/libCamShr.so: undefined symbol: xlate_logicalname

However, if I do:

$ strings /usr/local/mdsplus/lib/libCamShr.so | grep xlate
xlate_logicalname()
xlate(): name['%s'] ==>> HA[%c] id[%d] crate[%02d] slot[%d]
xlate_logicalname.c
xlate_logicalname

it seems to be there.

Any ideas or suggestions for alternative methods?

Thanks,
Brian

-- 
Brian A. Nelson
Research Professor Emeritus
Department of Electrical & Computer Engineering
University of Washington, Seattle

joshStillerman:
Brian -

  The problem is that the entry point is not external.  I will take a quick look to see if there is a different way, and then when there is not, I will move it from camshrp.h to camshr.h and make it external.

In the future, try adding questions / requests like this to our github issues list.  Your posts will get more eyes on them that way.

-josh

banelson:
Thanks Josh.  Was this call external previously?  (It used to work. ;)

Is there another way in TDI to do an mdscts "SHOW <logical_name>" and get the physical CAMAC address?

I guess I could wrap a Python cts() call and parse the output.

-Brian

nelson-zapenergy:
Update:  Josh's extern entrypoint works, thanks!

PR #2321, opened 2021-06-09
Feature: Add extern entrypoint CamXlateLogicalname

joshStillerman:
A user wants to call xlate_logicalname in camshr.  New entrypoint
added to expose this routine.

Closes issue:
https://github.com/MDSplus/mdsplus/issues/2319


santorofer (approved):
It looks good to me.

PR #2323, opened 2021-06-14
Fix: explicitly setting the WRTD_TICKNS from tree node value

santorofer:
The WRTD_TICKNS is now being set when INIT the device. The value is taken from the corresponding tree node:

uut.cC.WRTD_TICKNS = self.wr_init_wrtd_tickns.data()

The WRTD_TICKNS (clock tick interval in nsec), is the result of:

1. / MBCKL

where the MBCKL is the mother board clock in MHz, which comes from  using  new defined clock plans. The plan is automatically selected by the sync_role() script, when the sample rate is input. The choices are:

10 kSPS (5M12 / 512)
20 kSPS (10M24/512)
40 kSPS (20M48/512)
80 kSPS (20M48/256)
128kSPS (32M768/256)

For example, if the sample rate is chosen to be 20KHz, then, the plan 10M24 is automatically selected by sync_role(). Therefore, the value for WRTD_TICKNS should be:

WRTD_TICKNS = 1./10240000 = 0.00000009765625 secs = 97.65625 nsecs

### DIFF
--- a/pydevices/HtsDevices/acq2106_WRTD.py
+++ b/pydevices/HtsDevices/acq2106_WRTD.py
@@ -38,72 +38,159 @@ class ACQ2106_WRTD(MDSplus.Device):
     """
 
     parts = [
-        {'path': ':NODE',        'type': 'text',
-            'options': ('no_write_shot',)},
-        {'path': ':HOSTNAME',    'type': 'text',
-            'options': ('no_write_shot',)},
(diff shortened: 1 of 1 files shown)


santorofer:
This PR supersedes the closed PR #2305: 
"Fix: Add validation for the user's selection of the frequencies for the ACQ 2106 435 SC device."

joshStillerman (approved):

PR #2326, opened 2021-06-21
Fix: Compare ctx.value instead of ctx, fixes #2325

### LINKED ISSUE #2325: python: after getting active tree with MDSplus.Tree(), setDefault() fails: "TreeNode must be in same tree"
version: mdsplus-stable (7.96-17)

I'm not sure if this is a bug, or I'm not following correct practice. In my setup, I call a python function from tdi. In python, if I get handle of the active tree with MDSplus.Tree(), I can't use setDefault().

To reproduce failure:
```
TreeOpen("my_tree",-1)
py("import MDSplus; t=MDSplus.Tree(); print(t); t.setDefault(t.getNode('\\TOP.TXT'))")
```

output: 
```
Tree("MY_TREE",-1,"Normal")
  File "/usr/local/mdsplus/tdi/python/Py.py", line 58, in Py
    exec(compile(cmds,'<string>','exec'),env,env)
  File "<string>", line 1, in <module>
  File "/usr/local/mdsplus/python/MDSplus/tree.py", line 954, in setDefault
    raise TypeError('TreeNode must be in same tree')
TypeError: TreeNode must be in same tree
0
```
I tracked it down to:  [/tree.py#L1050](https://github.com/MDSplus/mdsplus/blob/68cba11e7ca48b0990eea9714273bba714f7878b/python/MDSplus/tree.py#L1050) :
```
if not node.ctx == self.ctx:
    raise TypeError('TreeNode must be in same tree')
```

I even printed out, and the both c_void_p object has same value:
```
TDI> py("import MDSplus;t=MDSplus.Tree();print(t.ctx);print(t.getNode('\\TOP.TXT').ctx)")
c_void_p(94186050127232)
c_void_p(94186050127232)
```

This problem is ONLY occuring if I use Tree() to get the active tree already opened in TDI. Following works fine:
```
TreeOpen("my_tree",-1)
py("import MDSplus; t=MDSplus.Tree('my_tree',-1); print(t); t.setDefault(t.getNode('\\TOP'))")
```

cenkoloji:
When instance of active tree is created (i.e. Tree()), comparing ctx
of node and tree result in node seeming to be from another tree

Note that there is another instance of ctx comparison: [tree.py#L697](https://github.com/MDSplus/mdsplus/blob/68cba11e7ca48b0990eea9714273bba714f7878b/python/MDSplus/tree.py#L697).
I'm not sure if the same should be applied there.

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -1047,7 +1047,7 @@ class Tree(object):
         old = self.default
         if not isinstance(node, TreeNode):
             raise TypeError('default node must be a TreeNode')
-        if not node.ctx == self.ctx:
+        if not node.ctx.value == self.ctx.value:
             raise TypeError('TreeNode must be in same tree')
         _exc.checkStatus(
             _TreeShr._TreeSetDefaultNid(self.ctx,
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -694,7 +694,7 @@ class Tree(object):
 
     def __eq__(self, obj):
         if isinstance(obj, (Tree,)):
-            return self.ctx == obj.ctx
+            return self.ctx.value == obj.ctx.value
         return False
 
     def __ne__(self, obj): return not self.__eq__(obj)


cenkoloji:
Related to issue #2325

zack-vii:
test this please

zack-vii (approved):
looks ok

PR #2331, opened 2021-06-25
Fix: removed extra free method in PutRecordRemote

### LINKED ISSUE #2330: PutRecordRemote free()'s caller's data
==28373==ERROR: AddressSanitizer: attempting free on address which was not malloc()-ed: 0x7fffffffb320 in thread T0
    #0 0x5047cf in __interceptor_free ../../../../libsanitizer/asan/asan_malloc_linux.cpp:123
    #1 0x7ffff66952de in PutRecordRemote (/usr/local/mdsplus/lib/libTreeShr.so+0x332de)
    #2 0x7ffff668200d in _TreePutRecord (/usr/local/mdsplus/lib/libTreeShr.so+0x2000d)
    #3 0x7ffff781c424 in putTreeData (/usr/local/mdsplus/lib/libMdsObjectsCppShr.so+0x50424)
    #4 0x7ffff7804b57 in MDSplus::TreeNode::putData(MDSplus::Data*) (/usr/local/mdsplus/lib/libMdsObjectsCppShr.so+0x38b57)

The putData() interface is such that it is the caller's responsibility to free the data (which makes sense, as the data can come from anywhere: a custom allocator, new'd data, stack data, alloca data, a chunk of another array, or something that just shouldn't be free'd at all, etc... none of these are safe to pass to free().)  But this code path seems to indicate that calling TreeNode::putData(ptr) will force that free(ptr) is called.

zack-vii:
I wanted to fix #2330. I am not unsure if it does. Seems like our tests dont cover this issue and i could not reproduce it. Nevertheless, one free seems extra.

### DIFF
--- a/mdstcpip/mdsip_connections.h
+++ b/mdstcpip/mdsip_connections.h
@@ -332,7 +332,7 @@ EXPORT Connection *PopConnection(int id);
 
 client_t GetConnectionClientType(int id);
 
-int GetConnectionVersion(int id);
+int MdsIpGetConnectionVersion(int id);
 
 extern void FlipData(Message *m);
 extern void FlipHeader(MsgHdr *header);
--- a/mdstcpip/mdsipshr/Connections.c
+++ b/mdstcpip/mdsipshr/Connections.c
@@ -116,7 +116,7 @@ Connection *FindConnectionSending(int id)
   return c;
 }
 
-EXPORT int GetConnectionVersion(int id)
+EXPORT int MdsIpGetConnectionVersion(int id)
 {
   MDSIPTHREADSTATIC_INIT;
   Connection *c = _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR);
--- a/mdstcpip/mdsipshr/Connections.c
+++ b/mdstcpip/mdsipshr/Connections.c
@@ -119,10 +119,8 @@ Connection *FindConnectionSending(int id)
 EXPORT int GetConnectionVersion(int id)
 {
   MDSIPTHREADSTATIC_INIT;
-  int version;
   Connection *c = _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR);
-  version = c ? (int)c->version : -1;
-  return version;
+  return (c) ? (int)c->version : -1;
 }
 
 Connection *FindConnectionWithLock(int id, con_t state)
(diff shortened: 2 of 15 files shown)


kgerickson:
It might be worth waiting on merging this until we can better understand why it only affects connections between a new client and an old server.  I recall hitting a bug with similar server version characteristics, and the root cause was not so trivial.

kgerickson:
I think perhaps it would be good to split "@zack-vii
removed extra free method in PutRecordRemote" into two commits, one to do as the commit says, and another to rename the variable.  The commit of removing the extra free should only be a single line. This will help with future history management (for instance, with git bisect, or some such thing).

zack-vii:
well we squash commit anyway. i need to check the MDSIP_MAX_VERSION sswitch for windows though.. for whatever reason the confirming print does not show in tditest

kgerickson:
Why do you squash commits? That explains why trying to go through the code history is a nightmare.  Can you stop doing that? :)

kgerickson:
As a side note, this is currently blocking mhd data recording on KSTAR, so if you could backport this to the stable branch when it's fixed, I'd appreciate it. I've run a few days of shots without being able to record, which is obviously not ideal.

zack-vii:
does it fix your issue though.. i added the test to get closer to your setup, i.e. older server/new client. but it is just a simulation and the core issue may be somewhere else.

zack-vii:
seems like wine swallows the confirmation print... use MDSMSG if env is defined (those are filtered out by default). now test should not fail under windows.
also the connection version is now adjusted during connect, which should reduce its impact on performance.

zack-vii:
@kgerickson : lacking time, I did not split the commit, but you may do so if you wish. Squashing makes it easier to apply patches to stable as well as it is only one cherry-pick alpha. I think given that we usually debug by releases to narrow down the issue we did not come across the history mess so  often. On the other hand we sometimes forget to commit while working on a patch resulting on two things happening at once, its pretty tedious to disentangle the steps when the share lines.  ;)
well if we squash or not.. i fine either way.. but squashing leads to a linear history without rebasing. I think that is why we decided on squashing.

zack-vii:
retest this please

kgerickson:
I'm building your branch on the affected system right now, will report back.

kgerickson:
I can confirm that this branch fixes my particular problem. I don't fully understand the change, so I can't review it, but at least it works for me.

I can split that one commit, but it will require rebasing this branch. I hope that's ok.

EDIT: I split the commit and pushed it. This should be good to merge.

EDIT2: what's the current process for backporting this to the release branch?

zack-vii:
hm i could make a branch for stable.. not sure why the thread is not cleaning ip propery (asan test). i had it run for many iteration yesterday and it did not trip once. but jenkins failed again.. maybe i should switch back to processed (aka local://1)

kgerickson (commented):
(This comment is in response to the review request)
The changes appear ok to me, but I don't have enough knowledge of the internals to really evaluate them. My previous comment still applies, though, that commits should be specific and self contained.  For example, there are still unrelated changes in these that are not reflected in the commit message, making review harder.  For the future, I would suggest:

1. Be more clear in the commit message as to what is changing and why
2. Separate out unrelated changes into individual commits
3. Don't squash merges; instead, rebase them onto  alpha / stable to give a clear and unambiguous commit history

Those points are generic, though, and so perhaps I should have a new github issue about them.

zack-vii:
The reason why we do squash merge ,, now that i think of it is as this will trigger the version increments minor or release, it is based on the commit message of the  merge. so we would have to think of a new logic/mechanism to figure out what kind of improvement the merge request aimed at. That should be considered in the issue as well.

PR #2332, opened 2021-06-25
Docs:  Fix name of TreeShrHook.py example

joshStillerman:
The distributed TreeHook.py.example is old and would not be called
if put in place.  This new one will work if enabled.


zack-vii (dismissed):
fixed indent but look good otherwise

PR #2346, opened 2021-07-12
Fix:change incorrect jScope property file

GabrieleManduchi:


AndreaRigoni (approved):
Ok, but you will have to move all files in all the servers then .. users got used by this new directory name.

PR #2355, opened 2021-07-19
Fix: Get clock plan and calculate the correct nanosec per ticks for the ACQ2106 WRTD TICKNS.

santorofer:
The changes are in the acq2106_WRTD device.

1- Now the clock plan name is checked directly from the ACQ.
2- The name is used to to calculate the correct nanosec per tick value for the ACQ's uut.cC.WRTD_TICKNS


WhoBrokeTheBuild (approved):
Looks good, great comments

PR #2356, opened 2021-07-19
Fix: Influxdb signal with delta step

santorofer:
Add INFLUX GROUP BY to the query:

groupBy = 'GROUP BY time(%sms)' % (deltaTimeMS,)

if the getSetTimeContext() contains the delta step. 

Includes changes to 435st and 423st so that the start of the shot, the trig_time is recorded into the node at the start of stream.

### DIFF
--- a/tdi/python/influxSignal.py
+++ b/tdi/python/influxSignal.py
@@ -23,6 +23,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
+import re
 import MDSplus
 
 try:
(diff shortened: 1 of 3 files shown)


tdi/python/influxSignal.py
```diff
@@ -125,10 +158,26 @@ def influxSignal(fieldKey, where, series=None, database=None, config=None, shotS
...
+    groupBy = ''
+    if deltaTime is not None:
+        # Convert deltaTime from seconds to milliseconds, if deltaTime is less than 1ms this will break
+        deltaTimeMS = int(deltaTime * 1000)
+        groupBy = 'GROUP BY time(%sms)' % (deltaTimeMS,) # The trailing 'ms' is for milliseconds
+        # If the fieldKey does not have an aggregation function, Influx will error out
+        # Attempt to wrap the field key with an aggregation function, but don't affect
+        # any equations stored in it
+        # For example:
+        #   10*fVal will become 10*MEAN(fVal)
```
WhoBrokeTheBuild, line 170: It turns out that you can't do math unless there's an aggregation (I believe), so we could potentially simplify this section and make it just wrap the fieldKey without the regex

WhoBrokeTheBuild (approved):

PR #2361, opened 2021-08-11
Build: Add support for nix packaging (flake)

fluffynukeit:
This PR makes the mdsplus repo a nix flake by adding `flake.nix` and `flake.lock` files.

I am new to mdsplus and started going through [this tutorial](https://www.mdsplus.org/index.php?title=Documentation:Tutorial&open=76203664636339686323781681&page=Documentation%252FThe+MDSplus+tutorial) to get acquainted.  I saw in the mdsplus repo and [MDSplus/Docker ](https://github.com/MDSplus/Docker) repos that Docker is used for heavily for both building mdsplus from source as well as running mdsplus programs without building.  This approach is reasonable and is one I have tried before myself many times but have been frustrated by all the hoops to jump through with `docker-compose`, environment variables, mapping X11 permissions, mapping external files into the container, inscrutable command line options, and more.  My opinion is that Docker is a good fit for deploying web services to the cloud, but a poor fit for a CLI or GUI interactive ecosystem.

Nix flakes from the nix package manager are an alternative build and installation solution that some, like myself, might find more approachable or more robust in this case.  An overview of flakes is provided [here](https://www.tweag.io/blog/2020-05-25-flakes/).  In short, the flake provides a fully specified build and install recipe for a package and all its dependencies, and installs in a way that is isolated from the rest of the user's system.  Here is an example.

```
dan@dan-desktop:~$ mdstcl
mdstcl: command not found
dan@dan-desktop:~$ jTraverser
jTraverser: command not found
dan@dan-desktop:~$ export my_tree_path=$(realpath Desktop/my_tree_tutorial)
dan@dan-desktop:~$ nix shell github:fluffynukeit/mdsplus/stable
dan@dan-desktop:~$ mdstcl
TCL> set tree my_tree
TCL> dir

\MY_TREE::TOP

 :NUM1         :NUM2         :NUM3         :TXT         

  MY_SUBTREE    SUB1        

Total of 6 nodes.
TCL> exit
dan@dan-desktop:~$ jTraverser
dan@dan-desktop:~$ exit
exit
dan@dan-desktop:~$ mdstcl
mdstcl: command not found
```

The `nix shell` command will build and/or install the package and all its dependencies the first time it's called.  Most of the dependencies are cached in binary form and don't need to be built from scratch.

I don't think that the PR is ready to be taken as-is because I don't have enough experience with mdsplus to test everything or anticipate every problem.  I'm hoping this PR starts a discussion on whether this path is a viable alternative to the docker route.  I'm hoping being able to more easily build from source helps to encourage more contributions from the community.

## Anticipated Issues Collected from Discussion

1. Mdsplus gets installed in the nix store, which is read-only.  I've only skimmed the surface of the tutorial so far, but I have seen that new TCL functions and python functions are added by putting new files into the MDSPLUS_DIR path (read only in this case).  Is there any method of adding access to the files using environment variables instead of modifying the contents of MDSPLUS_DIR?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ The MDSplus developers want to know who you are. If you or your site is using MD
 [MDSplus User Survey](https://docs.google.com/forms/d/e/1FAIpQLScSsA-fY2yTsW076bBreJmNbBqY9jsd-m4vmAdPvfCxXidiOQ/viewform?usp=sf_link).
 
 ## Building MDSplus
+
+### Building from source manually
 To build and install MDSplus on unix systems, you will need to obtain the
 MDSplus distribution from github either as a git repository or as a 
 compressed tar file.
@@ -35,6 +37,7 @@ When MDSplus is built to generate releases for public distribution a deploy
 script is used which uses Docker to pull specially configured Docker images
 from https://hub.docker.com/r/mdsplus/docker/
 
+### Building using Dockerized build environment
 You can build MDSplus for any of the linux and windows operatings systems
 without needing to install any special compilers or libraries. If you have
 the mdsplus source code and Docker installed on your linux system and your
@@ -60,6 +63,20 @@ You can find the available operating systems that you could specify for the
     # ls <mdsplus-src-dir>/deploy/os/*.opts
 
     
+### Building with nix package manager
+The mdsplus repo is a nix flake, meaning that it specifies a build recipe
+for mdsplus and all of its dependencies.  You can automatically download,
+build, and install mdsplus using the following nix command (after enabling the 
+currently experimental flakes feature):
+
+    nix shell github:mdsplus/mdsplus/<branch or tag name>
+    
+Upon first invocation, the above command will build and install mdsplus into
+the nix store, and then put the user into a subshell in which the mdsplus
+programs are available on the PATH.  The commands will be accessible until
+the subshell is exited.  In subsequent invocations, the user will go directly
+into the subshell without rebuilding or reinstalling.
+
 ---------------------------------------------------------------------------
 Who Uses MDSplus
 This map shows world fusion sites using MDSplus.  
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1622516815,
+        "narHash": "sha256-ZjBd81a6J3TwtlBr3rHsZspYUwT9OdhDk+a/SgSEf7I=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "7e9b0dff974c89e070da1ad85713ff3c20b0ca97",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "21.05",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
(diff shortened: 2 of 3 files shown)


joshStillerman:
Daniel -
  Thank you for your PR.  I had never heard of the nix distribution. Until we can set up a docker to build and test the install, I think the right place for this would be in the User Contrib section of mdsplus.org <http://mdsplus.org/>.  There is a downloads page: https://www.mdsplus.org/index.php?title=User_Contrib:downloads&open=60654047435862686302239&page=User+Contrib.%2FDownloads <https://www.mdsplus.org/index.php?title=User_Contrib:downloads&open=60654047435862686302239&page=User+Contrib./Downloads> that it could be added to. 

I defer to the rest of the MDSplus core developers if they think otherwise.  Did you build for nix or just create the packaging from one of our installation platforms?

Where are you from?  What are you working on?  You are welcome to join the MDSplus development effort, or just become a contributor to www.mdsplus.org <http://www.mdsplus.org/>. I think there is a button that says, ask for an account.

joshStillerman:
Sorry I read further down in your PR and have a few more answers for you:

- you should not need to write to /usr/local/mdsplus ($MDSPLUS_DIR) with the possible exception of /usr/local/mdsplus/local. which is left undisturbed during updates
- MDSplus looks for extensions to the TDI interpreter in a list of places specified by the **';'** delimited list in the environment variable MDS_PATH.  For example we use:
```
jas@mfews03:~$ echo $MDS_PATH 
/usr/local/cmod/tdi;/usr/local/mdsplus/tdi
```
- MDSplus looks for additional device method code in `MDS_PYDEVICE_PATH `.  I believe this is also **;** delimitied.  Ours only has one element:
```
jas@mfews03:~$ echo $MDS_PYDEVICE_PATH 
/usr/local/mdsplus/pydevices

```
There is also $MDSMSG_PATH I don't know off the top of my head how this is used, it currently points to  '/usr/local/mdsplus/lib', so it is probably a place to look for shared image to find error messages in.

I apologize again for the piecemeal response, We will more fully read and digest your PR as soon as we have time.
-Josh

fluffynukeit:
@joshStillerman , thanks for your responses.  I do want to clarify that nix is a package manager available for any Linux distribution.  For instance, I am running nix on Ubuntu 20.04.  There is a related effort, NixOS, that is a Linux distribution that extends the philosophy of the nix package manager to the entire system, not just installed package management, but NixOS is not needed to install packages with nix.

In creating the nix "derivation" (their word for "build recipe"), I did not use docker at all or make any assumptions about a particular Linux distribution.  Like docker, nix derivations allow software to be installed and run in a manner that is isolated from the rest of the host system, packaged with all its dependencies.  It differs in the manner in which isolation is achieved, though, which has nice consequences for the user experience for CLI and GUI interfaces, IMO.  In short, I expect that any Linux host computer with nix installed (and nix flake support enabled) could run the commands I gave and be similarly successful as I was.

I am from WA state and am joining the Zap Energy Inc startup next week.  Part of my responsibilities will include managing mdsplus.  I stared to dig into it but kept running in to an all-too-familiar friction with docker, so I gave the nix route a try and got it working, at least to my limited experience so far.

zack-vii:
seems to be a thing then.. we have now way to test it though.. please report back if you are satisfied with the results

fluffynukeit:
I do have some additional updates in the works that will make it more robust.  I also have to put it through the paces a bit more.

zack-vii:
great .. can you this time branch of alpha please. =) i will leave it in for now. feel free to open a new PR with you improvements when you are ready

PR #2365, opened 2021-08-18
Fix: matlab support for 2021a; removed deprecated stuff

### LINKED ISSUE #2357: MATLAB 2021a does not work with MDSplus module
Seems like checks for python code and prevents partial .m files from being loaded. Solution would be to export the python implementation in separate files.

zack-vii:
fixes #2357, replaces #2358


zack-vii:
- tested it without python:
  - mdstest(0) - passes, i.e. returns logical 1
  - mdstest(1) - prints error and switches to java bridge which passes
- tested it with python:
  - mdstest(0) - passes
  - mdstest(1) - passes

all as expected

PR #2367, opened 2021-08-18
Fix: camshr, for case when scsi host is > 9 and scsi linux device > 7

morgak:
Recently got a new dataserver connected to a CAMAC system.  The disk system on this dataserver uses the first 8 SCSI hosts/devices, these changes allow for higher-valued SCSI host numbers to work.

3 small changes:
1 - updated Makefile.in dependencies related to building camshr.
2 - In camshr/QueryHighwayType.c, allowed it to recognize SCSI host IDs with more than 1 digit (in my case, SCSI host for the controller is 14)
3 - In camshr/parse_crate_db.c, switched the strtol base to 10.  Base 0 incorrectly reads the device number (/dev/sgN) when N is greater than 7. (in crate.db it is written as a 3-digit number, the leading 0 becomes a problem).

Thanks,

Kyle Morgan (University of Washington)

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -221,7 +221,7 @@ tests-env:
 
 # Interdependent directories:
 actions: mdsshr tdishr treeshr xmdsshr mdstcpip servershr
-camshr: mdsdcl
+camshr: mdsdcl mdstcpip
 ccl: camshr  mdsshr mdsdcl
 tcl: mdsshr treeshr tdishr mdstcpip servershr mdsdcl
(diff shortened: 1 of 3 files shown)


zack-vii:
looks good, great someone used camshr.. we made some general changes and i was wondering if the driver still works. thanks for fixing the makefile and improving it, ok to test

zack-vii:
test this please

zack-vii (approved):
looks ok

PR #2373, opened 2021-08-25
Fix: mdsmisc GetXYSignalXd() invalid free xd on error

### LINKED ISSUE #2372: mdsip  / gcc-9.3 optimization issue?
Here is my specific problem: When using jScope to connect to a mdsip server compiled with gcc-9.3 (Ubuntu 20.04) with the default optimization flag (-O3), requesting data from an empty node more than once crashes the mdsip process. Not much of an error message given, just:

free(): invalid pointer
Aborted (core dumped)

When I compile with gcc-7.5 (same Ubuntu 20.04 system) or turning gcc-9.3 down to optimization -O1, I no longer see the problem - so it seems like gcc-9.3's optimization is either responsible or finding some subtle bug?  I believe I was able to track the crash down to the free(*vm) call in LibFreeVm inside of mdsshr/librtl.c, but I have no idea what is going on there.  

I have a workaround that is acceptable for my use case (just compile with gcc-7.5), but figured it would be good to mention this.  Additionally I only see this problem when jScope attempts to access the empty node, a Python Connection object to the same mdsip server does not encounter any issues.

Thanks,

Kyle Morgan (University of Washington)

zack-vii:
should fix #2372 :  uninitialized title, xlabel, ylabel on error.

### DIFF
--- a/mdsmisc/ScopeUtilities.c
+++ b/mdsmisc/ScopeUtilities.c
@@ -886,6 +886,9 @@ EXPORT int GetXYSignalXd(mdsdsc_t *const inY, mdsdsc_t *const inX,
     return TdiNULL_PTR;
   EMPTYXD(yXd);
   EMPTYXD(xXd);
+  EMPTYXD(title);
+  EMPTYXD(xLabel);
+  EMPTYXD(yLabel);
   int estimatedSegmentSamples = 0;
   int isLong = FALSE;
   double xmin = -INFINITY, xmax = INFINITY; // requested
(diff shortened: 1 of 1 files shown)


zack-vii:
@GabrieleManduchi: you think this could be tested .. maybe using the python getSignalXY ,, would not help here though as we test with O0 ... unless asan and co.? might be worth adding a test a some point.

GabrieleManduchi:
Why only changing the place where EMPTYXD are defined should solve the issue?

zack-vii:
because this method uses the disencouraged goto statement. the variables are in scope but a previous goto can skip the initialization. as a consequence the pointer may not be null when it should. a better way would be to encapsulate the body into a static function and use return value to indecate the errorstate and ensure the cleanup in the calling function. feel free to improve it.

GabrieleManduchi:
Ok, thanks, for the moment let's proceed in this way

GabrieleManduchi (approved):

PR #2383, opened 2021-09-15
Fix: add list fall-back implementation to support older servers

zack-vii:
jet is using a way old version of mdsplus, this may help reading data via mdsplus-api


joshStillerman:
I will leave this one to the Java types

zack-vii:
also i am not happy with the formater results,,

PR #2394, opened 2021-10-05
Gm fix cpp ctx

GabrieleManduchi:
Handle the new limitation about tree context in multithreaded applications in order to let Tree and TreeNode instances be exchanged among different threads. The Tree contexts are now contained in a Tree private array indexed by thread id. When the Tree instance is used in a new thread, the tree is open again and the new context recorded in the array

Changed C++and Java versions + corresponding tests


zack-vii:
on fc32
looks like a java test fails.. with a critical exception core dump

16:50:27 /git/mdsplus/conf/test-driver: line 120: 21647 Aborted                 (core dumped) "$@" > $log_file 2>&1
16:50:27 FAIL: MDSplus.MdsTreeTest

zack-vii:
http://jenkins2.mdsplus.org/job/fc32/ws/tests/64/java/mdsobjects/tests/MDSplus.MdsTreeTest.log/*view*/

GabrieleManduchi:
retest this please

zack-vii:
since the issue in jenkins memcheck seems to be a some unrelated python cleanup thing. you may review the code and merge it.
 I would suggest you use the ./clang-format before checking it in though.

zack-vii:
unfortunately, i currently dont have more time to review it further than that

AndreaRigoni (approved):
Ok ... nice !

PR #2407, opened 2021-10-30
Jas add compression methods two

joshStillerman:
This PR adds numbered compression methods to node's NCIs.  The spare2 byte in the NCI is now 'COMPRESSION_METHOD'.

0 - original delta compress
1 - gzip compression


_include/_ncidef.h
```diff
@@ -0,0 +1,18 @@
+#ifndef INTERNAL_NCIDEF_H
+#define INTERNAL_NCIDEF_H
+/**************************************
+  _NCIDEF.H - definitions of constants
+  used in  to define compression method 
+  names and routines.
+**************************************/
+
+#define DEFINE_COMPRESSION_METHOD_STRINGS \
+    static const char *compression_methods[] = {"standard", "gzip"};
```
WhoBrokeTheBuild: We should declare both this and the `compression_images` / `compression_routines` arrays with a size, otherwise if you set the compression method index to `2`, it would read out of the bounds of the arrays.

The compression method identifier is an unsigned byte, right? If so, we should make the array 256 in length.

WhoBrokeTheBuild (changes requested):

_include/_ncidef.h
joshStillerman: I agree we should make an array of images.  Since we are checking the index against the length of the array, I see no reason to make the array bigger than it is.

WhoBrokeTheBuild (approved):

Issue #2401, opened 2021-10-21
Inconsistent behavior for empty records


merlea:
I am experiencing some issues using the latest stable version, when using empty arrays (actually of shape `[0,0]`).

I first create the tree with the following TCL script
```
edit test/new
add node data/usage=numeric
write
quit
```

I then test putting some data and retrieving both locally and with mdsip, first with `tdic`:
```
TDI> TreeOpen("test",-1);
TDI> %
TDI> mdsconnect('localhost');
TDI> mdsvalue('setenv("test_path=/home/amerle")');
TDI> mdsopen("test",-1);
TDI> %
TDI> TreePut(":DATA","$1",SET_RANGE(0,0,0));
TDI> Tcl("show data :data");
\TEST::TOP:DATA
 DTYPE_L                       :  Array [ 0,0 ]
TDI> write(*,"local");
local
TDI> shape(:DATA)
[0,0]
TDI> shape(getnci(":DATA","record"))
%TDI Error in GETNCI(":DATA", "record")
%TDI Error in SHAPE(GETNCI(":DATA", "record"))
%TDI Error in EXECUTE('shape(getnci(":DATA","record"))')

TDI> write(*,"remote return value");
remote return value
TDI> shape(mdsvalue(':DATA'))
[]
TDI> shape(mdsvalue('getnci(":DATA","record")'))
[]
TDI> write(*,"remote value");
remote value
TDI> mdsvalue('shape(:DATA)')
[0,0]
TDI> mdsvalue('shape(getnci(":DATA","record"))')
*
TDI> %
TDI> TreePut(":DATA","SET_RANGE(0,0,0)");
TDI> Tcl("show data :data");
\TEST::TOP:DATA
 DTYPE_FUNCTION                : OPC$SET_RANGE
     DTYPE_L                       : 0 
     DTYPE_L                       : 0 
     DTYPE_L                       : 0 
TDI> write(*,"local");
local
TDI> shape(:DATA)
[0,0]
TDI> shape(getnci(":DATA","record"))
[0,0]
TDI> write(*,"remote return value");
remote return value
TDI> shape(mdsvalue(':DATA'))
[]
TDI> shape(mdsvalue('getnci(":DATA","record")'))
[]
TDI> write(*,"remote value");
remote value
TDI> mdsvalue('shape(:DATA)')
[0,0]
TDI> mdsvalue('shape(getnci(":DATA","record"))')
[0,0]
TDI> %
[...]

PR #2408, opened 2021-11-01
Fix: Inconsistent behavior for empty records #2401

merlea:
* For mdsip, accept 0 as dimension when data length is also 0
* For Tdi1GetNci, do not throw error for empty arrays

A more detailed description of the changes is in one of the issue comments.

### DIFF
--- a/mdstcpip/mdsipshr/GetAnswerInfo.c
+++ b/mdstcpip/mdsipshr/GetAnswerInfo.c
@@ -95,7 +95,7 @@ int GetAnswerInfoTO(int id, char *dtype, short *length, char *ndims, int *dims,
       dims[i] =
           i % 2 ? m->h.dims[i / 2] & 0xffffffff : (*m)->h.dims[i / 2] >> 32;
 #else
-      if (m->h.dims[i])
+      if (m->h.dims[i] || !datalen)
         dims[i] = m->h.dims[i];
       else
       {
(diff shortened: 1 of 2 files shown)


joshStillerman (dismissed):

joshStillerman:
test this please

merlea:
I agree and I don't quite understand the behaviour of the Tdi1GetNci function enough to see how to fix it.

By the way even on the `alpha` branch, I always get failing tests where the TDI function `a12__part_name.fun` seems to be missing, but the tests on the docker server are fine. Any clue what might be going wrong?

merlea:
@joshStillerman looking at the logs there is 2 errors:
- when asking for CHILD for a node without childs we now get an empty array instead of the TreeNNF error. The previous behaviour might indeed make more sense as we expect a NID in return.
- when asking for CHILDREN_NIDS in the same situation we also now get an empty array instead of the TreeNNF error. There I think the empty array might be a better answer as the return type is a NID array.
What do you think?

I will provide a new version which only changes the behaviour in the cases I was mentioning in the original error but I think it might be worth considering this CHILDREN_NIDS case (or we can open another issue for that).

merlea:
This new version should pass the tests (I hope...).

joshStillerman:
retest this please

joshStillerman (commented):
Not sure if my comment in the code made it into the review:

line 659 pf GetNci - please add {} to avoid compiler warning
```
if (STATUS_OK && out_ptr->pointer && out_ptr->pointer->class == CLASS_A &&
      ((struct descriptor_a *)(out_ptr->pointer))->arsize == 0)
{
    if (key_ptr->item_code != RECORDIDX)
    {
      status = TreeNNF; // Keep previous behaviour when not asking for RECORD
    }
    else if (out_ptr->pointer->length == 0)
    {
      status = TreeNODATA; // If RECORD is truly empty, return TreeNODATA error
    // If RECORD is not empty but has zero size, return it
    }
}
```

joshStillerman:
retest this please

joshStillerman (approved):

PR #2411, opened 2021-11-03
Gm fix marte2

GabrieleManduchi:
Removed Mode=ExpectsReply from generated cfg that potentially introduced deadlocks


AndreaRigoni (approved):
Ok

PR #2415, opened 2021-11-08
Gm fix resample

GabrieleManduchi:
In XTreeDefaultResampleMode, when evaluating timebase in order to perform resampling, TdiData() is called. In this case the whole timebase must be read, not a resampled one, even if the timebase expressions refers to segmented nodes.
To make sure this happens, the current time context is reset to null start, end, delta before TdiData() and then reset to its original values after calling TdiData()


AndreaRigoni (approved):
Great Gabriele !

PR #2418, opened 2021-11-12
Fix: can not get/set current shot over thick

joshStillerman:
The code to access the current shot id for thick client was checking
that the connection ID > 0.  0 is a valid (1st) connection ID>


joshStillerman:
retest this please

joshStillerman:
retest please

zack-vii (approved):
that is correct

mwinkel-dev:
The title of this PR mentions "get/set" (i.e., `show/set current`).   But it only fixed `show current` for thick-client.

The `set current` fix is being added with PR #2723.

PR #2420, opened 2021-11-15
Fix: #2419: fixed monitor queue

### LINKED ISSUE #2419: MdsServerShr: monitor queue needs do-over
the SendMonitor worker with its conditional variables should be re-implemented; 
currently it bears a race-condition because the mutex to the send_monitor_queue_mutex is not the mutex of the conditional variable. 

in this situation the usage of Condition is contra productive. and possibly the 
* send_monitor_wait sould be inlined (lock is already held, the condition variable is checked here, would reduce to a pthread_cond_timedWait() )
* send_monitor_signal would simply spell out the macro using the right mutex and a pthread_cond_t instead of a struct Condition.

i possibly on a long run would be beneficial to improve the P2W*.hpp headers and add similar ones for stdc.

zack-vii:
should fix helgrind issue in MdsServerShr

### DIFF
--- a/servershr/ServerDispatchPhase.c
+++ b/servershr/ServerDispatchPhase.c
@@ -137,9 +137,9 @@ static inline int is_abort_in_progress()
 }
 
 static Condition JobWaitC = CONDITION_INITIALIZER;
-static Condition SendMonitorC = CONDITION_INITIALIZER;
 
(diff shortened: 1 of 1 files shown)


zack-vii:
retest this please

PR #2425, opened 2021-12-01
Fix: In a pydevice, move tree copy from init() to run() in the MDSWorker thread

santorofer:
In order to resolve issues related to using distributed client (see #2360) when using MDSplus python devices (specifically, for acq2106_423st.py and acq2106_435st.py):
- we make a copy of the tree object in the MDSWork thread, not in the __init__() but in the run() of that thread.


zack-vii:
in fact the dev.copy() may remain in \_\_init\_\_ but it should not be used/accessed outside of the thread. so the caching of the nodes in \_\_init\_\_ caused the tree to be opened in the main thread. making a copy in the thread may do the trick as copy() uses infos cached in python fields (.tree and .shot). this should not be alarming for other devices that do the copy during \_\_init\_\_() as long as they dont access it outside of their thread method.

zack-vii (approved):

PR #2427, opened 2021-12-06
Gm fix redpitaya

GabrieleManduchi:
Avoid blocking burst recording in the possible case  samples from previous burst are pending in queue without interrupt (no threshold reached)


zack-vii:
maybe this branch should be rebased to disentangle all the merge commits.. btw you can use rebase as your default merge strategie

PR #2433, opened 2021-12-22
Switch Ubuntu18 to use python3-numpy

WhoBrokeTheBuild:
Currently, Ubuntu18 installs python-numpy, which in turn installs python2
I would like to switch it to python3-numpy, which will not install python2

@zack-vii I would like your input on this one


zack-vii:
i support it. i think python3 is installed by default anyway sonit makes sense to use the python3 version of numpy. to set this for ubuntu-20 we might need to add a new filter similar to dist to the packaging... not sure.

zack-vii (approved):
oh i see this was the pr already.. great =)

PR #2446, opened 2022-02-07
Feature: CompileTree will exit with non-zero status code for error messages.

Louwrensth:
This would make `CompileTree` a bit more POSIX compatible and scripting based on exit status would become possible (instead of reading (locale-dependent) error messages, which now also go to standard error as they should). 

N.B. Some error messages are only messages themselves and have no `System.exit(1)` statement. Is this by choice? I left it this way.

Hope the intention is appreciated and applied through out mdsplus, not just `CompileTree`.


GabrieleManduchi (dismissed):

GabrieleManduchi:
retest this please

GabrieleManduchi:
Apparently there is no space left on device in the build&test system. Josh, could you have a look at it?

joshStillerman:
sure

GabrieleManduchi:
retest this please

joshStillerman:
retest this please

zack-vii:
retest this please

mwinkel-dev:
I also concur that the proposed changes are good.

Note though that MDSplus uses two different schemes for error codes.  (Of course, error codes returned by functions are not the same thing as an exit code returned by a terminating process.   But they are often related.)

- POSIX is used in some modules (such as the Java code in this PR).

- VAX/VMS is used by a lot of the functions in the core MDSplus libraries.   It has many non-zero error codes that denote success.   For more details, refer to the `deploy/gen_messages.py` file around line 280.
[description of MDSplus error codes](https://github.com/MDSplus/mdsplus/blob/8939e2e195959141091bf46f9e1ce48a8bd67675/deploy/gen_messages.py#L281)

These two schemes treat an error code of `1` quite differently.   

- In POSIX, `1` is a failure (i.e., success is a `0`).
- In VAX/VMS, `1` is a success (as is any error code with the low order bit set).

mwinkel-dev:
There are two `CompileTree.java` files in the entire source code.   This PR only changes the `jtraverser` version.   Should a similar change be made to the `jtraverser2` version?
https://github.com/MDSplus/mdsplus/blob/alpha/java/jtraverser2/src/main/java/mds/jtraverser/tools/CompileTree.java

If so, should that change be added to this PR?   Or done as a separate PR?

mwinkel-dev (approved):
As per discussion with @WhoBrokeTheBuild, we will handle the `jtraverser2` file with a  new PR.   Therefore approving this PR for the `jtraverser` file.

PR #2461, opened 2022-04-29
Feature:Add MARTe2 execution without building configuration

GabrieleManduchi:
- Added MARTE2_SUPERVISOR method startMarteStraight that does not breate MARTe2 configuration file
- Fixed offsets in MARTE2_NI6368 device setup


GabrieleManduchi:
It is a very old pull request and it can be safely closed

PR #2463, opened 2022-05-26
Fix: use size of packed NCI instead of 42

joshStillerman:
The code was using the known size of the packed NCI structure (42)
instead of asking the compiler to fill this in for us.

Replace the explicit byte offset memory copies in SerializeNCI with
structure references.

### DIFF
--- a/treeshr/TreeSerializeNci.c
+++ b/treeshr/TreeSerializeNci.c
@@ -29,65 +29,65 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 void TreeSerializeNciOut(const NCI *in, char *out)
 {
-  char *ptr = out;
-  memset(out, 0, 42);
-  putint32(&ptr, &in->flags);
-  putint8(&ptr, &in->flags2);
-  putint8(&ptr, &in->spare);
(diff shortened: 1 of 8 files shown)


treeshr/TreeSerializeNci.c
```diff
@@ -29,65 +29,65 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 void TreeSerializeNciOut(const NCI *in, char *out)
 {
-  char *ptr = out;
```
zack-vii, line 32: I like the intention but the change does not handle endianess (handled by put/getint32/64)

zack-vii (changes requested):

treeshr/TreeSerializeNci.c
```diff
@@ -29,65 +29,65 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
...
-  memset(out, 0, 42);
-  putint32(&ptr, &in->flags);
```
zack-vii, line 34: I like the intention but the change does not handle endianess (handled by put/getint32/64)

zack-vii (changes requested):

Issue #2467, opened 2022-06-28
Incorrect time handling in MdsShr->LibTimeToVMSTime


merlea:
Looking at the code of `LibTimeToVMSTime` there are 2 separate calls to get the seconds (via `time(NULL)`) and microseconds (via `gettimeofday(tv,NULL)`) part of the current time. When the current time is close to a round second (i.e. `tv.tv_usec` is close to 0), this can result in the VMS time to be off by close to a second.

A particularly sensitive consequence of this is the "TIME_INSERTED" NCI can be off by 1 s which makes it hard to rely on it to check for causality problems.

The following program illustrates the problem:
```
#include <stdio.h>
#include <sys/time.h>
#include <stdint.h>
#include <unistd.h>

int LibTimeToVMSTime(const time_t * time_in, int64_t * time_out);

int main()
{
  int i;
  int64_t ans;
  int64_t VMS10seconds=100000000;
  LibTimeToVMSTime(0,&ans);
  printf("[initial] %08ld\n",ans%VMS10seconds);
  int64_t waitt;
  waitt = (9990000-ans%10000000)/10;
  if (waitt<0) waitt += 1000000;
  printf("\tWaiting %08ldus\n",waitt);
  usleep(waitt);
  LibTimeToVMSTime(0,&ans);
  printf("[updated] %08ld\n",ans%VMS10seconds);
  for (i=0;i<100;i++)
  {
    LibTimeToVMSTime(0,&ans);
    printf("[%04d] %08ld\n",i,ans%VMS10seconds);
    usleep(10);
  }
}
```
with sample output:
```
[initial] 89265350
	Waiting 00072465us
[updated] 89991350
[00] 89991390
[01] 89992040
[02] 89992680
[03] 89993260
[04] 89993900
[05] 89994530
[06] 89995150
[07] 89995780
[08] 89996410
[09] 89997040
[10] 89997670
[11] 89998300
[12] 89998930
[13] 89999560
[14] 80000190  <--- This is not OK
[15] 80000820
[16] 80001450
[...]
[62] 80031670
[63] 80032300
[64] 80032930
[65] 90033420  <--- After a while it's OK again
[66] 90034160
[67] 90034790
[...]
```

PR #2468, opened 2022-06-28
Fix: MdsShr->LibTimeToVMSTime: Avoid separate calls for current seconds and microseconds (#2467)

merlea:
See #2467.

In MdsShr->LibTimeToVMSTime, if `time_in` is not provided in input, then the current time is fetched using a single call to `gettimeofday`.

### DIFF
--- a/mdsshr/librtl.c
+++ b/mdsshr/librtl.c
@@ -1133,14 +1133,21 @@ EXPORT int LibConvertDateString(const char *asc_time, int64_t *const qtime)
 EXPORT int LibTimeToVMSTime(const time_t *const time_in,
                             int64_t *const time_out)
 {
-  time_t time_to_use = time_in ? *time_in : time(NULL);
   struct timeval tv;
   if (time_in)
+  {
+    tv.tv_sec  = *time_in;
     tv.tv_usec = 0;
+  }
   else
     gettimeofday(&tv, 0);
+#ifdef _WIN32
+  /* MinGW defines timeval as (long,long) so an explicit cast is needed */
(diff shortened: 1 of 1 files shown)


joshStillerman:
I am reviewing this now.  Seems odd to me that the input to this routine is a time_t but I guess that is the way we wrote it, or I am confused about what a time_t is. Do you have time to discuss?

joshStillerman:
retest this please

merlea:
I am not an expert at all about compilation on Windows and I don't know what is your build setup but my guess is that since `gettimeofday` is a POSIX extension, it might have to be added in an ad-hoc way on Windows systems.

Otherwise as the `tv` argument is a `timeval` and its `tv_sec` member is declared as a `time_t`, there should be no problem when passing that to `get_tz_offset`. So I suppose on your Windows build `timeval.tv_sec` is not declared as a `time_t`.

I then see 2 options:
- we can introduce an intermediate variable which casts explicitely the `tv.tv_sec` value to `time_t` and pass its address to `get_tz_offset` (this is quite easy but does feel like a workaround only).
- Fix the definitions of `timeval` and `gettimeofday` in the Windows build (this feels like a better option but I don't know how to do it).

merlea:
I have looked further and it seems the definition of `timeval` and `gettimeofday` come from mingw directly. And according to [this ticket](https://sourceforge.net/p/mingw-w64/bugs/720/), they will not change it. So we must accept and take into account that for Windows, `timeval` is `[long,long]` and not `[time_t,suseconds_t]`.

PR #2471, opened 2022-07-01
Stephen fixed jTraverser2 so long node names now display correctly

mwinkel-cfs:
Stephen fixed a TDI bug so it properly handles long node names.   That in turn fixed display of long node names in jTraverser2 (i.e., eliminated trailing garbage characters).   Mark ran Stephen's changes on an Ubuntu 22 VM running the old-trees branch, namely the same system where the bug was found.   After the debugging session, Mark removed debug printf() statements and confirmed that jTraverser2 still works fine (as do mdstcl, jTraverser and jScope).  Mark also tested jTraverser2 with edge cases that used the longest possible node name, and all worked A-OK.  Thus this fix is now ready for inclusion on the old-trees branch.

### DIFF
--- a/treeshr/TreeGetNci.c
+++ b/treeshr/TreeGetNci.c
@@ -503,7 +503,6 @@ int TreeGetNci(int nid_in, struct nci_itm *nci_itm)
         string = strncpy(malloc(sizeof(NODE_NAME) + 1), node->name,
                          sizeof(NODE_NAME));
         int length = minInt(strlen(node->name), sizeof(NODE_NAME));
-        string[length] = '\0';
         memset(string + length, '\0', sizeof(NODE_NAME) - length + 1);
       }
       else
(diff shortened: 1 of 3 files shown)


treeshr/TreeGetNci.c
```diff
@@ -502,7 +502,9 @@ int TreeGetNci(int nid_in, struct nci_itm *nci_itm)
       {
         string = strncpy(malloc(sizeof(NODE_NAME) + 1), node->name,
                          sizeof(NODE_NAME));
-        string[sizeof(NODE_NAME)] = '\0';
+        int length = minInt(strlen(node->name), sizeof(NODE_NAME));
+        string[length] = '\0';
```
zack-vii: is this not also part of the next line

WhoBrokeTheBuild: Yes it is, we solved this twice and forgot to delete one of them, good catch

mwinkel-cfs: Stephen and I chose the memset() option as it is the more conservative approach (i.e., clears the remainder of the node name memory).

Testing shows that the fix still works fine after removing the extraneous line.   This screenshot shows a new tree displayed in jTraverser2.   Nodes that start with X were added using TCL.   All other nodes were added using jTraverser2.  The tree was written and closed, plus then exited TCL and jTraverser2.   Then re-opened the tree using jTraverser2 and took this screenshot.

pr2471

PR #2475, opened 2022-07-07
Add a fixed-length string specifier to the printf for buildtag()

WhoBrokeTheBuild:
There was an error generated if your branch was 12 or more characters. The string would be truncated by git_revision.sh, but it would throw an error about printing a "non-null-terminated string".


zack-vii:
retest this please

zack-vii (approved):
That is true. the VERSIONCONST struct will be read by tdi, java, and python and must have a fixed length for it to work without leaks. we wont have longer branch names in releases anyway ans stable and alpha fit nicely ;)

PR #2476, opened 2022-07-07
Add DbiTREE_VERSION, add version to `show db`

WhoBrokeTheBuild:
Add DbiTREE_VERSION, which returns `db->info->header->version`
Add the tree file format version number to the output of `mdstcl show db`

Example:
```
TCL> set tree oldtree
TCL> set tree newtree
TCL> show db
000  NEWTREE       shot: -1 [\NEWTREE::TOP]    version: 2
001  OLDTREE       shot: -1 [\OLDTREE::TOP]    version: 1
```


zack-vii (approved):
i assume cleanup for the unrelated stuff.. adding a test may be good.. maybe in tdi test .. also python should be extended

PR #2477, opened 2022-07-08
In the 423/435 stream devices, dev.copy() was moved back to the main process.

santorofer:
The 423/435 streaming devices contain a call to copy the tree (`dev.copy()`) inside the MDSWorker's new thread. 
This breaks the code when executing it from an mdsip service (i.e. when dispatching the INIT action).
Moving that statement back to the main process thread fixes the issue.

### DIFF
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -167,7 +167,10 @@ class _ACQ2106_423ST(MDSplus.Device):
         def __init__(self, dev):
             super(_ACQ2106_423ST.MDSWorker, self).__init__(name=dev.path)
 
-            self.dev = dev.copy()
+            # Tuple designed to bring a copy of the tree to the MDSWorker thread
+            self.info = (dev.tree.name, dev.tree.shot, dev.path)
+
+            self.dev = dev
 
             self.nchans = self.dev.sites * self.dev.NUM_CHANS_PER_SITE
 
@@ -192,6 +195,9 @@ class _ACQ2106_423ST(MDSplus.Device):
                 for e in arr:
                     ans = lcm(ans, e)
                 return int(ans)
+
+            tree = MDSplus.Tree(self.info[0], self.info[1])
+            self.dev = tree.getNode(self.info[2])
 
             if self.dev.debug:
                 print("MDSWorker running")
--- a/pydevices/HtsDevices/acq2106_423st.py
+++ b/pydevices/HtsDevices/acq2106_423st.py
@@ -167,8 +167,10 @@ class _ACQ2106_423ST(MDSplus.Device):
         def __init__(self, dev):
             super(_ACQ2106_423ST.MDSWorker, self).__init__(name=dev.path)
 
-            # Tuple designed to bring a copy of the tree to the MDSWorker thread
-            self.info = (dev.tree.name, dev.tree.shot, dev.path)
+            # Variables designed to bring a copy of the tree to the MDSWorker thread
+            self.tree = dev.tree.name
+            self.shot = dev.tree.shot
+            self.path = dev.path
 
             self.dev = dev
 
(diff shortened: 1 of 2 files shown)


pydevices/HtsDevices/acq2106_435st.py
```diff
@@ -188,9 +188,9 @@ class MDSWorker(threading.Thread):
         def __init__(self, dev):
             super(_ACQ2106_435ST.MDSWorker, self).__init__(name=dev.path)
 
-            self.dev = dev
+            self.dev = dev.copy()
 
-            self.nchans     = self.dev.sites*32
+            self.nchans     = self.dev.sites * 32
```
WhoBrokeTheBuild: Why 32? Can we put that in a constant or add a comment?

pydevices/HtsDevices/acq2106_435st.py
```diff
@@ -469,12 +466,12 @@ def init(self, resampling = False, armed_by_transient = False):
         coeffs = uut.cal_eslo[1:]
         eoff = uut.cal_eoff[1:]
 
-        self.chans = []
-        nchans = uut.nchan()
+        chans = []
+        nchans = self.sites * 32
```
WhoBrokeTheBuild: Same question as above, maybe do a static constant in the class

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_423st.py
```diff
@@ -159,15 +159,17 @@ class _ACQ2106_423ST(MDSplus.Device):
...
         NUM_BUFFERS = 20
 
         def __init__(self, dev):
             super(_ACQ2106_423ST.MDSWorker, self).__init__(name=dev.path)
 
-            self.dev = dev
+            self.dev = dev.copy()
 
-            self.nchans = self.dev.sites*32
+            self.nchans = self.dev.sites * NUM_CHANS_PER_SITE
```
WhoBrokeTheBuild: You still need to use `self.NUM_CHANS_PER_SITE`

pydevices/HtsDevices/acq2106_435st.py
```diff
@@ -182,15 +182,17 @@ class _ACQ2106_435ST(MDSplus.Device):
...
         NUM_BUFFERS = 20
 
         def __init__(self, dev):
             super(_ACQ2106_435ST.MDSWorker, self).__init__(name=dev.path)
 
-            self.dev = dev
+            self.dev = dev.copy()
 
-            self.nchans     = self.dev.sites*32
+            self.nchans     = self.dev.sites * NUM_CHANS_PER_SITE
```
WhoBrokeTheBuild: You will need to use `self.dev.NUM_CHANS_PER_SITE`

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_423st.py
```diff
@@ -159,15 +159,20 @@ class _ACQ2106_423ST(MDSplus.Device):
...
+    NUM_CHANS_PER_SITE = 32
+
     class MDSWorker(threading.Thread):
         NUM_BUFFERS = 20
 
         def __init__(self, dev):
             super(_ACQ2106_423ST.MDSWorker, self).__init__(name=dev.path)
 
+            # Tuple designed to bring a copy of the tree to the MDSWorker thread
+            self.info = (dev.tree.name, dev.tree.shot, dev.path)
```
WhoBrokeTheBuild: Originally this was made to emulate the to_info and from_info functions that I made for Josh, but we've replaced those with just multiple variables, so I think we should do the same here.
```
self.tree = dev.tree.name
self.shot = dev.tree.shot
self.path = dev.path

...

tree = MDSplus.Tree(self.tree, self.shot)
self.dev = tree.getNode(self.path)
```

WhoBrokeTheBuild (changes requested):

pydevices/HtsDevices/acq2106_423st.py
santorofer: Ok, this looks good to me too. I'll make those changes then.

WhoBrokeTheBuild (dismissed):

WhoBrokeTheBuild (approved):

PR #2484, opened 2022-08-11
Improved the setting of gains and offsets

santorofer:
The setting of gains and offsets are now improved.
The function that sets the gains also commits them using an EPICS call as well.

### DIFF
--- a/pydevices/HtsDevices/acq2106_435sc.py
+++ b/pydevices/HtsDevices/acq2106_435sc.py
@@ -137,18 +137,16 @@ class _ACQ2106_435SC(acq2106_435st._ACQ2106_435ST):
         if input_path:
             input_node = self.__getattr__(input_path)
 
-        # For versions of the firmware v498 or greater: 
-        # coefficients and offsets takes into account the gains and offsets of the Signal Conditioning.
+        # For versions of the firmware v498 or greater in all the ACQ SC32 the calibration coefficients and offsets 
+        # already take into account the gains and offsets of the signal conditioning stages.
         
         # = (coefficient * value) + offset
         target_node.setSegmentScale(
             MDSplus.ADD(
-
                 MDSplus.MULTIPLY(
                     input_node.COEFFICIENT, 
                     MDSplus.dVALUE()
                 ), 
-
                 input_node.OFFSET
             )
         )
--- a/pydevices/HtsDevices/acq2106_435sc.py
+++ b/pydevices/HtsDevices/acq2106_435sc.py
@@ -65,14 +65,6 @@ class _ACQ2106_435SC(acq2106_435st._ACQ2106_435ST):
                 "FREQ must be 10000, 20000, 40000, 80000 or 128000; not %d" % (freq,))
 
         for card in self.slots:
-            self.resetGainsOffsetsDefaults(card)
-            self.slots[card].SC32_GAIN_COMMIT = 1
-            if self.debug:
-                print("GAINs Resetted to Default value (1) and Committed for site %s" % (card,))
-
-        time.sleep(1)
-
-        for card in self.slots:
             self.setGainsOffsets(card)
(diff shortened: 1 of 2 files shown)


pydevices/HtsDevices/acq2106_435sc.py
```diff
@@ -46,6 +46,8 @@ class _ACQ2106_435SC(acq2106_435st._ACQ2106_435ST):
     ]
 
     def init(self):
+        import time
```
WhoBrokeTheBuild: We can remove this

pydevices/HtsDevices/acq2106_435sc.py
```diff
@@ -81,80 +82,83 @@ def getUUT(self):
         uut = acq400_hapi.Acq2106(self.node.data(), monitor=False, has_wr=True)
         return uut
 
+    def computeGains(self, g):
+        g1opts = [ 1000, 100, 10, 1 ]
+        g2opts = [ 1, 2, 5, 10 ]
 
-    def setGainsOffsets(self, card):
+        for g1opt in g1opts:
```
WhoBrokeTheBuild, line 88: We should add a TODO to this:
I want to support explicitly setting both gains in the form of
* single number -> gain auto configuration
* two numbers separated by a comma -> set gain1 and gain2 explicitly

pydevices/HtsDevices/acq2106_435sc.py
```diff
@@ -81,80 +82,83 @@ def getUUT(self):
...
+                    raise MDSplus.DevBAD_PARAMETER(
+                            "SC_GAIN must be computable from (one of: 1000, 100, 10, 1) * (one of: 1, 2, 5, 10) ; not %d" % (g,))
+                return (g1, g2)
+
+    def setGainsOffsets(self, site):
         import epics
         import socket
 
+        if site not in [1, 3, 5]:
+            # (slw) TODO: Improve
```
WhoBrokeTheBuild, line 102: I don't want to throw an exception / error here, because this could be called for other sites and I think we just want it to silently continue... hmm

pydevices/HtsDevices/acq2106_435sc.py
```diff
@@ -81,80 +82,83 @@ def getUUT(self):
...
+            print('site is not 1, 3, or 5')
+            return
+
         domainName = socket.gethostbyaddr(str(self.node.data()))[0]
```
WhoBrokeTheBuild, line 107: We should add a note about needing to replace this with an EPICS_NAME node or somesuch

pydevices/HtsDevices/acq2106_435st.py
```diff
@@ -240,8 +240,8 @@ def lcma(arr):
                 if nacc_str == '0,0,0':
                     nacc_sample = 1
                 else:
-                    nacc_tuple = ast.literal_eval(nacc_str)
-                    nacc_sample = nacc_tuple[0]
+                    nacc_sample = int(nacc_str.split(",")[0])
```
WhoBrokeTheBuild, line 243: We might want better error checking around this, in case the value isn't what we expect

WhoBrokeTheBuild (changes requested):

joshStillerman:
retest this please

joshStillerman:
retest this please

joshStillerman:
retest this please

pydevices/HtsDevices/acq2106_435sc.py
santorofer: This has been resolved in the Slowmon PR: a new tree node was created to input the EPICS name.

pydevices/HtsDevices/acq2106_435st.py
santorofer: The error checking is done in the INIT method: there are checks to be sure that the values are in a range.

WhoBrokeTheBuild (approved):

PR #2498, opened 2022-11-01
Fix XNCI writes corrupting node data

WhoBrokeTheBuild:
The following code is likely supposed to do something useful, but currently it clobbers the data in a node when you write data into an XNCI.

Steps to reproduce with current alpha:
```py
import MDSplus
tree = MDSplus.Tree('tree', 0)
tree.NODE_NAME.record = 1234
tree.setExtendedAttribute('hello', 'world')
tree.NODE_NAME.record # throws data corrupted exception
```

@zack-vii This code was part of one of your original commits to add / restructure XNCIs, I would love for you to look at this before I remove these lines.

### DIFF
--- a/treeshr/TreeSegments.c
+++ b/treeshr/TreeSegments.c
@@ -311,12 +311,6 @@ inline static void begin_extended_nci(vars_t *vars)
   {
     vars->attr_offset =
         RfaToSeek(vars->local_nci.DATA_INFO.DATA_LOCATION.rfa);
-    if (vars->attr.facility_offset[STANDARD_RECORD_FACILITY] != -1)
-    {
-      vars->attr.facility_offset[STANDARD_RECORD_FACILITY] = -1;
-      vars->attr.facility_length[STANDARD_RECORD_FACILITY] = 0;
-      vars->attr_update = 1;
-    }
   }
   else
   {
@@ -565,11 +559,13 @@ static int set_xnci(vars_t *vars, mdsdsc_t *value, int is_offset)
   {
     value_offset = vars->xnci_header_offset;
     value_length = -1;
+    printf("is_offset value_offset=%ld\n", value_offset);
   }
   else
   {
     if (value)
     { // NULL means delete
+      printf("calling tree_put_dsc()\n");
       RETURN_IF_NOT_OK(tree_put_dsc(
           vars->dblist, vars->tinfo, *(int *)vars->nid_ptr, (mdsdsc_t *)value,
           &value_offset, &value_length, vars->compress));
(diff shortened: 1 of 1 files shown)


joshStillerman:
How are all these in 1 PR?  A bad merge?Sent from my iPhoneOn Oct 4, 2023, at 11:45 AM, Stephen Lane-Walsh ***@***.***> wrote:﻿
@zack-vii Can you please look at this?

mwinkel-dev:
The lines suggested for removal date back to the very early history of the `TreeSegments.c` file.  The third commit made to the file (hash `bd30483` on 29-Jan-2007) contains similar lines.  (Of course the code segment has been refactored a few times over the years.)

[bd30483 excerpt](https://github.com/MDSplus/mdsplus/blob/bd304834b3c710835dbeae4ac4f76bb09f665915/treeshr/TreeSegments.c#L104C1-L116C6)

Given this long history, there are three possibilities to consider:

1. perhaps this is a latent bug that has been present from 2007, or
2. maybe these lines were needed for a feature that is no longer used, or
3. perhaps these lines are still needed in some scenarios.

It will take time to determine which of those possibilities is true.   Thus, the review of this PR will take some time.

zack-vii:
in branch zck-xnci I added _TreeSetXnci operations trying to break segmented nodes but that seems to work fine.

https://github.com/MDSplus/mdsplus/blob/7bae51551c1a7316bf642361267eb08091a86a96/treeshr/testing/TreeSegmentTest.c#L191

Writing a record over a node with existing xnic preserves that xnci. However, writing an xnci after writing a normal record breaks that record.
https://github.com/MDSplus/mdsplus/blob/7bae51551c1a7316bf642361267eb08091a86a96/treeshr/testing/TreeSegmentTest.c#L209
https://github.com/MDSplus/mdsplus/blob/7bae51551c1a7316bf642361267eb08091a86a96/treeshr/testing/TreeSegmentTest.c#L211

I dont think we should remove those lines as they seem to make sense. 
It may be that TreeGetRecord falsely assumes it must be a segmented node when it detects any attribute. I this case the issue was introduced when we implemented the feature of extendedAttributes to the user.  User attributes are not associated with the data of a node and may be unexpected when reading a node.

zack-vii (changes requested):
As explained in the comments. I think the issue is rather in TreeGetRecord jumping to wrong conclusions in the existence of extended attributes.

zack-vii:
I revoke my concern.. it is related to the writing, however we cannot omit clearing the standard record if we are writing segments. i cleaned it up a little. it clears the record now only when writing segments or when updating xncis. This lead no into an alternative solution  #2785 and a few other PRs that fix unrelated issues that i spotted during my inspection  #2782, #2783,  #2784.
I assume leaving the reference intact will keep accessible data even after tree cleanup.

mwinkel-dev:
Hi @zack-vii -- Thanks for looking at this XNCI issue and for the additional PRs.

Stephen asked me to review this PR #2498 a while ago.   I will now compare it to your PRs (and thereby learn a lot more about XNCI stuff).   It will likely take a few days before we have all of the PRs reviewed and approved.

Thanks again for the help with this XNCI issue!

PR #2500, opened 2022-11-03
Fix: Use _Exit() instead of exit() from forked process in tunnel_rout…

merlea:
…ines->io_disconnect

This commit avoids that the forked process designed to kill the connection subprocess (such as ssh) calls exit(0) which will call registered functions using `atexit` or `onexit` (especially ones not related to the MDSplus libraries) but _exit(0) instead.

In our particular use case (a MATLAB MEX-file), the exit(0) function would actually fail and the process would live forever.

### DIFF
--- a/mdstcpip/io_routines/IoRoutinesTunnel.c
+++ b/mdstcpip/io_routines/IoRoutinesTunnel.c
@@ -59,7 +59,8 @@ static int io_disconnect(Connection *c)
         usleep(100000);
       if (!i)
         kill(p_pid, SIGTERM);
-      exit(0);
+      /* Do not call any functions registered with atexit or onexit */
+      _Exit(0);
     }
     if (pid < 0)
       kill(p->pid, SIGTERM);


zack-vii:
you would want t to cleanup when possible though, you should rather do both try to `exit(0)` and `abort()` only if it fails.

merlea:
I am not sure I understand what there is to cleanup. The main process forks itself to kill the subprocess (for example ssh). I think if there was any task that was needed to cleanup when the main process exits, I wouldn't want to do them when we are only disconnecting an MDSip ssh/pipe connection. But maybe I am missing something.

zack-vii:
Well if there would not be anything to clean, the mex file would not hang .. if you deliberately kill the process every time, you may increase the chance for file locks to remain in place. in particular file locks on the server that client holds on the connected server would prevent other clients from accessing data. with this dirty fix you may find yourself wondering one day why your writing routines are slow or even randomly hang. This is even worse if the server is a windows machine. Usually there is a good reason why cleanup methods have been pushed to the on-exit stack.

merlea:
I feel like we're not understanding each other.

The main MDSplus process creates a fork with the sole purpose of killing the SSH connection, once this is done, this forked process has no purpose and must be killed in some way, but the main MDSplus process still lives on and will perform its cleaning actions in due time.

merlea:
In an attempt to clarify my point I made the following diagram which I think summarizes the current flow of actions when opening and closing an MDSip connection with SSH (or any tunnel-like protocol):
```
             (PID: 1, main)
                   |
    -------------------------------
    |           Step 1            |
    -------------------------------
    | Different actions,might     |
    | involve registering exit    |
    | handlers                    |
    -------------------------------
                   |
                  ...
                   |
    -------------------------------
    |           Step 2            |
    | Open MDSplus SSH connection |
    -------------------------------
    | call io_connect             |
    | |_ call fork                |---------------------------------------------------------
    -------------------------------                                                        |
                   |                                                                 (PID: 2, main)
                   |                                                                       |
                   |                                                        -------------------------------
                   |                                                        |           Step 2.1          |
                   |                                                        -------------------------------
                   |                                                        | exec mdsplus-client-ssh     |
                   |                                                        -------------------------------
                   |                                                                       |
                   |                                                                 (PID: 2, ssh)
                   |                                                                       |
                  ...                                                                     ...
                   |                                                                       |
    -------------------------------                                                        |
    |           Step 3            |                                                        |
    | Close MDSplus SSH connection|                                                        |
    -------------------------------                                                        |
    | call io_disconnect          |                                                        |
    | |_ call fork                |---------------------                                   |
    | |_ close pipes              |                    |                                   |
    ------------------------------               (PID: 3, main)                            |
                   |                                   |                                   |
                   |                    -------------------------------                    |
                   |                    |           Step 3.1          |                    |
                   |                    | make sure PID 2 dies        |                    |
                   |                    -------------------------------                    |
                   |                    | call waitpid(2,...)         |                    |
                   |                    | call kill(2,SIGTERM)        |                    |
                   |                    -------------------------------                    |
                   |                                   |                                   x
                   |                    -------------------------------
                   |                    |           Step 3.2          |
                   |                    |     Forked process exit     |
                   |                    -------------------------------
                   |                    | call exit                   |
                   |                    | |_ call exit handlers       |
                   |                    -------------------------------
                   |                                   |
                   |                                   x
                  ...
                   |
    -------------------------------
    |           Step 4            |
    |      Main program exit      |
    -------------------------------
    | call exit                   |
    | |_ call exit handlers       |
    -------------------------------
                   |
                   x
```
My main problem is that the exit handlers that will be called at step 3.2 are the same that will be called at step 4 (if none are registered in between of course ...) and so if all they do is clean up the process memory, that's fine although unnecessary as the forked process (PID 3) is about to die, but if they have some external actions as file operations or communications with other processes etc., then this has the potential to be harmful to the main process (PID 1) either during its continued running or when it finally exits.

So I would argue that we need to skip these exit handlers altogether and call `_exit` instead which gives the following modified flow (only step 3.2 has been modified):
```
             (PID: 1, main)
                   |
    -------------------------------
    |           Step 1            |
    -------------------------------
    | Different actions,might     |
    | involve registering exit    |
    | handlers                    |
    -------------------------------
                   |
                  ...
                   |
    -------------------------------
    |           Step 2            |
    | Open MDSplus SSH connection |
    -------------------------------
    | call io_connect             |
    | |_ call fork                |---------------------------------------------------------
    -------------------------------                                                        |
                   |                                                                 (PID: 2, main)
                   |                                                                       |
                   |                                                        -------------------------------
                   |                                                        |           Step 2.1          |
                   |                                                        -------------------------------
                   |                                                        | exec mdsplus-client-ssh     |
                   |                                                        -------------------------------
                   |                                                                       |
                   |                                                                 (PID: 2, ssh)
                   |                                                                       |
                  ...                                                                     ...
                   |                                                                       |
    -------------------------------                                                        |
    |           Step 3            |                                                        |
    | Close MDSplus SSH connection|                                                        |
    -------------------------------                                                        |
    | call io_disconnect          |                                                        |
    | |_ call fork                |---------------------                                   |
    | |_ close pipes              |                    |                                   |
    ------------------------------               (PID: 3, main)                            |
                   |                                   |                                   |
                   |                    -------------------------------                    |
                   |                    |           Step 3.1          |                    |
                   |                    | make sure PID 2 dies        |                    |
                   |                    -------------------------------                    |
                   |                    | call waitpid(2,...)         |                    |
                   |                    | call kill(2,SIGTERM)        |                    |
                   |                    -------------------------------                    |
                   |                                   |                                   x
                   |                    -------------------------------
                   |                    |           Step 3.2          |
                   |                    |     Forked process exit     |
                   |                    -------------------------------
                   |                    | call _exit                  |
                   |                    |                             |
                   |                    -------------------------------
                   |                                   |
                   |                                   x
                  ...
                   |
    -------------------------------
    |           Step 4            |
    |      Main program exit      |
    -------------------------------
    | call exit                   |
    | |_ call exit handlers       |
    -------------------------------
                   |
                   x
```
As you can see the exit handlers will still be called in due time and any cleaning operation necessary will be performed.

Let me know if you agree with this approach.

merlea:
Would it be possible to get a new review for this. I think this is an important issue and I made an extra effort to explain what has been changed and why.

joshStillerman:
I already made a or for this. I think

zack-vii (commented):
from man of [fork](https://man7.org/linux/man-pages/man3/atexit.3.html):
When a child process is created via [fork(2)](https://man7.org/linux/man-pages/man2/fork.2.html), it inherits copies
       of its parent's registrations.  Upon a successful call to one of
       the [exec(3)](https://man7.org/linux/man-pages/man3/exec.3.html) functions, all registrations are removed.

so exit would indeed be not so nice to the main process.. however.. according to 
https://stackoverflow.com/questions/2329640/how-to-exit-a-child-process-exit-vs-exit

it may rather be `_Exit(0)`

merlea:
The man page in Linux indicates: "The function _Exit() is equivalent to _exit()." It's a question of using the C99 or POSIX.1 standards ... Either way is fine by me. Just let me know if I need to change it.

merlea:
I replaced `_exit` with `_Exit` following to the C99 standard.

zack-vii (dismissed):

zack-vii:
retest this please

zack-vii:
retest this please

merlea:
I don't quite understand why the tests failed. The first failure I see is from `tditest/testing/test-mdsip.tdi` which uses the `thread` protocol and so is not affected by this PR ...

Also before the last rebase the tests had also failed but at a different point (I think it was the rhel9 build).

zack-vii:
```log

==9713==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 52 byte(s) in 1 object(s) allocated from:
    #0 0x7f586fbf6667 in __interceptor_malloc (/lib64/libasan.so.6+0xb0667)
    #1 0x7f586c1bdaa6 in _send_response /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:506
    #2 0x7f586c1bdaa6 in send_response /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:559
    #3 0x7f586c1c3d81 in return_status /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:571
    #4 0x7f586c1c3d81 in standard_command /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:1036
    #5 0x7f586c1c3d81 in ProcessMessage /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:1332
    #6 0x7f586c1cdbae in ConnectionDoMessage /git/mdsplus/mdstcpip/mdsipshr/DoMessage.c:46
    #7 0x7f586c1d18bb in io_listen /git/mdsplus/mdstcpip/io_routines/IoRoutinesThread.c:66
    #8 0x7f586f3dc431 in start_thread (/lib64/libpthread.so.0+0x9431)

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x7f586fbf6667 in __interceptor_malloc (/lib64/libasan.so.6+0xb0667)
    #1 0x7f586c1b955f in GetMdsMsgTOC /git/mdsplus/mdstcpip/mdsipshr/GetMdsMsg.c:115
    #2 0x7f586c1cdad8 in ConnectionDoMessage /git/mdsplus/mdstcpip/mdsipshr/DoMessage.c:45
    #3 0x7f586c1d18bb in io_listen /git/mdsplus/mdstcpip/io_routines/IoRoutinesThread.c:66
    #4 0x7f586f3dc431 in start_thread (/lib64/libpthread.so.0+0x9431)

SUMMARY: AddressSanitizer: 100 byte(s) leaked in 2 allocation(s).
```

This change seems to cause a memory leak.

merlea:
How can a change in the disconnect method for tunnel-like protocols cause a memory leak in a test using the thread protocol?

zack-vii:
retest this please

zack-vii:
hm  you are right .. this might be an unrelated issue then. No idea why it only triggered now.

merlea:
This time the tests passed...

zack-vii (approved):

PR #2510, opened 2022-12-13
Add socket/service files for running per-connection mdsip servers with systemd

WhoBrokeTheBuild:
These will be placed in /etc/systemd/system/ but not enabled by default.

I'll need to add instructions somewhere to run the following:
```
sudo systemctl enable mdsip.socket
```


joshStillerman (commented):
You need to update the distro files for it to pass the checks.  When you do so, it will invalidate my approval, which you have.

joshStillerman (approved):

PR #2522, opened 2023-01-13
Fix: null terminate truncated node names (2472)

### LINKED ISSUE #2472: MDSplus8: -- inconsistent max node name length
Using the old-trees branch, create a new tree.   Using TCL, attempt to add a 70 character node name.   Dir command shows it was truncated to 63 characters.  Now open the same tree in jTraverser -- it shows 64 characters.   However jTraverser2 only shows 63 characters.

In the underlying data structures, is the node name a 65 character field (64 visible characters + 1 null terminator)?   Or is it a 64 character field (63 visible characters + 1 null terminator)?   Behavior demonstrated by this bug indicates inconsistent definition of the maximum node length.

old_trees_char64_bug

mwinkel-dev:
Fixes issue with mdsplus8 not properly truncating node names longer than 63 visible characters.

### DIFF
--- a/treeshr/TreeAddNode.c
+++ b/treeshr/TreeAddNode.c
@@ -159,8 +159,8 @@ int _TreeAddNode(void *dbid, char const *name, int *nid_out, char usage)
               strncpy(new_ptr->name, node_name, sizeof(new_ptr->name));
 #pragma GCC diagnostic pop
           for (i = strlen(node_name); i < sizeof(new_ptr->name); i++)
-            new_ptr->name[i] = 0;
-          new_ptr->name[MAX_NAME_LEN] = 0;  // null terminate truncated names
+            new_ptr->name[i] = '\0';
+          new_ptr->name[MAX_NAME_LEN] = '\0';
           new_ptr->child = 0;
           loadint16(&new_ptr->conglomerate_elt, &idx);
           if ((is_child && (usage == TreeUSAGE_ANY)) ||
--- a/treeshr/TreeAddNode.c
+++ b/treeshr/TreeAddNode.c
@@ -160,6 +160,7 @@ int _TreeAddNode(void *dbid, char const *name, int *nid_out, char usage)
 #pragma GCC diagnostic pop
           for (i = strlen(node_name); i < sizeof(new_ptr->name); i++)
             new_ptr->name[i] = 0;
+          new_ptr->name[MAX_NAME_LEN] = 0;  // null terminate truncated names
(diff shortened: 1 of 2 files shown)


treeshr/TreeAddNode.c
```diff
@@ -160,6 +160,7 @@ int _TreeAddNode(void *dbid, char const *name, int *nid_out, char usage)
 #pragma GCC diagnostic pop
           for (i = strlen(node_name); i < sizeof(new_ptr->name); i++)
             new_ptr->name[i] = 0;
+          new_ptr->name[MAX_NAME_LEN] = 0;  // null terminate truncated names
```
WhoBrokeTheBuild: When adding a null terminator to a string we prefer to use '\0'  instead, helps convey the intent. If you add that, you can drop the comment, it will be conveyed by the code

treeshr/TreeRenameNode.c
```diff
@@ -210,10 +210,16 @@ int _TreeRenameNode(void *dbid, int nid, char const *newname)
...
+  // Set name with logic similar to _TreeAddNode()
+#pragma GCC diagnostic push
+#if defined __GNUC__ && 800 <= __GNUC__ * 100 + __GNUC_MINOR__
+  _Pragma("GCC diagnostic ignored \"-Wstringop-truncation\"")
+#endif
+  strncpy(oldnode_ptr->name, newnode_name, sizeof(oldnode_ptr->name));
+#pragma GCC diagnostic pop
+  for (i = strlen(newnode_name); i < (int) sizeof(oldnode_ptr->name); i++)
+    oldnode_ptr->name[i] = 0;
+  oldnode_ptr->name[MAX_NAME_LEN] = 0;  // null terminate truncated names
```
WhoBrokeTheBuild: Same comment as above

WhoBrokeTheBuild (changes requested):

treeshr/TreeAddNode.c
mwinkel-dev: Agreed.

treeshr/TreeRenameNode.c
mwinkel-dev: Agreed.

mwinkel-dev:
Requested changes made and tested on Ubuntu 22.

WhoBrokeTheBuild (approved):

mwinkel-dev:
On 20-Apr-2023, manual testing done after the PR was merged confirmed that the fix works.

PR #2524, opened 2023-01-27
Fix: eliminate segfaults caused by erroneous unmap (#2523)

### LINKED ISSUE #2523: mdsplus8 has an intermittent segfault that Valgrind can trigger
There are two formats for treefiles.   Format-2 was introduced in mdsplus8; Format-1 is used by all prior versions of MDSplus.   Format-2 has 64-character node names (63 visible, plus a null terminator).   Format-1 has 12-character node names padded with trailing blanks (and no null terminator).

When mdsplus8 reads a Format-1 treefile, it has to do in-memory expansion of the data structures.   Thus there is a fair amount of memory shuffling going on.

The attached bug.c program shows that there is a bug in the mdsplus8 code that processes Format-1 treefiles.   Interestingly, the program runs fine at the Linux terminal prompt and also in gdb.   However when run in Valgrind, it causes a variety of crashes.

Although the Valgrind crash is repeatable and consistent for a given set of factors (valgrind parameters, same branch head, etc.), note that changing those factors will alter how memory is corrupted.   Most of the time the corruption is so bad that Valgrind crashes.   Occasionally, a set of factors can be selected that allows Valgrind to detect a segfault.   And the memory corruption also makes it difficult to debug the problem using Valgrind's remote monitor for gdb.

When the same program is linked with the alpha branch (MDSplus ver. 7), it runs fine in Valgrind (and also in gdb and at the terminal prompt).

To see the bug, follow the instructions in the header comments of the attached bug.c program.   Example output of a Valgrind crash has also been attached.  

[bug.txt](https://github.com/MDSplus/mdsplus/files/10514504/bug.txt)
[valgrind_err.txt](https://github.com/MDSplus/mdsplus/files/10514414/valgrind_err.txt)

mwinkel-dev:
The root cause of the crash is that close_top_tree() assumes it is dealing with a memory mapped file, thus calls munmap() to free up memory.  However when mdsplus8 reads a Format-1 treefile, it does not map the file into memory.   Because of the in-memory expansion of the node names, mdsplus8 has to instead use calloc() to allocate memory and then read in the file.   Applying munmap() to an address generated by calloc(), trashes the memory pages and marks them as inaccessible.   Thus causing a variety of segfaults and crashes.

The bug appears to be intermittent simply because it all depends on where data and variables are allocated in the heap.

The fix is simple, namely to set the TREE_INFO "mapped" bit to FALSE when mdsplus8 is dealing with Format-1 treefiles.

Note that with mdsplus8 there are two ways to read a file: memory mapped for Format-2 files and the slower calloc() / read process for Format-1 files.   This fix assumes that the entire MDSplus codebase only uses TREE_INFO's "mapped" bit to indicate how a file has been loaded.

However, if portions of MDSplus use the "mapped" bit to merely indicate that a file has been loaded (regardless of method and file format), then this fix will have to be revisited.   In that case, an alternative is to add logic to close_top_tree() to prevent applying the munmap() call when dealing with Format-1 treefiles.   The munmap() call is in TreeOpen.c around line 422.

Because this fix potentially breaks other features of MDSplus (i.e., any code that checks the "mapped" bit), it must be thoroughly tested.   Stated another way, this fix should be considered tentative until the mdsplus8 branch passes the standard regression tests used on all production releases of MDSplus.

### DIFF
--- a/treeshr/TreeOpen.c
+++ b/treeshr/TreeOpen.c
@@ -1122,6 +1122,7 @@ static int MapFile(int fd, TREE_INFO *info, int nomap)
 //    TreeEstablishRundownEvent(info);
     if (nomap)
     {
+      info->mapped = FALSE;
       if (header.version < 2)
       {
         int idx;


WhoBrokeTheBuild (approved):

mwinkel-dev:
Manual testing after the PR was merged confirmed issue is fixed.   Opened / closed a C-Mod shot 10,000 times without any segfaults.

PR #2526, opened 2023-01-30
Fix: Support thick mixed with local and distributed in a tree path

WhoBrokeTheBuild:
When getting/setting the current shot with a multi-part tree path that ends with ::, we currently fail to traverse the list of paths.

For example:
```
$ export "cmod_path=/tmp;alcdata-new::"
$ mdstcl show current cmod
Connect failed to host: /tmp;alcdata-new
Failed to get shotid.
```

This commit allows traversal of the list of paths, so this now works:
```
$ export "cmod_path=/tmp;alcdata-new::"
$ mdstcl show current cmod
Current shot is 1170112002
```

If there is a more standard way to traverse the list of paths, I would love to use it.

### DIFF
--- a/treeshr/TreeGetSetShotId.c
+++ b/treeshr/TreeGetSetShotId.c
@@ -73,6 +73,7 @@ static char * PathToFileName(const char * experiment, char * path)
   strcpy(filename, tmp);
   free(tmp);
   
+  // (slwalsh, jas, fsantoro) NOTE: We have no idea if this is even used anymore, possibly remove?
   if (filename[strlen(filename) - 1] == '+')
   {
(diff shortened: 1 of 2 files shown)


zack-vii:
i briefly skimmed the changes.. unfortnatly git fails to showa clean diff. Anyway on first glimps as much as I like the simplicity of strtok is threadsafe, is it. it seems like its using static variables.

Edit: ok   acvording to https://man7.org/linux/man-pages/man3/strtok_r.3.html
it is thread-safe, but not multi-thread-safe, you should use strtok_r instead.

joshStillerman (approved):

PR #2527, opened 2023-01-31
CMake Build System

WhoBrokeTheBuild:


dgarnier:
The current ./bootstrap is not actually working.  It makes .c files from .l and .y in the source tree, but the cmake files expect them to be in the build directories.  I have a branch that replaces the bison/flex part of bootstrap.  I did have to make small changes to the .l and .y files though, because of the hardwiring of the output files inside those input files.  See: https://github.com/dgarnier/mdsplus/tree/feature/cmake_bison_flex

mwinkel-dev:
I have perused the 343 files changed by this work-in-progress PR and didn't see anything alarming.   This high-level review confirmed that the changes are in the expected categories (Jenkins build scripts, CMake scripts, tests, environment variables, platforms, generation scripts, removing obsolete files, etcetera).   With such a large PR, there could of course be some overlooked bugs.   However, the risk of that is low because testing apparently shows that this CMake PR does build on many platforms and the resulting MDSplus does pass the automated test suite.   Thus, if there are overlooked bugs, they are likely minor and thus should be easy to fix after this PR is eventually merged.

mwinkel-dev (dismissed):
I have also reviewed the two commits made after my previous post.   Those commits also look reasonable, so am approving this PR.

mwinkel-dev (dismissed):
The last commit has three minor changes:

1. Bootstrap script for backwards compatibility for people used to the old build scripts.
2. Deletion of extraneous printf() statements.
3. Removal of an obsolete file.

All looks good, so approving this PR.

mwinkel-dev (approved):
Most recent commit has a retry loop for downloading the Java library that implements SSH in Java programs (aka jsch.jar).   Plus a minor formatting change in the CMakeLists.txt file.

The changes are straightforward, so approving this PR.

PR #2531, opened 2023-02-04
Intel Fortran (ifort) errors because the boolean type doesn't match the prototype

dgarnier:
The prototype had explicit boolean .false. in it.  Just convert to use boolean.

### DIFF
--- a/math/csint.f
+++ b/math/csint.f
@@ -254,7 +254,7 @@
       Do j = 1,ndata
             idx(j) = j
       Enddo
-      Call SVRGP(ndata,xdata,break,1,idx)
+      Call SVRGP(ndata,xdata,break,.true.,idx)
       Do j = 1,ndata
             c(1,idx(j)) = fdata(j)
             c(2,idx(j)) = dfdata(j)
--- a/math/dcsint.f
+++ b/math/dcsint.f
@@ -248,7 +248,7 @@
       Do j = 1,ndata
             idx(j) = j
       Enddo
-      Call DSVRGP(ndata,xdata,break,1,idx)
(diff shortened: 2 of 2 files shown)


math/csint.f
```diff
@@ -254,7 +254,7 @@ Subroutine C9INI(ndata,xdata,fdata,dfdata,break,c)
       Do j = 1,ndata
             idx(j) = j
       Enddo
-      Call SVRGP(ndata,xdata,break,1,idx)
```
joshStillerman, line 257: I liked this code better before

math/dcsint.f
```diff
@@ -248,7 +248,7 @@ Subroutine DC9INI(ndata,xdata,fdata,dfdata,break,c)
       Do j = 1,ndata
             idx(j) = j
       Enddo
-      Call DSVRGP(ndata,xdata,break,1,idx)
+      Call DSVRGP(ndata,xdata,break,.true.,idx)
```
joshStillerman, line 251: Is it really supposed to be .true. - I hate fortran

math/csint.f
dgarnier: then should I change the prototype that doesn't match?

math/dcsint.f
mwinkel-dev: Yes, in FORTRAN the boolean values are `.TRUE.` and `.FALSE.`.   That has apparently been so since FORTRAN IV /  FORTRAN 66 was introduced in the mid/late 1960s.

WhoBrokeTheBuild (approved):

PR #2534, opened 2023-02-16
Fix:corrected missing copy of isEdit flag in Tree(Tree *) constructor

GabrieleManduchi:
This fix was already reported.....for some obscure reason it disappeared


GabrieleManduchi:
retest this please

joshStillerman (approved):

PR #2535, opened 2023-02-16
Fix: mdsconnect incorrectly aborts on the first connection

WhoBrokeTheBuild:
Somewhere along the line TdiShrExt->rMdsConnect started returning the connection number (starting at 0) instead of a status code.
So on an error, it actually returns -1, not 0.
If mdsconnect is called with _abort set, then it will abort() on the first connection (_status == 0).


mwinkel-dev:
This is a duplicate of PR #2889 (merged to "alpha" on 25-Mar-2025).

Here is background on the root cause . . . 

PR #2288 changed the definition of `INVALID_CONNECTION_ID` to `-1` (it used to be zero). This change was made in mdstcpip/mdsip_connections.h on 19-May-2021. (The reason for the change is unknown, but perhaps there were inconsistent hardcoded definitions scattered throughout the source code.).  Because the PR involved hundreds of files, the impact of this redefinition of `INVALID_CONNECTION_ID` was overlooked.

PR #2541, opened 2023-03-08
Add RHEL9 build, remove xinetd package dependency, bug fixes

WhoBrokeTheBuild:
Add rhel9.opts to use mdsplus/builder:rhel9 (which is based on Rocky Linux 9.1, as CentOS 9 does not exist)

Remove xinetd as a package requirement in linux.xml.
If `/etc/xinet.d/` is present, it will still install the config files, but the packages will no longer depend on it. This is because we now have systemd files as an alternative, and RHEL9 has officialy dropped support for it, so this seems like a good time to cut ties.

When installing the `mdsplus-*-kernel` package on a system without xinetd or systemd, the post install script will throw errors trying to copy files. Instead, we wrap them in if statements to only copy the files if those directories exist.

Replace `.getiterator()` with `.iter()` to fix error when using `xml.ElementTree` with python3 in `redhat_build_rpms.py` and `alpine_build_apks.py` This already seemed to be fixed in `debian_build_debs.py`

Removed seemingly duplicate include for `libdc1394_support*.so` in `linux.xml`

### DIFF
--- a/deploy/packaging/redhat/redhat_build_rpms.py
+++ b/deploy/packaging/redhat/redhat_build_rpms.py
@@ -97,7 +97,7 @@ def build():
     root = common.get_root()
     bin_packages = list()
     noarch_packages = list()
-    for package in root.getiterator('package'):
+    for package in root.iter('package'):
         attr = package.attrib
         if attr["arch"] == "noarch":
             noarch_packages.append(package)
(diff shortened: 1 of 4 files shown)


deploy/packaging/redhat/redhat_build_rpms.py
```diff
@@ -97,15 +97,15 @@ def build():
...
+    for package in root.iter('package'):
         attr = package.attrib
         if attr["arch"] == "noarch":
             noarch_packages.append(package)
         else:
             bin_packages.append(package)
     architectures = [{"target": "x86_64-linux",
                       "bits": 64, "arch_t": ".x86_64"}]
-    if info['dist'] != 'el8':
+    if info['dist'] != 'el8' and info['dist'] != 'el9':
```
joshStillerman, line 108: Shouldn't this be 'or'

WhoBrokeTheBuild: I actually did a double take on the same place

It only enables 32-bit when it's neither el8 nor el9, so `!= 8 && != 9`

joshStillerman (approved):

PR #2543, opened 2023-03-17
Gm fix events

### LINKED ISSUE #2542: Configuration of  mdsip server for TCP/IP-based events transferred among subnetworks
I try to send mdsevent across  subnetworks. From the [wiki](https://www.mdsplus.org/index.php/Documentation:Tutorial:Events), I set the define env variable mds_event_server in Python as the mdsip address, and the event object returns an error.
```
from MDSplus import Event
import os
os.environ['mds_event_server'] = '10.78.2.113'
os.environ.pop('UDP_EVENTS')
class MyEvent(Event):
    def run(self):
            print("RECEIVED EVENT " + self.getName())
            print(self.getTime())
            print(self.getRaw())

eventObj=MyEvent('test_event')
#ConnectToMds_: Success
#Error connecting to 10.78.2.113>>> 
```
The MDSplus server has no extra configuration except the tree path. So how to configure the server to support the event to transfer the event based on TCP/IP?

If the mds_event_server env is not defined in the script, the eventObj can receive the event from other machines which send the event in the same subnet. 

Another question is that the Jscope doesn't update the data,  even if the jscope is in the same subnet.

GabrieleManduchi:
The past changes in the management of static variabled in mdsip communication broke the management of MDS events over mdsip (i.e. when either mds_event_server or mds_event_target env variables are set).
Fixed in this version of MdsEvents.c

### DIFF
--- a/mdsshr/MdsEvents.c
+++ b/mdsshr/MdsEvents.c
@@ -396,6 +396,7 @@ static void *handleRemoteAst(void *arg )
   struct MdsipEventInfo *eventInfo = arg;
   int curr_eventid;
 
+  if(!arg) return NULL;
   getServerDefinition("mds_event_server", receive_servers, &num_receive_servers);
   newRemoteId(eventInfo->eventid);
   for (idx = 0; idx < num_receive_servers; idx++)
--- a/mdsshr/MdsEvents.c
+++ b/mdsshr/MdsEvents.c
@@ -216,9 +216,9 @@ static void ReconnectToServer(int idx, int recv)
     return;
   DisconnectFromMds_(ids[idx]);
   ids[idx] = ConnectToMds_(servers[idx]);
-  if (ids[idx] < 0)
+  if (ids[idx] == INVALID_CONNECTION_ID)
   {
-    printf("\nError connecting to %s", servers[idx]);
+    printf("\nError connecting to %s\n", servers[idx]);
     perror("ConnectToMds_");
     sockets[idx] = 0;
   }
@@ -406,9 +406,9 @@ static void *handleRemoteAst(void *arg )
     {
       receive_thread_ids[idx] = ConnectToMds_(receive_servers[idx]);
     }
-    if ( receive_thread_ids[idx] < 0)
+    if ( receive_thread_ids[idx]  == INVALID_CONNECTION_ID)
     {
-      printf("\nError connecting to %s", receive_servers[idx]);
+      printf("\nError connecting to %s\n", receive_servers[idx]);
       perror("ConnectToMds_");
       receive_thread_sockets[idx] = 0;
     }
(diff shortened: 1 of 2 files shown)


mdsshr/MdsEvents.c
```diff
@@ -394,41 +454,33 @@ static void *handleRemoteAst(void *arg __attribute__((unused)))
     }
     for (i = 0; i < num_receive_servers; i++)
     {
-      if (receive_ids[i] > 0 && FD_ISSET(receive_sockets[i], &readfds))
```
zack-vii, line 397: you broke the select choice

mdsshr/MdsEvents.c
```diff
@@ -366,20 +383,63 @@ static void KillHandler()
...
+    }
+  }
   while (1)
   {
     FD_ZERO(&readfds);
+
     for (i = 0; i < num_receive_servers; i++)
-      if (receive_sockets[i])
-        FD_SET(receive_sockets[i], &readfds);
+    {
```
zack-vii, line 438: what was the motivation to switch from receive_sockets to receive_thread_sockets .. copy and paste or was this actually a bug

mdsshr/MdsEvents.c
```diff
@@ -366,20 +383,63 @@ static void KillHandler()
...
   Message *m;
   int selectstat;
   fd_set readfds;
+  int receive_thread_ids[256];
+  int receive_thread_sockets[256];
+  struct MdsipEventInfo *eventInfo = arg;
+  int curr_eventid;
+
+  getServerDefinition("mds_event_server", receive_servers, &num_receive_servers);
+  newRemoteId(eventInfo->eventid);
```
zack-vii, line 401: you never check if arg is not NULL

mdsshr/MdsEvents.c
```diff
@@ -199,7 +216,7 @@ static void ReconnectToServer(int idx, int recv)
     return;
   DisconnectFromMds_(ids[idx]);
   ids[idx] = ConnectToMds_(servers[idx]);
-  if (ids[idx] <= 0)
```
zack-vii, line 202: maybe its worth using the constant INVALID_CONNECTION_ID  instead.  is this conform?

mdsshr/MdsEvents.c
GabrieleManduchi: Now the thread must open the connection itself, this is the reason of the receive_thread_ids and receive_thread_sockets that are private to the thread. Formerly the connection was opened by the main thread but connection structures were not visible by the created thread.

zack-vii: I see. hmm i am not sure if our test cover it but i remember for backward compatibility it was important that the main thread holds the threads.. i think it was because jobs may run in a thread and dynamically open a connection.. which would then be closed once the job is done. resulting in very slow network performance.. this may be addressed elsewhere no (treeshr was altered in the way it connects to remote trees) but it should be verified. its quite a while ago.. but i f remember correctly some older mdsip serverse.g. do not multiple connections and hence it is required to used a shared one.

zack-vii: I see. hmm i am not sure if our test cover it but i remember for backward compatibility it was important that the main thread holds the threads.. i think it was because jobs may run in a thread and dynamically open a connection.. which would then be closed once the job is done. resulting in very slow network performance.. this may be addressed elsewhere no (treeshr was altered in the way it connects to remote trees) but it should be verified. its quite a while ago.. but if i remember correctly some older mdsip serverse.g. do not multiple connections and hence it is required to used a shared one.

mdsshr/MdsEvents.c
GabrieleManduchi: True. corrected

mdsshr/MdsEvents.c
GabrieleManduchi: True, corrected

mdsshr/MdsEvents.c
GabrieleManduchi: In any case I would not be concerned here for very old mdsip servers. Usage of mdsip events is very rare and in any case before this fix it was simply broken. So I don't see a limitation if mdsip events do not work on very old mdsip servers (unlike remote tree access)

zack-vii: that may be but mdsip is not just used for event.. you would break access to old server hosting mdsplus trees

mdsshr/MdsEvents.c
GabrieleManduchi: True, added again

mdsshr/MdsEvents.c
GabrieleManduchi: I don't know if such mdsip servers still exist, i.e. mdsip servers that do not accept more than one connection, I have no memory of such a limitation.

GabrieleManduchi: And in any case a single connection is established in MdsEventAst

joshStillerman (approved):

PR #2556, opened 2023-03-31
Fix: Change public SendArg decl match private

### LINKED ISSUE #2404: Conflicting declarations of SendArg (mdsipshr)
In `include/ipdesc.h` `SendArg` is declared as:
```
extern int SendArg(int s, unsigned char i, char dtype, unsigned char nargs,
                   short len, char ndims, int *dims, char *ptr);
```
while in `mdstcpip/mdsip_connections.h` and `mdstcpip/mdsipshr/SendArg.c` it is defined and declared as:
```
int SendArg(int id, unsigned char idx, char dtype, unsigned char nargs,
            unsigned short length, char ndims, int *dims, char *bytes)
```
The difference is in the 5th argument `length`.

I think the declaration in `include/ipdesc.h` should be changed to `unsigned short len`. I am not sure whether other parts of the build would be impacted by this.

joshStillerman:
As per https://github.com/MDSplus/mdsplus/issues/2404, changed the public include to match the private one.  MDSplus source code still compiles without warnings.

closes #2404

### DIFF
--- a/include/ipdesc.h
+++ b/include/ipdesc.h
@@ -43,7 +43,7 @@ extern "C"
 #define INVALID_CONNECTION_ID -1
   extern int ConnectToMds(char *host);
   extern int SendArg(int s, unsigned char i, char dtype, unsigned char nargs,
-                     short len, char ndims, int *dims, char *ptr);
+                     unsigned short len, char ndims, int *dims, char *ptr);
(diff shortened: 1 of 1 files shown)


mwinkel-dev (approved):
Change is A-OK.   (Thanks for making this fix for me.)

PR #2569, opened 2023-05-05
Gm fix java apd

GabrieleManduchi:
Added fixes both to MDSplus.Apd.java and jni mdsobjects.c for reding and writing APD objects in tree


AndreaRigoni (approved):
OK

PR #2571, opened 2023-05-09
Fix: Add upgrades to PG devices

santorofer:
Some clean up of the code.
Add debugging when using D-Tacq API that loads the STL.
Remove the usage of putData and instead directly add STL as text to the output nodes.

Note:
Passed tests.


santorofer:
Retest this please

WhoBrokeTheBuild (approved):

PR #2573, opened 2023-05-11
Gm matlab apd

GabrieleManduchi:
Added support for storing/retrieving any MATLAB (array of)structures, possibly recursive.
This required some fixes in APD management both in Java and python


joshStillerman (approved):
I am approving this as is.
Can you check (separately) that the python version of the MATLAB API also works?

GabrieleManduchi:
Ok, never did it before, but I will look into it and add the APD support in MATLAB also for python

PR #2577, opened 2023-05-22
Fix: Update the regex in influxSignal

WhoBrokeTheBuild:
Change the * to a + in the regex replace triggered by deltaTime being set. This was leading to a strange bug (possibly caused by going from python2 to python3) where it would match the entire fieldKey, and then match the empty space at the end of the line. 

So instead of:
`MEAN(fVal)`
it would become:
`MEAN(fVal)MEAN()`

.. which would then make InfluxDB reject our query.


santorofer:
This has been tested and works correctly.

santorofer (approved):

PR #2578, opened 2023-05-26
Gm update devices

GabrieleManduchi:
A bunch of small fixes for jTraverser,  jScope, jDispatcher
A Set of new and updater devices for RFX


AndreaRigoni (approved):
OK

Issue #2580, opened 2023-05-30
Multiple alpha build versions' IDL fails on first MDSconnect


ModestMC:
**Affiliation**
General Atomics / DIII-D CSS

**Version(s) Affected**
Alpha Releases 7.139.23, 7.139.28, 7.139.35-37 

**Platform**
RHEL 8

**Describe the bug**
Currently on our systems we use  alpha_release-7.130.1, but we are needing to update our build to some version which includes the fixes made in #2537. We have now tested several versions of MDSplus to try and find a working build, however each version we have tested has a bug where the first IDL mdsconnect attempt errors out. Furthermore the error occurs each time a new IDL shell is created (see screenshot below).

Also, it appears the status variable doesn't catch anything either:
```
IDL> mdsconnect,'[the server]',status=status
% Compiled module: MDSCONNECT.
% Compiled module: MDSDISCONNECT.
% MDSCONNECT: Error connecting to [the server]
% Execution halted at: $MAIN$
IDL> help,status
STATUS          UNDEFINED = <Undefined>
IDL> mdsconnect,'[the server]',status=status
IDL> help,status
STATUS          INT       =        1
```

**To Reproduce**
``` git checkout [the particular tag]
./bootstrap
./configure --prefix=[your build directory]
make
make install
[whatever you do for pathing]
idl
>mdsconnect, '[the URL]
```

**Expected behavior**
It should just connect with no error the way it does on the second attempt.

**Screenshots**
![Screenshot 2023-05-30 at 2 49 39 PM](https://github.com/MDSplus/mdsplus/assets/45748401/42a2445f-f748-4f55-8400-7088f797d244)

**Additional context**
We just need a new version of MDSplus that fixes the numpy string issue/update that works otherwise.

joshStillerman:
I am pretty sure we have a fix for this. Will investigate tomorrow. Sent from my iPhoneOn May 30, 2023, at 7:35 PM, Mitchell Clark ***@***.***> wrote:﻿
Affiliation
General Atomics / DIII-D CSS
Version(s) Affected
Alpha Releases 7.139.23, 7.139.28, 7.139.35-37
Platform
RHEL 8
Describe the bug
[...]

PR #2581, opened 2023-05-31
Fix: socket returned can now be zero

joshStillerman:
IdlConnectToMds can now return zero as a valid answer.

This is true at least on linux.  It should be tested on other platforms, the code may need further adjustment.

There is specific conditional code for MacOS.

Please approve this PR so that a linux customer can proceed

addresses: https://github.com/MDSplus/mdsplus/issues/2580


mwinkel-dev (dismissed):
I am assuming that the new "print" statement at line 165 is intentional (and not a vestigial debug statement).   If so, this fix looks good to me.

mwinkel-dev (approved):
Extraneous debug "print" statement was removed.   Now is a clean fix.

joshStillerman:
retest this please

mwinkel-dev:
This PR is likely related to Issue #2625 (see notes in that issue).

PR #2583, opened 2023-06-06
Build.sh now has a --disable-labview option, plus allows comment lines in deploy/os/*.opts files

mwinkel-dev:


deploy/build.sh
```diff
@@ -417,7 +423,7 @@ fi
 #
 if [ -r "${SRCDIR}/deploy/os/${OS}.opts" ]
 then
-    os_opts=$(cat ${SRCDIR}/deploy/os/${OS}.opts)
+    os_opts=$(fgrep -v "#" ${SRCDIR}/deploy/os/${OS}.opts)
```
WhoBrokeTheBuild: is `fgrep` common enough to rely on here?

mwinkel-dev: Good question (and a good catch).   Every system I've worked on has had fgrep.   However, it would be safer to use grep.   Especially because the man page (on Ubuntu 20) indicates that egrep, fgrep and rgrep are now deprecated.

I will make the change, test it, and then resubmit.

mwinkel-dev:
Closing this PR because wanted a rebase, but the GitHub website instead did a merge.   So will delete my dev branch and submit as a new PR.

mwinkel-dev:
Restored the mw-build branch.   And re-opening this PR (good training exercise for me).

WhoBrokeTheBuild (approved):

PR #2591, opened 2023-06-28
Fix: issue 2589 -- tostring() deprecated since numpy 1.19.0, replaced with tobytes()

### LINKED ISSUE #2589: numpy deprecation warning for .tostring()
Hello,

Numpy is warning that `.tostring()` is a deprecated.

Here is the warning:
```terminal
  /tmp/mdsplus-stable/python/MDSplus/mdsscalar.py:437: DeprecationWarning: tostring() is deprecated. Use tobytes() instead.
    return cls(_N.array(_C.cast(d.pointer, _C.POINTER((_C.c_byte*d.length))).contents[:], dtype=_N.uint8).tostring())
```
the line which is causing the issue is here: [mdsscalar.py#L437](https://github.com/MDSplus/mdsplus/blob/alpha/python/MDSplus/mdsscalar.py#L437).

I am using:
```terminal
numpy                           1.24.3
MDSplus                         7.132.0
```

Here is the [numpy documentation](https://numpy.org/devdocs/reference/generated/numpy.ndarray.tostring.html), which states **"Despite its name, it returns bytes not strs."**, so it's not surprising that it has been deprecated, and changing to `tobytes()` should produce exactly the same results.

mwinkel-dev:
Should also be cherry-picked into the stable branch.

### DIFF
--- a/python/MDSplus/mdsscalar.py
+++ b/python/MDSplus/mdsscalar.py
@@ -434,7 +434,7 @@ class String(Scalar):
     def fromDescriptor(cls, d):
         if d.length == 0:
             return cls('')
-        return cls(_N.array(_C.cast(d.pointer, _C.POINTER((_C.c_byte*d.length))).contents[:], dtype=_N.uint8).tostring())
+        return cls(_N.array(_C.cast(d.pointer, _C.POINTER((_C.c_byte*d.length))).contents[:], dtype=_N.uint8).tobytes())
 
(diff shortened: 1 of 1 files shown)


WhoBrokeTheBuild (dismissed):

mwinkel-dev:
retest this please

WhoBrokeTheBuild (approved):

PR #2595, opened 2023-07-12
Fixed error in `Connection.getMany` python API

AreWeDreaming:
The `Connection.getMany`  was using the wrong number of arguments for the  class GetMany constructor. 
This fix works for my purposes but I cannot say how general it is.
Without this fix `Connection.getMany` is broken.


ModestMC:
@WhoBrokeTheBuild may want to see this.

joshStillerman:
Before making this change, I would want to look at what the keyword 'value' argument used to do.

WhoBrokeTheBuild (approved):

ModestMC:
@AreWeDreaming success.jpg

PR #2601, opened 2023-08-04
Feature/cmake bison flex

dgarnier:
I dunno.. I ran ./bootstrap and it doesn't actually work currently.  This is because it puts its products in the source tree and not the binary / build tree where cmake is expecting them.  Anyway, here is an alternative for the bison/flex part of bootstrap.


dgarnier:
oops.. this is wrong.

PR #2606, opened 2023-08-30
Gm fix connection

GabrieleManduchi:
Like TreeNode, also Connection must keep track of the current thread, possibly re-instantiating the connection in order to a;;pw Connection object be exchanged among threads


AndreaRigoni (approved):
Perfect !!@

PR #2611, opened 2023-09-03
Fix: minor change to UDT4 protocol for CMake and macOS

### LINKED ISSUE #2597: MacOS Apple Silicon MDSplus Build
**Affiliation**
DIII-D

**Description**

mwinkel-dev:
The UDT4 protocol was failing on macOS Ventura because the CMake build does not define the OSX environment variable.

Additional detail:
- With the Autotools build system, the OSX environment variable was defined in the Makefile.   In the new CMake based build system, OSX is not defined and thus the UDT4 test cases fail. 
- Issue #2597 requires this PR.

### DIFF
--- a/mdstcpip/udt4/src/channel.cpp
+++ b/mdstcpip/udt4/src/channel.cpp
@@ -124,7 +124,7 @@ void CChannel::open(UDPSOCKET udpsock) {
 }
 
 void CChannel::setUDPSockOpt() {
-#if defined(BSD) || defined(OSX)
+#if defined(BSD) || defined(__APPLE__)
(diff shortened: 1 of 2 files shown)


joshStillerman (approved):

dgarnier:
You could also do this without changing the code, and just add OSX definition to the CMakeLists.txt

mwinkel-dev:
Before submitting this PR, Stephen and I discussed the various options for fixing the missing OSX environment variable (including having CMake supply the environment variable).   This PR implements the solution that Stephen recommended.

mwinkel-dev:
Retest this please

PR #2612, opened 2023-09-03
Fix: change ports on test case to avoid conflict on macOS Big Sur

### LINKED ISSUE #2597: MacOS Apple Silicon MDSplus Build

mwinkel-dev:
On Josh's Intel macOS Big Sur, a test failed because port 7000 was already being used by ControlCenter.

Additional detail:
- It is unclear why there was a port conflict on Intel macOS Big Sur but not on Intel macOS Ventura.
- Josh requested that the test case be changed to use port 7500, so changed all three port assignments accordingly.
- It is possible (likely?) that port assignments for all test cases (including this one) will change again when the switch from Autotools to CMake occurs.
- Issue #2597 requires this PR.

### DIFF
--- a/python/MDSplus/tests/connection_case.py
+++ b/python/MDSplus/tests/connection_case.py
@@ -69,7 +69,7 @@ class Tests(_common.TreeTests, _common.MdsIp):
                 print(nci, t, l, c)
                 raise
         server, server_port = self._setup_mdsip(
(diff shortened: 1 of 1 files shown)


joshStillerman (approved):

mwinkel-dev:
Retest this please

Issue #2576, opened 2023-05-17
jTraverser2 fails to display VMS floating point numbers (DTYPE_F)


santorofer:
**Affiliation**
MIT / PSFC

**Version(s) Affected**
MDSplus Release:  alpha_release_7.139.33

**Platform**
Ubuntu 20

**Describe the bug**
jTraverser2 fails to display VMS floating point numbers (DTYPE_F).

For example, when opening CMOD tree for a particular shot number, the node TSTART should contain the value -4. 
While instead, it contains the value 69.414721F-42, see screenshot below:

![Screenshot 2023-05-17 at 5 07 26 PM](https://github.com/MDSplus/mdsplus/assets/44955673/df32fca8-f114-4545-97a1-b06e14856ebf)

While, if the same node is accessed using Python:

```
>>> import MDSplus
>>> t=MDSplus.Tree('cmod', 1160909001)
>>> t.TSTART.data()
-4.0
>>>
```

**To Reproduce**
1. Start jTraverser2
2. Open a CMOD shot.
3. Hover over the node TSTART
4. See the erroneous value being displayed

**Expected behavior**
The node value shown by jTraverser should be a correct floating point number, in this case, -4.0

zack-vii:
thisbius expected as non ieee floats have not been implemented yet.. one would have to convert them based on dtype before creating Float32Arrays or Float32 or implement a wrapping class e.g. FloatVMS* that handles those dtypes. same holds for other exotic float types.

santorofer:
Timo, we have been debugging (jTraverser2 and mdsplus-api) this issue for a while, and found that it stems from ByteBuffer being used to read the floats from their binary representation. We want to add a call to TDI's `FT_FLOAT`/`FS_FLOAT`, but couldn't figure out where to put it. Since you are more familiar with that API, can you take a crack at it?.

zack-vii:
the conversion could be done during deserialization
https://github.com/MDSplus/mdsplus/blob/0df72258c108db26df57160e483796c98d941d81/java/mdsplus-api/src/main/java/mds/data/descriptor/Descriptor_S.java#L19

here as an example for scalars the dtype is used to map to the appropriate java class. one could either convert here before passing the buffer to the Float32. or one could make a dedicated FFloat class that handles the conversion in getValue().

same could be done for the DFloat (same as FC, DC, G, and GC).

I will try to find some time to give the getValue() attempt a try.. I would implement the conversion in java though rather than relying on tdi if its not too complicated.

PR #2616, opened 2023-09-05
feature: jtraverser2 dynamically convert non-native floating formats

zack-vii:
* this is an untested attempt to address issue #2576. 
* this is a cheap way to intercept unsupported dtypes converting them to native types.
* it may be more performant to convert only in getValue/setValue/getAtomic ... but requires sub-classes of each Float32/64 and Compex32/64 as well as their Array variants.
* this does not mark DTYPE_F etc. as supported but merely allows to import them from serialized sources.
* this will not preserve the original dtype (mdsip cannot do raw float_x other than FT and FS)

### DIFF
--- a/java/mdsplus-api/src/main/java/mds/data/DTYPE.java
+++ b/java/mdsplus-api/src/main/java/mds/data/DTYPE.java
@@ -30,7 +30,7 @@ public enum DTYPE
 	OU("Octaword_Unsigned", "OU"), // 25
 	O("Octaword", "O"), // 26
 	G("G_Float", "G"), // 27
-	H("H_Float", "H"), // 28
+	H("H_Float", "H"), // 28 /*bad*/
 	GC("G_Complex", "G"), // 29 IEEE
 	HC("H_Complex", "H"), // 30 128bit
 	CIT, // 31 COBOL Intermediate Temporary
--- a/java/mdsplus-api/src/main/java/mds/Mds.java
+++ b/java/mdsplus-api/src/main/java/mds/Mds.java
@@ -85,7 +85,7 @@ public abstract class Mds implements AutoCloseable
 	public static final Pattern dollar = Pattern.compile("\\$[0-9]*(?=[^a-zA-Z]|$)");
 	protected static final int MAX_NUM_EVENTS = 256;
 	private static Mds active;
-	private static MdsIp shared_tunnel = null;
+	private static Mds shared_tunnel = null;
 
 	@SuppressWarnings(
 	{ "unchecked", "rawtypes" })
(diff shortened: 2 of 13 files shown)


zack-vii:
maybe one can test this and/or write an appropriate unit test

WhoBrokeTheBuild (dismissed):

santorofer:
Timo, I was testing your changes, but then I realized this will not work, because Float32(b) returns back already the wrong answer. 
And indeed, the call:
`(Float32)Mds.getLocal().getDescriptor("FS_FLOAT($)", new Float32(b));`
does not work.

zack-vii:
hm if the data gets into Float32 via deserialization, it should contain the original byte code of a F_Float. the tdicall should use this byte code unchanged and feed it to FS_FLOAT. The returned bytcode for a FS float should deserialize into a Float32 with FS_FLOAT  dtype.. well in theory .. I will check where it takes a different turn

zack-vii:
it seems the problem is that mdsip supports neither F_FLOAT nor G_FLOAT nor D_FLOAT as argument extension.
```tdi
mdsconnect('thread://1')
Connected: user
1
mdsvalue('DECOMPILE($)', 1.234G0)
"1.234D0"
```
As you can see the value was transformed from G_FLOAT (G) to FT_FLOAT (D) when passed down the mdsip protocol. Hence mdsplus-api cannot use the mdsip backend to convert.

zack-vii:
@santorofer this time it should work, took me long enough to pick this one up again but here it hopefully is ;)

santorofer:
Hi Timo, no problem. Excellent, I will test it and approve it as soon as I can.

java/mdsplus-api/src/main/java/mds/data/DTYPE.java
```diff
@@ -30,7 +30,7 @@ public enum DTYPE
 	OU("Octaword_Unsigned", "OU"), // 25
 	O("Octaword", "O"), // 26
 	G("G_Float", "G"), // 27
-	H("H_Float", "H"), // 28
+	H("H_Float", "H"), // 28 /*bad*/
```
WhoBrokeTheBuild, line 33: Does this comment mean that H_Float still doesn't work?

santorofer (approved):
Tested (repeat the workflow from the Issue #2576) and it seems to work correctly.

WhoBrokeTheBuild (approved):

java/mdsplus-api/src/main/java/mds/data/DTYPE.java
zack-vii: there does not seem to be support for h_float anymore. so jtraverser2 cannot use it either. if i remember correctly i tried to craft an H_FLOAT value in tditest and failed.

WhoBrokeTheBuild: Works for me, just wanted to call it out explicitly, I saw the error handling for it too which looks good :+1:

PR #2617, opened 2023-09-06
Info: add comments explaining MDSplus status codes

### LINKED ISSUE #2597: MacOS Apple Silicon MDSplus Build

mwinkel-dev:
As a maintenance developer, it was a surprise to learn the following:
- the 3-bit severity field interacts with the `STATUS_OK` macro to create 4 flavors of failure and 4 flavors of success,
- the SsINTERNAL code is not treated as a failure but is instead a flavor of success (because low-order bit is set),
- the Ss facility code is zero, yet SsINTERNAL has a facility code of minus one,
- that status codes consist of facility code, message number and severity

Status codes are an important concept in the MDSplus source code, thus deserve a few comments of explanation.

Work on Issue #2597 resulted in many exceptions and segfaults.  The above information was useful in debugging / fixing the issues.

### DIFF
--- a/deploy/gen_messages.py
+++ b/deploy/gen_messages.py
@@ -278,6 +278,29 @@ jma_tail = """\t\t\tdefault:
 \t}
 }"""
 
+# Severity scheme is modeled after VAX VMS.
+#
+# The status code is 32 bits, with these three fields:
+#    16 bit facility code in high order bits,
+#    13 bit message number, 
+#     3 bit severity (low order bits).
+#
+# Every odd severity is a flavor of success ("status & 0x1" is easy check).
+# Every even severity is a flavor of error.
+#   W = warning = 0   -- bad
+#   S = success = 1   -- OK
+#   E = error = 2     -- bad
+#   I = info = 3      -- OK
+#   F = fatal = 4     -- bad
+#   ? = unused = 5    -- OK
+#   ? = unused = 6    -- bad
+#   ? = internal = 7  -- OK (see usage in mdsshr/mdsshr_messages.xml)
+#
+# SsINTERNAL is all bits set in all fields (= -1), and is unusual because it
+# is both informational (when trapped by calling code) and an error (when
+# it is an unhandled exception that the user sees).  It is also unusual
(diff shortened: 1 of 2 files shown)


mwinkel-dev:
Regarding SsINTERNAL, here is additional detail:
- As an experiment, changed severity "internal" from 7 to 6 (a failure) whereupon the build failed three tests.   It appears that `TdiCull.c` treats SsINTERNAL as an informational code, which it traps and then either clears or converts to a different exception.
- Should the text label for the "internal" severity code remain `?` or be changed to `X`?
- Should SsINTERNAL be changed to use facility code 0 like the rest of the Ss facility (i.e., SsINTERNAL = 0x0000FFFF)?
- Should SsINTERNAL be renamed SsxINTERNAL, where `Ssx` is a new facility with facility code = 0xFFFF (i.e., SsxINTERNAL = 0xFFFFFFFF)?
- Should the "internal" severity code be changed to a flavor of failure (i.e., = 6), and thus SsINTERNAL changed to 0xFFFFFFFE?   If so, that will require fixing three failing tests that presently treat SsINTERNAL as a flavor of success.
- Should the comments about SsINTERNAL be removed from the PR (i.e., for brevity in the source code, instead have the comments refer to this PR)?

Given the widespread use of SsINTERNAL in the source code, the most conservative approach is to leave the code "as is" and merely add comments to explain why SsINTERNAL is unusual.

zack-vii:
Hi, Sorry about the confusion. the SsINTERNAL was introduced to give a name-tag to the widely used unnamed literal -1.  With an appropriate tool it should be quite easy to rename all tags in one commit.  However as the name suggests, SsINTERNAL should never surface to the user. It may however be returned through user code that returns -1 which falsely interpreted as status by the calling routine. In this case you will still run into the problem that success bit is set.

mwinkel-dev:
Hi @zack-vii -- Thanks for the explaining the history of SsINTERNAL.   I will edit this PR's proposed comments accordingly.

WhoBrokeTheBuild (approved):

PR #2620, opened 2023-09-11
Allow APD data be accessed via thin client via Connection.get()

GabrieleManduchi:
Implemented by explicitly sending over mdsip the serialized version of the expression evaluation result and deserializing it locally

### DIFF
--- a/include/mdsobjects.h
+++ b/include/mdsobjects.h
@@ -4459,6 +4459,20 @@ namespace MDSplus
     virtual void dataReceived(Data *samples, Data *times, int shot = 0) = 0;
   };
 
+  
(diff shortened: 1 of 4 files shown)


GabrieleManduchi:
retest this please

joshStillerman (commented):
It seems odd to me that the code is connecting to localhost:8001.  Please explain.
-Josh

GabrieleManduchi:
It is just the main method of a the Connection Java class. Main methods for internal classes are  used for tests. The main method of Connection will never be called, outside tests.

joshStillerman (approved):

PR #2621, opened 2023-09-12
Fix:invalid tree context in Data::execute

GabrieleManduchi:
The Tree was not passed tothe final evaluation of the compilation result in MDSplus::executeWithArgs()


AndreaRigoni (approved):
ok

PR #2622, opened 2023-09-19
Gm fix devices

GabrieleManduchi:
Added devices, labVIEW fix


AndreaRigoni (approved):
Ok, only RFX devices involved.

PR #2626, opened 2023-09-26
Fix: 2625 - start connection IDs at 1 (not zero) 

### LINKED ISSUE #2625: IDL - mdsisclient can't check against socket=0 causing further issues when set_database claims !mdsdb_socket=0
**Affiliation**

DIII-D

-----

**Version(s) Affected**

Current / Alpha Branch

-----

**Platform**

Linux (RHEL8), IDL API (original bug), Python API (introduced bug)

-----

**Describe the bug**

If `set_database` is called with the intention to proxy through a mdshost, then multiple issues occur if `set_database` receives the first socket id inside the IDL session (i.e. `!MDSDB_SOCKET=0`).

This appears to occur because:
- MDSplus sockets allow for a socket id=0.
- Multiple IDL functions use `mdsisclient` to check a socket passed via keyword.
- The `mdsisclient` function then relies on `keyword_set(socket)`.
- The IDL `keyword_set` function returns `false` if defined and the value is zero.
- The `mdsisclient` function does fall through when `keyword_set(socket)` returns `false`, but... only checks against `!MDS_SOCKET`.

Thus weird failure states happen when `set_database` claims socket id=0, resulting in `mdsisclient` being unable to confirm the socket via `keyword_set`, and then ultimately `dsql` and `mdsvalue` don't know how and/or where to execute SQL calls through TDI.

-----

**Examples**

Note that my .sybase_login file for these examples define the mdshost='atlas-el8.gat.com', so we know whether a successful dsql call gets evaluated via the intended proxy (i.e. atlas-el8,gat.com), or via localhost (i.e. omega01.gat.com), or not at all.

1) The mdsip connection to the mdshost gets opened but the dsql still executes from the localhost:
```
[flanagan@omega01 ~]$ idl
IDL> set_database,'d3drdb'
IDL> help,!MDS_SOCKET,!MDSDB_SOCKET
<Expression>    INT       =       -1
<Expression>    INT       =        0
IDL> n = dsql('select host_name()',val) & print,val
omega01.gat.com
```

2) A subsequent mdsconnect to any server will prevent dsql from being able to execute:
```
[flanagan@omega01 ~]$ idl
IDL> set_database,'d3drdb'
IDL> mdsconnect,'atlas.gat.com'
IDL> help,!MDS_SOCKET,!MDSDB_SOCKET
<Expression>    INT       =        1
<Expression>    INT       =        0
IDL> n = dsql('select host_name()',val) & print,val
% DSQL: SET_DATABASE must preceed any DSQL calls
% PRINT: Variable is undefined: VAL.
% Execution halted at: $MAIN$
```

3) Until another set_database call claims a different non-zero socket id:
```
[flanagan@omega01 ~]$ idl
IDL> set_database,'d3drdb'
IDL> mdsconnect,'atlas.gat.com'
IDL> set_database,'d3drdb'
IDL> help,!MDS_SOCKET,!MDSDB_SOCKET
<Expression>    INT       =        1
<Expression>    INT       =        2
IDL> n = dsql('select host_name()',val) & print,val
atlas-el8.gat.com
```

-----

**Discussion**

I tried fixing the logic in mdsisclient but it exposed other technical debt that I'm not eager to delve into.  

Instead, I'm wondering if this scenario can be triaged by disallowing mdsplus sockets with id=0?  

Would starting at socket id=1 be a simpler solution?

-----

mwinkel-dev:
This fixes the IDL keyword_set() issue that arises when socket IDs (aka connection IDs) start at zero.   The fix passes the entire MDSplus test suite.   And also passed manual testing, as shown below.

However, there are additional refinements that will likely be made in the coming days.   Such as removing obsolete code and improving comments.

**NOTE:**  Using `print` or `help` to examine the IDL system variables, !MDSDB_SOCKET and/or !MDS_SOCKET, occasionally has an odd interaction with the `dsql()` function -- causes it to fail evaluation of the query.   Will investigate the `dsql.pro` file in the coming days.

#2625 Testing
```
IDL> ; Test of proposed fix for Issue #2625
IDL> set_database, 'logbook'
% Compiled module: SET_DATABASE.
% Compiled module: MDSVALUE.
% Compiled module: MDSCHECKARG.
% Compiled module: MDSISCLIENT.
% Compiled module: DSQL.
IDL> print, !MDSDB_SOCKET
       1
IDL> n = dsql('select host_name()', val) & print, val
% Compiled module: EVALUATE.
some-server
IDL> 
IDL> 
IDL> mdsconnect, 'some-server'
% Compiled module: MDSDISCONNECT.
IDL> print, !MDS_SOCKET  
       2
IDL> n = dsql('select host_name()', val) & print, val
some-server
IDL> print, !MDSDB_SOCKET
       1
IDL> n = dsql('select host_name()', val) & print, val
some-server
IDL> 
IDL> 
IDL> set_database, 'logbook'
IDL> print, !MDSDB_SOCKET                            
       3
IDL> n = dsql('select host_name()', val) & print, val
some-server
IDL> 
```

#2580 Testing
```
IDL> ; Test to ensure Issue #2580 is still fixed
IDL> mdsconnect, 'some_server'
% Compiled module: MDSCONNECT.
% Compiled module: MDSDISCONNECT.
IDL>      
IDL> print, !MDS_SOCKET
       1
IDL> 
```

### DIFF
--- a/mdstcpip/mdsipshr/Connections.c
+++ b/mdstcpip/mdsipshr/Connections.c
@@ -465,13 +465,14 @@ static inline int authorize_client(Connection *c, char *username)
 int AddConnection(Connection *c)
 {
   MDSIPTHREADSTATIC_INIT;
-  static int id = INVALID_CONNECTION_ID;
+  
+  // Connection IDs are issued in sequence.
+  // Issue #2625 requires connection IDs starting at 1 to avoid IDL issues,
+  // so initialize this static ID to zero.
+  static int id = INVALID_CONNECTION_ID + 1;
   static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
   pthread_mutex_lock(&lock);
-  do
-  {
-    id++; // find next free id
-  } while (id == INVALID_CONNECTION_ID && _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
+  id++;
   c->id = id;
   pthread_mutex_unlock(&lock);
   c->state |= CON_INLIST;
--- a/mdstcpip/mdsipshr/Connections.c
+++ b/mdstcpip/mdsipshr/Connections.c
@@ -467,12 +467,15 @@ int AddConnection(Connection *c)
   MDSIPTHREADSTATIC_INIT;
   
   // Connection IDs are issued in sequence.
-  // Issue #2625 requires connection IDs starting at 1 to avoid IDL issues,
-  // so initialize this static ID to zero.
-  static int id = INVALID_CONNECTION_ID + 1;
+  // Issue #2625 requires connection IDs starting at 1 to avoid IDL issues.
+  // Use of _FindConnection() is an integrity check of the data structures.
+  static int id = INVALID_CONNECTION_ID;
   static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
   pthread_mutex_lock(&lock);
-  id++;
+  do
+  {
+    id++;
+  } while ((id == INVALID_CONNECTION_ID) || (id == 0) || _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
   c->id = id;
   pthread_mutex_unlock(&lock);
   c->state |= CON_INLIST;
--- a/idl/mdsconnect.pro
+++ b/idl/mdsconnect.pro
@@ -6,7 +6,8 @@ endcase
 end
 
 function sockmin
-  if !version.os eq 'linux' then return, 0 else return, 11-(!version.os eq 'MacOS')
+  ; Linux connection IDs start at 1 (see Issue #2625 for details).
+  if !version.os eq 'linux' then return, 1 else return, 11-(!version.os eq 'MacOS')
 end
 
 function mds$socket,quiet=quiet,status=status,socket=socket
--- a/idl/mdsconnect.pro
+++ b/idl/mdsconnect.pro
@@ -6,7 +6,7 @@ endcase
 end
 
 function sockmin
-  ; Linux connection IDs start at 1 (see Issue #2625 for details).
+  ; Connection IDs start at 1 (see Issue #2625 for details).
   if !version.os eq 'linux' then return, 1 else return, 11-(!version.os eq 'MacOS')
 end
 


mdstcpip/mdsipshr/Connections.c
```diff
@@ -465,13 +465,14 @@ static inline int authorize_client(Connection *c, char *username)
...
+  // Connection IDs are issued in sequence.
+  // Issue #2625 requires connection IDs starting at 1 to avoid IDL issues,
+  // so initialize this static ID to zero.
+  static int id = INVALID_CONNECTION_ID + 1;
   static pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER;
   pthread_mutex_lock(&lock);
-  do
-  {
-    id++; // find next free id
-  } while (id == INVALID_CONNECTION_ID && _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
```
WhoBrokeTheBuild, line 474: I'm still worried about the unintended side effects of removing this, why was this loop removed again?

mwinkel-dev: That is indeed a reasonable worry.   Which is why this change needs scrutiny.   If there are holes in the following argument (and even if not), we can restore the loop.

- Note that `AddConnection()` defines the `id` as a static, thus it maintains its value across calls of this function.   That is an important fact when analyzing the original code.

- The original code, sets `id = INVALID_CONNECTION_ID` (i.e., -1) at compile time.
- And on the first call, the do-while loop increments `id`, so it is now 0.
- Thus the `while (id == INVALID_CONNECTION_ID && ...` will fail on the very first call.
- And it also fails on all subsequent calls (because `id` continually increases)
- And because C uses shortcut evaluation of `&&`, the `_FindConnection()` function is never called
- Thus the entire loop is just an overly complicated way of executing the statement `id++` once per call.
- Meaning that every time a new connection is added, it is just issued the next sequence number.   New connections never re-use an old connection number.
- Which is why the proposed fix tosses the do-while loop and replaces it with `id++`.

One way to determine if there are holes in this argument is to run the code in the debugger.   If it ever calls `_FindConnection()` or if it ever iterates on the loop, then the above reasoning is flawed.

A conjecture (likely wrong) is that years ago, connections were maintained in an array.   In which case, there would need to be a loop to scan through the array to find empty slots (i.e., connections no longer in use that can now be reused.).  Eventually though, the array was replaced by the current scheme which is a linked list of connection structures.   And the looping was moved into the `_FindConnection()` routine -- which is used by more than `AddConnection()`.  The net result is the current code which is just a complicated way to execute `id++`.

However, there is always the risk of an unintended side effect when tinkering with old code.   So if it is safer, we can retain the original do-while loop, and the fix should still work fine.   Because the crux of the fix is the `+ 1` in the following line.

```
static int id = INVALID_CONNECTION_ID + 1;
```

If we do retain the original loop, then we should add a comment explaining that it is just an elaborate way to issue ID numbers sequentially, as that comment will be an aid to future maintenance developers.

zack-vii: the point of the loop was rather to allow the id to overrun. and then check if the old id is still in use.. the condition should use a `||` instead of `&&` so it would skip invalid id as well as ids still in use. if you want id 0 to be invalid (i am struggeling with the motivation to do it here) you would need to skip it here as well or redefine the INVALID_CONNECTION_ID.  possobly one would restrict the valid range to positive integers. but as fare are the original implementation is concerned, it is only an id similar to a fileid.

mwinkel-dev: Hi @zack-vii -- Thanks for the explanation.   Changing the `&&` to `||` makes a big difference.   In that case, the do-while loop is useful as it provides an additional integrity check on the linked list of connection structures.   

In theory, the static `id` and the linked list should never be out of sync, and thus the proposed fix without the loop should be reliable.  (Indeed, the proposed fix is equivalent to the original code with the `&&` bug in the do-while loop.)

However, it is good defensive programming practice to assume that data structures can be corrupted.   And if that did occur, then it would be a serious problem if the same id was used with multiple structures in the linked list.   Using the do-while loop and `_FindConnection()` is thus a useful integrity check.

The reason that zero is no longer used for the connection id is because of an oddity of the IDL language.   In IDL, functions and procedures can be given optional "keyword" arguments, such as `socket=socket`.   And there is a related IDL function, `keyword_set()` that can be used to test if a keyword argument is present and has a non-zero value.   Stated another way, when the connection ID is zero, then `keyword_set(socket)` fails, and the IDL code that is in MDSplus doesn't behave as expected.   We evaluated fixes to just the IDL code, but they were complicated and involved a lot of code changes (e.g., slipping in a shim routine everywhere the connection ID is used).   The simplest fix is simply to ensure that connection IDs are integers greater than zero.

I will restore the loop to this PR, but with the changes you suggested: using `||` and also skipping zero.

mwinkel-dev: Just added the changes that @zack-vii suggested.

Passes a Docker build with the `--test` option.   Also passes all the manual testing shown above in this PR (i.e., connection IDs starting at 1 and issued in sequence).

Please re-review this PR.

mwinkel-dev: Hi @zack-vii -- I forgot to explain why `INVALID_CONNECTION_ID` was left as `-1`.   We would indeed like to redefine it as zero.   Unfortunately, a scan of the entire MDSplus source code reveals a few instances where a connection id variable is being compared to the `-1` literal.   So, it was deemed less risky to leave `INVALID_CONNECTION_ID` as is, and just modify `AddConnection()` to skip over zero.

This is a "blocking" bug for DIII-D, so we want the most robust fix (with the least risk) as soon as possible.   Hence the approach taken with this PR.

WhoBrokeTheBuild (approved):

PR #2627, opened 2023-09-28
Fix: Revert #2626 - start connection IDs at 1 (not zero) 

WhoBrokeTheBuild:
This reverts commit 389ed8679a61c6907be9d3cf0b4ae0cb72bcd57f.

### DIFF
--- a/idl/mdsconnect.pro
+++ b/idl/mdsconnect.pro
@@ -6,8 +6,7 @@ endcase
 end
 
 function sockmin
-  ; Connection IDs start at 1 (see Issue #2625 for details).
-  if !version.os eq 'linux' then return, 1 else return, 11-(!version.os eq 'MacOS')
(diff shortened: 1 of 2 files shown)


mwinkel-dev:
Although PR #2626 fixes the issue if installed on a server, it doesn't work when a site has multiple servers: some with the fix and some without.   Thus the PR needs to be reverted.

Because old servers are still handing out zero as a valid connection ID, the fix for Issue #2625 has to instead be made in the IDL code.   A strategy has been identified and will be investigated ASAP (because this is a "blocking" issued for DIII-D).

Note that a new PR should be created to implement the fix that @zack-vii suggested, namely replacing the `&&` with `||` in the `AddConnection()` do-while loop, so that the `_FindConnection()` integrity check is used.

mwinkel-dev (approved):
See previous comment by mwinkel-dev.

PR #2628, opened 2023-09-29
Fix: 2625 v2 - only change IDL for socket 0 issue

### LINKED ISSUE #2625: IDL - mdsisclient can't check against socket=0 causing further issues when set_database claims !mdsdb_socket=0

mwinkel-dev:
Replaces PR #2626 which did not work at sites that have a mixture of old and new software on their servers.

This version changes the IDL code so that zero is accepted as a valid connection ID.

### DIFF
--- a/mdstcpip/mdsipshr/Connections.c
+++ b/mdstcpip/mdsipshr/Connections.c
@@ -471,7 +471,7 @@ int AddConnection(Connection *c)
   do
   {
     id++; // find next free id
-  } while (id == INVALID_CONNECTION_ID && _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
+  } while ((id == INVALID_CONNECTION_ID) || _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
   c->id = id;
   pthread_mutex_unlock(&lock);
   c->state |= CON_INLIST;
--- a/idl/dsql.pro
+++ b/idl/dsql.pro
@@ -230,7 +230,8 @@ if (debug) then $
 ; now execute the query
 ; !!!  if it is a string it is a database error
 ;
-count = mdsvalue(expr, socket=!MDSDB_SOCKET)
+socket_var = !MDSDB_SOCKET
+count = mdsvalue(expr, socket=socket_var)
 
 sz = size(count)
 if (sz(n_elements(sz)-2) eq 7) then begin
@@ -254,7 +255,8 @@ endif else begin
         if (debug) then $
           print, 'Working on arg ', i
         arg = 'a'+string(i, format="(I3.3)")
-        val = MDSVALUE("_"+arg,socket=!MDSDB_SOCKET,/QUIET)
+        socket_var = !MDSDB_SOCKET
+        val = MDSVALUE("_"+arg,socket=socket_var,/QUIET)
         ok = evaluate(arg + ' = val')
         if (ok eq "OK") then status = 1L else status = 0L
         if (debug) then $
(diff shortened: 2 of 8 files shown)


mdstcpip/mdsipshr/Connections.c
```diff
@@ -471,7 +471,7 @@ int AddConnection(Connection *c)
   do
   {
     id++; // find next free id
-  } while (id == INVALID_CONNECTION_ID && _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
+  } while ((id == INVALID_CONNECTION_ID) || _FindConnection(id, NULL, MDSIPTHREADSTATIC_VAR));
```
WhoBrokeTheBuild, line 474: Just for tracking, this is fixed re: Timo's comments:
https://github.com/MDSplus/mdsplus/pull/2626#discussion_r1338879619

WhoBrokeTheBuild (approved):

PR #2630, opened 2023-10-03
Revert #2628 v2 - only change IDL for socket 0 issue

WhoBrokeTheBuild:
This reverts commit 443018ead5ae95bb74e89da924d795de5ab449a7.

### DIFF
--- a/idl/dsql.pro
+++ b/idl/dsql.pro
@@ -230,8 +230,7 @@ if (debug) then $
 ; now execute the query
 ; !!!  if it is a string it is a database error
 ;
-socket_var = !MDSDB_SOCKET
-count = mdsvalue(expr, socket=socket_var)
+count = mdsvalue(expr, socket=!MDSDB_SOCKET)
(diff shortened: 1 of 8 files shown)


mwinkel-dev (approved):
The v2 fix was a Rube Goldberg approach that we weren't certain would work -- in particular, unsure if it would handle `mdsvalue()` correctly for all cases.   Sure enough, it failed for two reasons:

- First, mwinkel-dev (me) only tested for the scenarios in the original bug report: `set_database`, `mdsconnect`, and `mdsdisconnect`.   Thus did not catch some issues with `mdsopen`.
- Second, there is indeed at least one `mdsvalue` usage that cannot be handled with this approach.

Thus, this v2 fix also has to be reverted.

The only viable IDL-only fix is the one that Sean F. suggested, namely to do an extra `mdsconnect` (if needed) so that the connection ID (socket) for set_database will always be greater than zero.

PR #2635, opened 2023-10-06
Fix: 2625 v4 -- IDL socket 0 issue and partial revert of PR #2620

mwinkel-dev:
The IDL API has an issue with socket = 0 (aka connection ID) because IDL's `keyword_set()` function doesn't detect the presence of optional keyword arguments when they have a zero value.   This fix replaces the IDL function with a similar function that does handle zero.

And this PR also has these fixes:
- database proxy (established by `set_database`) is no longer trashed by a subsequent `mdsconnect`
- `mdsdisconnect` now processes the optional socket (if specified)
- partial rollback of PR #2620 that broke Python connections
- now does an integrity check when adding a new connection

However, it does not address the limit that there can be no more than 64 concurrent connections.

### DIFF
--- a/deploy/packaging/debian/idl.noarch
+++ b/deploy/packaging/debian/idl.noarch
@@ -36,6 +36,7 @@
 ./usr/local/mdsplus/idl/mdsevent.pro
 ./usr/local/mdsplus/idl/mdsgetmsg.pro
 ./usr/local/mdsplus/idl/mdsidlimage.pro
+./usr/local/mdsplus/idl/mds_keyword_set.pro
 ./usr/local/mdsplus/idl/mdsisclient.pro
 ./usr/local/mdsplus/idl/mdslogin.pro
 ./usr/local/mdsplus/idl/mdsopen.pro
--- a/deploy/packaging/redhat/idl.noarch
+++ b/deploy/packaging/redhat/idl.noarch
@@ -40,6 +40,7 @@
 ./usr/local/mdsplus/idl/mdsgetmsg.pro
 ./usr/local/mdsplus/idl/mdsidlimage.pro
 ./usr/local/mdsplus/idl/mdsisclient.pro
+./usr/local/mdsplus/idl/mds_keyword_set.pro
 ./usr/local/mdsplus/idl/mdslogin.pro
 ./usr/local/mdsplus/idl/mdsopen.pro
 ./usr/local/mdsplus/idl/mdsput.pro
--- a/idl/mds_keyword_set.pro
+++ b/idl/mds_keyword_set.pro
@@ -1,0 +1,9 @@
+; IDL's keyword_set() only detects non-zero optional keywords (see Issue 2625).
+; So replace it with this function that also handles zero.
+function mds_keyword_set,socket=socket
+  if (n_elements(socket) gt 0) then begin
+    return, 1
+  endif else begin
+    return, 0
+  endelse
+end
--- a/idl/mds_keyword_set.pro
+++ b/idl/mds_keyword_set.pro
@@ -1,3 +1,4 @@
+; This function detects if an optional keyword is present, regardless of value.
 ; IDL's keyword_set() only detects non-zero optional keywords (see Issue 2625).
 ; So replace it with this function that also handles zero.
 function mds_keyword_set,socket=socket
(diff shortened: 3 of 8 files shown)


joshStillerman (approved):

smithsp:
@joshStillerman When you approved this, what IDL tests had you run?  
@mwinkel-dev What IDL tests did you run on this branch before merging?

joshStillerman:
I will let Mark answer about more formal testing...  He ask Stephen and I to see if we could break it, and neither of us could.

mwinkel-dev:
Hi @smithsp -- Regarding testing of this PR, there are several answers.

**Regression Testing**
- I ran the standard MDSplus test suite on both Ubuntu 20.04 and RHEL8.
- Unfortunately, that test suite does not include IDL tests.   (More on that at the end of this post.)
- The MDSplus test suite does not presently incorporate scripts comparable to GA's tests.   (See end of this post.)

**Manual Testing**
- I manually ran IDL tests to cover all scenarios described by Sean's posts in Issue 2625.   I exercised these aspects of the IDL API: mdsconnect, mdsdisconnect, set_database, mdsdbdisconnect, dsql, mdsvalue, mdsopen and mdsclose.  And in particular focused on the database proxy issues.   My testing uncovered two additional IDL bugs: one with mdsdisconnect (fixed in this PR), and a limit on 64 total concurrent connections (not fixed in this PR).
- Stephen exercised the full IDL API (including mdsput) plus exercised the database proxy
- Josh also did many manual IDL tests with a focus on the database proxy

**Ubuntu vs RHEL8**
- I ran all of my IDL tests on Ubuntu, and am assuming that Josh and Stephen also used Ubuntu.
- I used to be employed by Research Systems Inc. and was part of the Dev / QA team for IDL, thus am well aware of the cross-platform testing that occurs with releases of IDL.   I am thus confident that the Ubuntu IDL results are applicable to RHEL8 too.
- Nonetheless, the MDSplus test environment should have a RHEL8 / IDL virtual machine or private Docker container so that our test environment is closer to GA's production environment.   So today, I am working on setting up this RHEL8 / IDL combination.

**Root Causes**
- One root cause is that I am a new developer on the MDSplus team.   So although I tested each of the four fixes manually, my testing was limited by my ignorance.   I tested the scenarios presented in Issue 2625, but was unaware that I was also supposed to test the entire IDL API (hence Fix v2 failed).   And I was also unaware that GA has several MDSplus servers and thus needed a solution that would enable servers with new software to connect to severs running old software (hence Fix v1 failed).
- Late evenings are another root cause.   Fix v3 failed because I was working on it at 2:30 a.m. while on the business trip to MIT PSFC, was tired, and thus forgot to run the `set_database, dsql, mdsconnect, dsql` test case.   [I implemented both approaches suggested by Sean, Fix v1 and Fix v3.   And have provided ~1.2 day turnaround (excluding weekends and travel days) when notified by GA that a bug still exists in IDL.]
- The final root cause is that the MDSplus CI / CD pipeline does not presently include IDL tests.   This deficiency will be addressed in the coming days and weeks, but requires us to set up some additional infrastructure at MIT.  (Restrictions imposed by IDL licensing are likely why IDL was overlooked by the CI / CD pipeline.   That plus the fact that the IDL code has been stable for years -- the IDL code fixed by this PR hasn't been touched in ~8 years.)

**Future Improvements**
- We need to add a "tag" test to the Python connection tests (to catch errors like the one introduced by PR 2620).
- IDL on RHEL8 should be added to our test environment.
- We will use MIT infrastructure so that we can run scripts comparable to those run by GA.
- And cross-version testing is needed.

mwinkel-dev:
At one time in my career, I was manager of a software QA group.   That experience taught me that where one bug is found, there are often several more bugs lurking.   And that boosting software quality often involves two steps: near-term and long-term.  

- Near-term goal is to hammer on bad code to uncover many bugs and thus quickly assess the quality of the code in question.   This stage of the process is mostly spent on manual testing.   And yields test cases that can be used for the next stage.
- Long-term goal is to establish CI / CD procedures and regression tests that ensure the module is adequately tested in future releases.

GA reported 1 bug with the IDL API; I found 6 more.   Details are described in Issue 2625.

Next up is to work on the long-term improvements, namely incorporating the test cases I have recently developed, plus using regressions scripts comparable to GA's scripts.   This will take time, but will improve the quality of the IDL API.

PR #2641, opened 2023-10-16
Fix:untested use cases

GabrieleManduchi:
Fixed some new,  untested  use cases for InputDevice

Changed the way the MARTe2 process is killed by stopMarte method


GabrieleManduchi:
retest this please

joshStillerman (approved):

PR #2642, opened 2023-10-17
Gm marte2 devices

GabrieleManduchi:
Better way of killing MARTe2 engine


heidthecamp (approved):
This looks good to me but it looks like this was updated with a merge before. Can you update this so we can pull?

GabrieleManduchi:
That change refers to an old update, and afterword we decided to keep all the MARTe2-MDSplus stuff (only affecting python devices in pydevices/RfxDevices) on a separate branch (gm-rfx) to avoid cluttering alpha (it is still experimental and rapidly evolving)
So you can safely discard this pull request. Thanks!

PR #2644, opened 2023-10-19
Fix: Change forking to simple in the systemd mdsip service

WhoBrokeTheBuild:
`Type=forking` is expecting the parent process to die, which I thought ours was doing but apparently isn't


joshStillerman (approved):
Told u so :)

Issue #2646, opened 2023-10-31
mdstcl - dir/full segfault


sflanagan:
**Affiliation**
DIII-D

------

**Version(s) Affected**
Current

EDIT: 
I played with different versions on a RHEL8 test server... 
- `alpha-7.131.2` - ok
- `alpha-7.133.1` - ok
- `alpha-7.134.1` - ok
- `alpha-7.134.9` - ok
- `alpha-7.134.12` - ok
- `alpha-7.134.13` - bugged
- `alpha-7.134.14` - bugged
- `alpha-7.135.1` - bugged
- `alpha-7.139.38` - bugged

------

**Platform**
RHEL8

------

**Describe the bug**
Segmentation fault / general protection fault... when calling `mdstcl dir/full`:
```
TCL> set tree nb /shot=196500
TCL> dir/full pinj
\NB::TOP
 :PINJ         tags: \PINJ
      Status: on,parent is on, usage signal
      compress on put
      contains node references (node ids only)
      Data inserted: 22-JUN-2023 15:47:10.63    Owner: uid=117
      Dtype: DTYPE_SIGNAL          Class: CLASS_R             Length: 524170 bytes
Total of 1 node.

TCL> set tree nb /shot=196500
TCL> dir/full pinj0
Segmentation fault (core dumped)

TCL> set tree nb /shot=196500
TCL> dir/full shot
Segmentation fault (core dumped)
```

The `/var/log/messages` log:
```
Oct 30 19:07:38 mdstest kernel: traps: mdsdcl[11312] general protection fault ip:7f9cbead724b sp:7ffca4d5ba08 error:0 in libc-2.28.so[7f9cbea09000+1bb000]
Oct 30 19:07:38 mdstest systemd[1]: Started Process Core Dump (PID 11489/UID 0).
Oct 30 19:07:38 mdstest systemd-coredump[11490]: Resource limits disable core dumping for process 11312 (mdsdcl).
Oct 30 19:07:38 mdstest systemd-coredump[11490]: Process 11312 (mdsdcl) of user 0 dumped core.
```

------

**Expected behavior**
For a baseline, here's the `mdstcl dir/full` response from a production mdsplus server (circa RHEL7, MDSplus 7.96.9):

```
TCL> dir/full pinj
\NB::TOP
 :PINJ         tags: \PINJ
      Status: on,parent is on, usage signal
      compress on put
      contains node references (node ids only)
[...]

PR #2647, opened 2023-10-31
Fix: dir /full segfault

joshStillerman:
Issue #2646

This bug was introduced when alternate compression methods were added commit 9829bf64dd7f3b9158f4d98172594fc523e4b746

That code allocated the NCI byte `spare2` to hold the index of the compression method.  It assumed that if it was not filled in it would be zero.  This turned out not to be the case for some nodes in some trees.

This PR adds a DBI that says wether or not to look at this byte.  The ** the default value of this DBI is ignore the compression method byte. ** To enable alternate compression methods, which have been seldom used (never?) set the DBI DbiALTERNATE_COMPRESSION to true.

- open the tree for edit
- call TreeSetDbi with DbiAlternate_COMPRESSION set to 1
- write the tree

The database characteristics (DBI) have been initialized to zero since [zero database characteristics](https://github.com/MDSplus/mdsplus/blame/263cc8bce631b291f17ae2c3e09164c55a067b93/treeshr/TreeOpen.c#L659) 22 years ago.

Example that sets the flag:
```
int main()
{
  int status;
  int length;
  int one = 1;
  DBI_ITM cmp_itm[2];
  const char *tree = "test";
  const int shot = 42;

  cmp_itm[0].buffer_length=1;
  cmp_itm[0].code = DbiALTERNATE_COMPRESSION;
  cmp_itm[0].pointer = (void *)&zero;
  cmp_itm[0].return_length_address = &length;
  cmp_itm[1].code = DbiEND_OF_LIST;

  status = TreeOpenEdit(tree, shot);
  if (!(status & 1)) {
    printf("open - status = %d\n", status);
    exit(1);
  }
  status = TreeSetDbi(cmp_itm);
  if (!(status & 1)) {
    printf("SetDBI - status = %d\n", status);
    exit(1);
  }
  status = TreeWriteTree("nb", 196500);
  if (!(status & 1)) {
    printf("Write - status = %d\n", status);
    exit(1);
  }
}
```

```
TCL> show alternate_compression
  Alternate Compression is disabled.
TCL> show alt
  Alternate Compression is disabled.
TCL> set alternate_compression on
TCL> show alt
  Alternate Compression is enabled.
TCL> set alt off
TCL> show alternate_compression
  Alternate Compression is disabled.
```

Next steps
- [x] Update Language APIs to expose this new attribute
   - [x] TDISHR
   - [x] PYTHON
   - [x] CPP
   - others?
- [ ] Then steps are to add regression test for this bug.
- [x] Then TCL verbs to set and get this database attribute.

This PR also removes some ifdefs for no longer supported AIX platform


treeshr/treeshrp.h
```diff
@@ -370,21 +370,13 @@ typedef struct tag_info
 typedef struct tree_header
 {
   char version; /* Version of tree file format */
-#ifdef _AIX
```
WhoBrokeTheBuild, line 373: Can you call this out in the PR message?

include/dbidef.h
```diff
@@ -9,24 +9,21 @@
...
+#define DbiINDEX  5                /* Index of tree to use for subsequent information requests */
+#define DbiNUMBER_OPENED 6         /* Number of trees currently open */
+#define DbiMAX_OPEN 7              /* Maximum number of tree allowed open at one time */
+#define DbiDEFAULT 8               /* NID of default node */
+#define DbiOPEN_READONLY 9         /* True if tree has been opened readonly */
+#define DbiVERSIONS_IN_MODEL 10    /* True if using versioning in model */
+#define DbiVERSIONS_IN_PULSE 11    /* True if using versioning in pulse files */
+#define DbiREADONLY 12             /* True if making tree readonly */
+#define DbiDISPATCH_TABLE 13       /* Tree dispatch table */
+#define DbiADVANCED_COMPRESSION 14 /* Set to true to enable extended compression methods */
```
WhoBrokeTheBuild: Could we call this custom compression? Or alternate compression?

WhoBrokeTheBuild (changes requested):
We need to update the various API interfaces to DBIs as well

Python:
https://github.com/MDSplus/mdsplus/blob/alpha/python/MDSplus/tree.py#L146

Java, C++, etc

treeshr/treeshrp.h
```diff
@@ -370,21 +370,13 @@ typedef struct tag_info
...
-#else
   unsigned char sort_children : 1; /* Sort children flag */
   unsigned char sort_members : 1;  /* Sort members  flag */
   unsigned char versions_in_model : 1;
   unsigned char versions_in_pulse : 1;
   unsigned char readonly : 1;
-  unsigned char : 3;
-#endif
+  unsigned char advanced_compression : 1;
+  unsigned char : 1;
```
WhoBrokeTheBuild: Shouldn't this be `: 2`?

treeshr/treeshrp.h
```diff
@@ -640,6 +632,7 @@ typedef struct pino_database
   unsigned modified : 1;      /* Flag indicating tree structure modified */
   unsigned setup_info : 1;    /* Flag indicating setup info is being added */
   unsigned remote : 1;        /* Flag indicating tree is on remote system */
+  unsigned advanced_compression : 1;
```
WhoBrokeTheBuild: Do we want to add padding bits here?

include/dbidef.h
joshStillerman: Done

treeshr/treeshrp.h
joshStillerman: done

treeshr/treeshrp.h
joshStillerman: done

treeshr/treeshrp.h
joshStillerman: done

treeshr/treeshrp.h
```diff
@@ -632,7 +632,8 @@ typedef struct pino_database
   unsigned modified : 1;      /* Flag indicating tree structure modified */
   unsigned setup_info : 1;    /* Flag indicating setup info is being added */
   unsigned remote : 1;        /* Flag indicating tree is on remote system */
-  unsigned advanced_compression : 1;
+  unsigned alternate_compression : 1;
+  unsigned char : 1;
```
WhoBrokeTheBuild: The type for the flags is `unsigned`, not `unsigned char` so this should be `unsigned : 25`, and I would check the `sizeof` before and after to verify we didn't accidentally inflate it

WhoBrokeTheBuild (changes requested):

treeshr/treeshrp.h
joshStillerman: Well - 
  Some of the bitfields in this file are `unsigned name : width` and some are `unsigned char name: width`.  I changed them all to the former.

WhoBrokeTheBuild: Those are completely different things, you can't change one to the other. `unsigned char` is 1 byte and `unsigned` is shorthand for `unsigned int` which is 4 bytes

include/dbidef.h
```diff
@@ -9,24 +9,21 @@
...
+#define DbiINDEX  5                /* Index of tree to use for subsequent information requests */
+#define DbiNUMBER_OPENED 6         /* Number of trees currently open */
+#define DbiMAX_OPEN 7              /* Maximum number of tree allowed open at one time */
+#define DbiDEFAULT 8               /* NID of default node */
+#define DbiOPEN_READONLY 9         /* True if tree has been opened readonly */
+#define DbiVERSIONS_IN_MODEL 10    /* True if using versioning in model */
+#define DbiVERSIONS_IN_PULSE 11    /* True if using versioning in pulse files */
+#define DbiREADONLY 12             /* True if making tree readonly */
+#define DbiDISPATCH_TABLE 13       /* Tree dispatch table */
+#define DbiALTERNATE_COMPRESSION 14 /* Set to true to enable extended compression methods */
```
WhoBrokeTheBuild, line 26: Can this comment match the ones in python?

python/MDSplus/tree.py
```diff
@@ -1132,6 +1135,21 @@ def versionsInPulseEnabled(self):
         """
         return self.versions_in_pulse
 
+    def setAlternateCompression(self, flag):
+        """Enable/Disable versions in pulse
```
WhoBrokeTheBuild: This comment needs to be updated

python/MDSplus/tree.py
```diff
@@ -1132,6 +1135,21 @@ def versionsInPulseEnabled(self):
...
+        @param flag: True or False. True enabled versions
+        @type flag: bool
+        @rtype: None
+        """
+        self.alternate_compression = bool(flag)
+
+    def alternateCompressionEnabled(self):
+        """Check to see if versions in the pulse are enabled
```
WhoBrokeTheBuild: This comment needs to be updated

treeshr/TreeGetDbi.c
```diff
@@ -52,7 +52,7 @@ int TreeGetDbi(struct dbi_itm *itmlst)
   }
 #define set_ret_char(val)                              \
   memset(lst->pointer, 0, (size_t)lst->buffer_length); \
-  *((char *)lst->pointer) = val
+  *((char *)lst->pointer) = val;
```
WhoBrokeTheBuild: Please remove this semicolon, it will cause warnings

WhoBrokeTheBuild (changes requested):

joshStillerman:
Added C++ accessors and setters, and regression tests for this

joshStillerman:
retest this please

WhoBrokeTheBuild (dismissed):

WhoBrokeTheBuild (approved):

Issue #2650, opened 2023-11-03
Enhance the IDL Test Harness


mwinkel-dev:
**Affiliation**
MIT PSFC (but based on feedback from GA)

The first version of the "IDL Test Harness" (see PR #2643) has 19 tests that cover many basic features of the IDL API.   But lacks a test of writing data to a tree.

The "IDL Test Harness" will be enhanced to include the following:
- add test cases to write data to trees
- add more read test cases
- for test cases based on Issues / PRs, add a comment that has a URL for the associated Issue or PR
- Provide testing across versions: (Initially for IDL, and eventually for all thin client APIs)
  - Test each client build against an old server (roughly 7.96.9)
  - Test an old client (roughy 7.96.9) against each server build

mwinkel-dev:
On 6-Nov-2023, completed the following:

- Added an IDL test case that reads various nodes (ASCII, numeric, signal), various paths (full, relative, tag), GETNCI metadata (with wildcard path, and normal path), and other (rank, ndims, dim_of, units_of).
- Updated the existing 19 IDL tests with comments that have the URL to the associated GitHub issues.
- Inserted visual separator comment lines so easier to see where one test case ends and another begins.
- Confirmed that the existing 19 IDL tests cover these features of the IDL API: connect, disconnect, open, close, a smidgen of TDI, connect to database, and execute SQL.

Now working on a prototype of the IDL write (aka store) tests, similar to the expanded read tests.

When done with the above, will submit to GitHub as a "work in progress" PR.

mwinkel-dev:
PR #2656 added write tests, so closing this issue.

PR #2656, opened 2023-11-30
Build: enhance IDL test harness by adding write tests and more read tests

mwinkel-dev:
Issue #2650

This is the first draft of the enhanced IDL test harness.   (It will surely be revised before being upgraded to a PR.)

Note that the write tests presently use a local tree on the build server.   After some infrastructure upgrades, the write tests will be changed to use a mdsip connection to write to a tree on a different server.

Comments on this work in progress will be appreciated.

### DIFF
--- a/idl/testing/run_tests.py
+++ b/idl/testing/run_tests.py
@@ -3,6 +3,7 @@
 import os
 import subprocess
 import argparse
+import tempfile
 
 # The default values are intended to be used from within the PSFC network
 # If you want to run these tests on your own infrastructure, provide the
@@ -237,25 +238,7 @@ args = parser.parse_args()
 # TODO: The default_tree_path is temporary; will change when switch to mdsip.
 #
 def build_write_tree(tree, shot):
-    import os
     import MDSplus as mds
-
-    test_dir = os.getenv('MDSPLUS_DIR') + '/idl/testing/'
-    os.environ['default_tree_path'] = test_dir  # required to write tree
-
-    tree_name = tree + '_' + str(shot)
-    tree_file = test_dir + tree_name
-
-    tc = tree_file + '.characteristics'
-    td = tree_file + '.datafile'
-    tt = tree_file + '.tree' 
-
-    if os.path.exists(tc):
-        os.remove(tc)
-    if os.path.exists(td):
-        os.remove(td)
-    if os.path.exists(tt):
-        os.remove(tt)
 
     t = mds.Tree(tree, shot, 'new')
 
@@ -276,8 +259,12 @@ def build_write_tree(tree, shot):
     t.close()
 
 
-build_write_tree(args.write_tree, args.write_shot)
+# Temporary directory for transient test scripts and artifacts
+TEST_DIR = tempfile.TemporaryDirectory(prefix='test_idl_', dir='/tmp')
+os.environ['default_tree_path'] = TEST_DIR.name
+os.environ['IDL_PATH'] = os.getenv('IDL_PATH') + ':' + TEST_DIR.name
 
+build_write_tree(args.write_tree, args.write_shot)
 
 all_tests_passed = True
 def idl_test(code, expected_output):
@@ -291,7 +278,8 @@ def idl_test(code, expected_output):
     # weird differences in the evaluation, so we use a test.pro file
 
     code = 'pro test\n' + code + '\nend'
-    open('test.pro', 'wt').write(code)
+    test_file = TEST_DIR.name + '/test.pro'
+    open(test_file, 'wt').write(code)
 
     expected_lines = [ line.strip() for line in expected_output.splitlines() ]
     expected_lines = list(filter(None, expected_lines))
--- a/idl/testing/run_tests.py
+++ b/idl/testing/run_tests.py
@@ -216,6 +216,18 @@ parser.add_argument(
     help='The value of evaluating the --units-of expression, ignoring leading/trailing whitespace'
 )
 
+parser.add_argument(
+    '--write-tree',
+    default='idl_tests',
+    help='Name of tree to create for exercising write functions of the API'
+)
+
+parser.add_argument(
+    '--write-shot',
+    default=100,
+    help='Shot number of tree to create for exercising write functions of the API'
+)
+
 args = parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


idl/testing/run_tests.py
```diff
@@ -1213,6 +1325,119 @@ def idl_test(code, expected_output):
...
+
+mdsopen, '{write_tree}', '{write_shot}'
+mdsput, 'A_TEXT', ' "string_a" '
+mdsput, 'B_NUM', '22'
+mdsput, 'C_SIGNAL', 'build_signal([10,-10,5,-5,0],[10.2,-10.2,5.4,-5.4,0.0], [0 .. 4])'
+mdsput, '\TOP.SUBTREE_1:D_TEXT', ' "string_d" '
+mdsput, 'SUBTREE_1.E_UNITS', 'build_with_units(55, "volts")'
+mdsset_def, 'SUBTREE_1'
+mdsput, '.-.SUBTREE_2:F_NUM', '66'
+mdsput, '\TAG_G', '77'
```
ModestMC, line 1572: Do you ever try reading the data from this tag (or just some tag)? I looked and I don't think I saw something but I could have missed it.

ModestMC: I believe in line 1412 you read the data stored here, but it's not referenced by the tag specifically.

ModestMC:
Overall, I think this is really solid progress and I'm personally pretty happy with it. I think there are some more cases to add, but this covers a lot of the key stuff. 

One thing I think would be smart/helpful is maybe some kind of comment block at the top that lists the particular features which are known to be tested and whatever keyword(s) will mostly get a ctrl+f to find the lines of the code relevant for that test. I am definitely going to forget which things get tested where in this file and I figure such a thing will be useful when you guys want to rip out pieces to stick in the upgraded testing framework later.

Finally, I will also note that most of our users interact with Atlas via mdsip, so we're very glad it's on you guys' radar and we'll be keeping an eye out for it.

idl/testing/run_tests.py
mwinkel-dev: Hi @ModestMC -- In the `IDL-read-various` function at line 1344 is a read using a tag.

You are correct that in the `IDL-write-various` function the value written by `\TAG_G` is checked at line 1412 to confirm that the write by tag succeeded.

mwinkel-dev:
Thanks for reviewing the enhanced IDL tests.   Goal of this round is to cover the basic features of the API.   After we get the additional infrastructure set up, we'll beef up the mdsip testing.   And eventually, could add even more coverage of the IDL API if needed.

mwinkel-dev:
Retest this please

mwinkel-dev:
retest this please

idl/testing/run_tests.py
mwinkel-dev: The line numbers changed significantly when switched to the `argparse` template.   The `IDL-read-various` read of a tag is now around line 1518.   And the `IDL-write-various` check of a tag is around line 1586.

mwinkel-dev:
Retest this please

ModestMC:
Not sure if it'll matter, but you may want to add some tests that check for case sensitivity (depending how IDL deals with string datatypes and what magic characters it cares about).

idl/testing/run_tests.py
```diff
@@ -58,8 +58,227 @@
...
+#---------------------------------------------------------------------------
+# Each IDL write test should start with a clean tree.
+# Write tests use a local tree, but eventually will be upgraded to use mdsip.
+# TODO: The default_tree_path is temporary; will change when switch to mdsip.
+#
+def build_write_tree(tree, shot):
+    import os
+    import MDSplus as mds
+
+    test_dir = os.getenv('MDSPLUS_DIR') + '/idl/testing/'
```
WhoBrokeTheBuild: We should instead use a `tempfile.TemporaryDirectory`, but be sure to call that once at the top, not once for each test

WhoBrokeTheBuild (changes requested):

idl/testing/run_tests.py
mwinkel-dev: That is a great suggestion!

mwinkel-dev: The `/idl/testing` directory is where we already create the transient `test.pro` files that `run_tests.py` executes.   Should those `test.pro` files also be created in a `tempfile.TemporaryDirectory`?

The local tree used for testing the write features of the IDL API is only ~3 KB in size.   Which is why I placed it in the same `idl/testing` directory as the transient `test.pro` files.

Regardless of which directory is used, I favor having the transient `test.pro` files and the dinky local "write" tree in the same directory.  But am open to using two different directories if that makes Jenkins more robust.

Also note that after we upgrade the infrastructure to do mdsip testing and cross-version testing, instead of a local tree we instead need to create the dinky "write" tree on the remote server. 

Regardless, I am now experimenting with `tempfile.TemporaryDirectory`.   But will hold off on making that commit until we decide whether to move the transient `test.pro` files too.

mwinkel-dev:
Hi @ModestMC,

Thanks for suggesting that I also add some tests for lowercase items (tags, node names, node paths).   That is a good idea.
https://github.com/MDSplus/mdsplus/pull/2656#issuecomment-1939788672

The goal is to gradually evolve this test into the mdsip / cross-version testing of the API for IDL.   That evolution will take several iterations.  My plan is to get this version merged in now as that will add tests of using the API to write to a tree.   In the next iteration, I will add some lowercase items.

idl/testing/run_tests.py
mwinkel-dev: A nice feature of using `tempfile.TemporaryDirectory` is that the directory will be empty on creation and will be deleted after the test completes.   This means that a few lines could be deleted from the existing code.

The same feature though also has a downside.   If the "write" test fails, might be nice to have the dinky tree left behind so that we can inspect it and figure out what just broke in MDSplus.

WhoBrokeTheBuild: Actually, yeah we should probably put the `.pro` there, it'll keep a truly temporary file out of the working directories

mwinkel-dev: A-OK, I will move the `test.pro` file to the `tempfile.TemporaryDirectory` too.

WhoBrokeTheBuild (approved):

PR #2669, opened 2024-01-02
Fix: baseline.fun needs ` ` around `..`

### LINKED ISSUE #2667: Intergrate TDI function cannot add baseline points
**Affiliation**
Tsinghua University

**Version(s) Affected**
7.132.0 amd64

**Platform**
Ubuntu 22.04   5.15.0-83-generic

**Describe the bug**
The integrate tdi function cannot add the baseline points argument. The function can work when only the signal expression argument is added.  However, when the base point is added to the expression, the jtraveser returns the misspelled error. 

The source code of integrate function shows that the base point argument is still supported as optional.

**Screenshots**
img_v3_026d_f8b13104-e4d0-4715-b1a2-20c5a5e49b7g
img_v3_026d_f8b13104-e4d0-4715-b1a2-20c5a5e49b7g
img_v3_026d_99458c1d-c4b8-4992-a604-bc5e70c6402g
image

joshStillerman:
The baseline.fun function which is called by integrate.fun uses a range that is not delimited by spaces.  Changes to TdiLex.l and TdiYacc.y have broken the syntax of `n..m` requiring spaces around the `..`.

Fixes issue #2667

### DIFF
--- a/tdi/math/signal/baseline.fun
+++ b/tdi/math/signal/baseline.fun
@@ -21,13 +21,7 @@ public fun baseline(in _sig, optional in _start, optional in _end)
 
 		_sum = 0.;
 		
-		/*
-		for(_i = 0; _i < _points; _i++)
-			_sum = _sum + _y[_i];
-		*/
-		
-		_sum = sum(_y[0.._points-1]);
-		_offset = _sum / _points;
+		_sum = sum(_y[0 .. _points-1]);
 			
 		_offset = _sum / _points;
 	}


mwinkel-dev (approved):
The proposed fix is fine.

As Stephen suggested on 2-Jan-2024, please manually run the test suite.  (Until the build server maintenance work is completed, we will have to manually run the test suite.).  Also, please add a note to this PR when the fix passes the tests.

PR #2676, opened 2024-01-04
Fix: tdic uses correct env var and path to launch tcl

### LINKED ISSUE #2663: TDIC fails to launch TCL
**Affiliation**
MIT PSFC

**Version(s) Affected**
Alpha, but surely in all versions

**Platform**
Ubuntu 20, but surely on all platforms.

**Describe the bug**
The `tdic` utility claims to be able to pass commands to `tcl`, but that feature doesn't work.   As shown below, `tdic` can't find `tcl`.

```
$ tdic
TDI> help
TDI(C) reference (October 1999)
      help func <help function>
      help func* <lists 1st line of each help>
      use func* <lists lines in comment starting with [call:]>
      type func(*) <type/p all the functions requested>
      tcl <TCL command>
      open <shot#>
      !<shell command>
      man [TDI intrinsic]
      exit (guess)
      ; behaves as in MatLab
TDI> tcl
sh: 1: /bin/tcl: not found
TDI> 
```

**To Reproduce**
1. See the code block above.

**Expected behavior**
Should be able to run `tcl` commands from `tdic`.

**Screenshots**
n/a

**Additional context**
Was found with mdsplus installed in a non-standard directory.   However, should also arise when installed in `/usr/local/mdsplus`.

Root cause is that the `tdic` source code is using the wrong environment variable and path.

mwinkel-dev:
Fix for Issue #2663.

### DIFF
--- a/tdic/tdic.c
+++ b/tdic/tdic.c
@@ -315,7 +315,7 @@ int main(int argc, char **argv)
       }
       else if (!strncmp(line_in, "tcl", 3))
       {
-        strcpy(line_in, "$MDSPLUS/bin/tcl"); /* move the text to the left */
+        strcpy(line_in, "$MDSPLUS_DIR/bin/mdstcl"); /* move the text to the left */
         system(line_in);
       }
 #ifdef NEVER


WhoBrokeTheBuild (approved):

mwinkel-dev:
Yeah, I'm not wild about the `$MDSPLUS_DIR` either -- it is just as clunky as the original `$MDSPLUS`.   However, the `setup.sh` file doesn't define the original environment variable, hence this change.

PR #2678, opened 2024-01-08
Fix test event port, collect and import test results into Jenkins

WhoBrokeTheBuild:
The event port is now computed from OSList.indexOf(OS) instead of $EXECUTOR_NUMBER Artifacts will now archive regardless of the build result


joshStillerman (dismissed):

mwinkel-dev (approved):
Looks fine.

PR #2679, opened 2024-01-09
Work on adding the IDL tests to the Jenkinsfile

WhoBrokeTheBuild:
Refactor idl/testing/run_tests.py to use argparse instead of env variables Add defaults so we don't have to specify them all in the Jenkinsfile Leave a stub for adding MATLAB later


mwinkel-dev (dismissed):
The use of argparse is a nice improvement over the TEST_* environment variables.   Good that this single file contains all settings, and thus no need for an auxiliary file to set environment variables.

mwinkel-dev (dismissed):

mwinkel-dev (dismissed):

mwinkel-dev (approved):

PR #2692, opened 2024-01-19
Build: Treat unprefixed commits as "Fix:"

WhoBrokeTheBuild:
This will change how version numbers are incremented.
Previously, commits without one of the commit message prefixes weren't considered when incrementing the version in `get_new_version.py`. Now, commits without a prefix will increment the PATCH portion of the version.


mwinkel-dev (approved):
Nice that the commented print() has been remove.   And the change to bump the `PATCH` field of the version (for unprefixed commits) is simple and clean.

PR #2694, opened 2024-01-19
Build: Give the tgz files absolute paths

WhoBrokeTheBuild:


mwinkel-dev (approved):
Note that `WORKSPACE` is a Groovy variable; it is not a shell environment variable.   Thus no need for a leading `$`.

This change to absolute paths is good.

PR #2695, opened 2024-01-19
Build: Actually build in /mdsplus

WhoBrokeTheBuild:
This will not actually deploy quite yet, we need to setup an rsync as well


mwinkel-dev (approved):
Change discussed via Slack.   Looks good.

PR #2529, opened 2023-02-02
change python setup to work with newer setuptools

dgarnier:
When installing MDSplus with newer setuptools, it refuses to continue do to multiple subdirectories.  More at Issue
https://github.com/MDSplus/mdsplus/issues/2528

PR #2698, opened 2024-01-31
Switch to pyproject.toml and stub setup.py for better pip / setuptools compatibility

dgarnier:
This gives me consistently good results for installing either with or without -e on ubuntu with newer versions of pip and setuptools.  It is also possible to switch to a different build system like poetry or Hatch, but setuptools seems to have the minimum requirments met.

This is probably prefered over #2618 and #2529.


dgarnier:
<s>Unfortunately... this does not work with python2.7. </s>  Actually, it now works with python2.7 and for old versions of `setuptools <= 60`

joshStillerman:
retest this please

dgarnier:
I can't see why the commit isn't building.  https://mdsplus-jenkins.psfc.mit.edu/job/MDSplus/job/PR-2698/ doesn't load for me.  I was able to log into Jenkins, but I still don't see it.

joshStillerman:
retest this please

dgarnier (commented):
Woot!  Found a way to get python2.7 to work and also _old_ (< 60.0.0) setuptools to work.

python/MDSplus/pyproject.toml
```diff
@@ -0,0 +1,57 @@
+[build-system]
+    requires = ["setuptools>=44.1.1", "wheel"] # python2.7 support
+    build-backend = "setuptools.build_meta"
+
+[project]
+    name='MDSplus'
+    authors = [
+        {name = "MDSplus Development Team", email = "twf@www.mdsplus.org"},
```
WhoBrokeTheBuild: We should probably change the email here, maybe mdsplusadmin@psfc.mit.edu ?

python/MDSplus/setup.py
```diff
@@ -23,80 +23,91 @@
...
         long_description=(
             "This module provides all of the functionality of MDSplus TDI natively in python.\n"
             "All of the MDSplus data types such as signal are represented as python classes.\n"
         ),
-        author='MDSplus Development Team',
-        author_email='twf@www.mdsplus.org',
-        url='http://www.mdsplus.org/',
-        license='MIT',
+        author="MDSplus Development Team",
+        author_email="twf@www.mdsplus.org",
```
WhoBrokeTheBuild: Same here

WhoBrokeTheBuild (dismissed):

python/MDSplus/pyproject.toml
dgarnier: OK.. done.

python/MDSplus/setup.py
dgarnier: done.

WhoBrokeTheBuild (approved):

mwinkel-dev:
Merged the approved PR as per a request from @dgarnier.

PR #2702, opened 2024-02-06
Build: Actually publish to the real directory

WhoBrokeTheBuild:
This will publish to the local copy of the MDSplus repository mirror, an additional rsync will be needed to publish everything to the webserver


mwinkel-dev (approved):
A-OK

PR #2703, opened 2024-02-07
Fix: corrected typo that caused MdsPut2() to clear all flags

### LINKED ISSUE #2658: MDSplus .data file gets very large
**Affiliation**
General Atomics

**Version(s) Affected**
7.132-0.el7

**Platform**
CentOS Linux 7

**Describe the bug**
A development branch of PCS code outputs data to PTDATA (GA in house database) and MDSplus. The data is assumably identical. We ran 14 shots and the average PTDATA file size is 142 MB while MDSplus size is 928 MB

**To Reproduce**
1. Run a development branch of PCS on Saga or Iris computer (at GA)
2. Cycle PCS 
3. Read the data in /fusion/d3d/d3share/d3data/d3dpcs_965507.datafile

**Expected behavior**
We expect that the file size of MDSplus is on par with PTDATA especially since compression is automatically turned on for MDS

**Screenshots**

**Additional context**
Data can also be seen from Omega cluster

mwinkel-dev:
Fixes Issue #2658

The `mds_put2_vargs()` function is a clone of the original `mds_put_vargs()`, with some minor edits.   

Unfortunately, the edits introduced a typo in `mds_put2_vargs()`.   The fix involved undoing the typo and using the same construct seen in `mds_put_vargs()` at the following source line.

[MdsLib.c, line 1054, mds_put_vargs()](https://github.com/MDSplus/mdsplus/blob/dae624da75b90a4f1fadae2ed5e285b4988ce79d/mdslib/MdsLib.c#L1054)

### DIFF
--- a/mdslib/MdsLib.c
+++ b/mdslib/MdsLib.c
@@ -1192,8 +1192,7 @@ static int mds_put2_vargs(va_list incrmtr, int connection, char *pathname,
       if (STATUS_OK)
       {
         if ((status =
-                 TreePutRecord(nid, (struct descriptor *)arglist[argidx - 2]),
-             0) &
+                 TreePutRecord(nid, (struct descriptor *)arglist[argidx - 2], 0)) &
             1)
         {
           TreeWait();
--- a/mdslib/MdsLib.c
+++ b/mdslib/MdsLib.c
@@ -1050,8 +1050,7 @@ static inline int mds_put_vargs(va_list incrmtr, int connection, char *pathname,
 
       if (STATUS_OK)
       {
-        if ((status = 
-                 TreePutRecord(nid, (struct descriptor *)arglist[argidx - 2], 0)) & 1)
+        if ((status = TreePutRecord(nid, (struct descriptor *)arglist[argidx - 2], 0)) & 1)
         {
           TreeWait();
         }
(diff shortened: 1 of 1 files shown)


mdslib/MdsLib.c
```diff
@@ -1192,8 +1192,7 @@ static int mds_put2_vargs(va_list incrmtr, int connection, char *pathname,
       if (STATUS_OK)
       {
         if ((status =
-                 TreePutRecord(nid, (struct descriptor *)arglist[argidx - 2]),
-             0) &
+                 TreePutRecord(nid, (struct descriptor *)arglist[argidx - 2], 0)) &
```
WhoBrokeTheBuild: I think this was mangled during the run of `clang-format`, maybe we should put it all back on one line while we're at it

WhoBrokeTheBuild (changes requested):

mdslib/MdsLib.c
mwinkel-dev: Yes, we could put the trailing `1)` back on this line.   I didn't do that simply because the original `mds_put_vargs()` was also mangled.   I figured that preserving the mangled structure made it easier to diff the two functions (as might be needed for future maintenance work).

I'm happy though to move the trailing `1` up to line 1195 if you wish.   Additional thoughts?

WhoBrokeTheBuild: If they're both mangled, I would unmangle them both, but either way is fine

mwinkel-dev: OK, I will unmangle them both.

mwinkel-dev: Hmm . . . Just occurred to me that you likely wanted `status = TreePutRecord . . . ` all on one line.   If so, let me know and I'll make another commit.

WhoBrokeTheBuild: Yeah, that would probably be best

mwinkel-dev: OK, will do.

WhoBrokeTheBuild (approved):

PR #2705, opened 2024-02-08
Fix: Release versioning was always 1 behind

WhoBrokeTheBuild:
Jenkins now re-runs bootstrap after calculating and tagging the new version
This also moves the --release/--publish=VERSION into a separate parameter called --version=VERSION
This now properly passes the BRANCH/RELEASE_VERSION into bootstrap for us to override the git info


mwinkel-dev (approved):
Nice to have the branch_name explicitly passed.   All looks good.

PR #2707, opened 2024-02-09
Fix: Restructure the Jenkinsfile to better support versioning

WhoBrokeTheBuild:
Move the Calculate Version step to the top
Call Bootstrap once with the version (or 0.0.0 for PRs) Instead of having Test Packaging/Release steps, just have one Release step


mwinkel-dev (approved):
Confirmed that the refactoring matches the description.   And that the refactoring merely reshuffled the existing code blocks (i.e., nothing accidentally dropped).

PR #2711, opened 2024-02-13
Build: configure PYTHONPATH for IDL test harness

mwinkel-dev:
Partial fix for Issue #2650.

The enhanced IDL test harness creates a tiny local tree that is used for ensuring that the IDL API can write to a tree.   The Python code that creates that tree, imports the MDSplus Python API, thus needs the PYTHONPATH environment variable configured accordingly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -175,7 +175,7 @@ pipeline {
                                 sh """
                                     set +x
                                     . \$MDSPLUS_DIR/setup.sh
-                                    export PYTHONPATH=$MDSPLUS_DIR/python/
+                                    export PYTHONPATH=${MDSPLUS_DIR}/python/
                                     set -x
                                     ./idl/testing/run_tests.py
                                 """
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -175,7 +175,7 @@ pipeline {
                                 sh """
                                     set +x
                                     . \$MDSPLUS_DIR/setup.sh
-                                    export PYTHONPATH=${MDSPLUS_DIR}/python/
+                                    export PYTHONPATH=\$MDSPLUS_DIR/python/
                                     set -x
                                     ./idl/testing/run_tests.py
                                 """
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -175,6 +175,7 @@ pipeline {
                                 sh """
                                     set +x
                                     . \$MDSPLUS_DIR/setup.sh
+                                    export PYTHONPATH=$MDSPLUS_DIR/python/
```
WhoBrokeTheBuild: This needs to be `\$MDSPLUS_DIR` to escape the Jenkinsfile

WhoBrokeTheBuild (changes requested):

Jenkinsfile
mwinkel-dev: Just noticed your comment.   Am assuming you mean there should be a backslash in front of the `$MDSPLUS_DIR`.   Correct?

```
export PYTHONPATH=\$MDSPLUS_DIR/python/
```

mwinkel-dev: Based on how the Jenkins file references other environment variables, I changed the line by adding braces.   I hope that is the correct syntax change for the line.

```
export PYTHONPATH=${MDSPLUS_DIR}/python/
```

mwinkel-dev: Just noticed that the initial comment was edited to clarify the change.   Means that my second guess was wrong.   So will change the line to use the first guess, `\$MDSPLUS_DIR`.

joshStillerman:
I can not tell from the above if it still needs my review
- was force push the right thing ?

mwinkel-dev:
Hi @joshStillerman,

Darren just merged in his PR, so I had to rebase this PR.  (When I clicked on GitHub's "Update with rebase" button, it automatically did the force push to my branch: mw-2650-idl-tests.)   

Yes, the review is still needed for this PR.   Line 178 has to be . . . 
```
export PYTHONPATH=\$MDSPLUS_DIR/python/
```
. . . as per Stephen's comment above.   (When I first edited the file, I was unaware that I needed to include the backslash.)

Thanks for looking at this.

WhoBrokeTheBuild (approved):

PR #2727, opened 2024-03-08
Build: Add 'Release' as a commit prefix

WhoBrokeTheBuild:
Using 'Release' will now use the exact version that commit is tagged with, instead of trying to bump the version based on the other commits Add comments to get_new_version.py


joshStillerman (approved):

mwinkel-dev (approved):
Simple change looks OK.   Also like the additional comments.

PR #2728, opened 2024-03-11
Fix: remove nix package text so matches alpha README.md

mwinkel-dev:
When alpha is eventually merged into stable, the `README.md` file for stable should be the same as that for alpha.   To achieve that, must omit the section about the "nix package manager" that is presently in stable's `README.md` file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,20 +63,6 @@ You can find the available operating systems that you could specify for the
     # ls <mdsplus-src-dir>/deploy/os/*.opts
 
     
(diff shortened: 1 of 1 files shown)


WhoBrokeTheBuild (approved):

mwinkel-dev:
The new Jenkins system is not yet checking the `stable` branch.   However, this change is only to a text file.   So as per Stephen, am doing a forced merge of this PR.

PR #2732, opened 2024-04-01
Docs: Improve documentation for getSegment* python wrappers

WhoBrokeTheBuild:
Add explanation and rename parameters for:
* getSegmentLimits
* getSegmentList

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -2305,6 +2305,12 @@ class TreeNode(_dat.TreeRef, _dat.Data):
         return dim.value
 
     def getSegmentLimits(self, idx):
+        """Return the start and end times of a given segment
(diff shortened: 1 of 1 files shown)


mwinkel-dev (approved):
The comments add clarity.  As does replacing `start` with `startTime`, and `end` with `endTime`.

santorofer:
Stephen, in the comments, would it make sense to say that the times are in seconds? Just to be explicit.

zack-vii:
> Stephen, in the comments, would it make sense to say that the times are in seconds? Just to be explicit.

it wpuld be a good place to mention a envoronment var that controlls if thebtimes are in sec (default) or nsec. There was an env (or some kind of switch at least) if i am not mistaken.

in fact they are as is but treated by xtreeshr as nsec if (u)int64 and as sec otherwise.

PR #2733, opened 2024-04-02
Build: bootstrap - export PYTHON for scripts like python/generate_tests

zack-vii:
If you do not have python or at least a link (python -> python3) bootstrap will fail as underlying scripts do not see the PYTHON environment. in particular python/generate_tests will try 'python' and fail.


WhoBrokeTheBuild (approved):

zack-vii:
can you merge this if the tests go nice ;) going to bed now

PR #2744, opened 2024-04-19
Fix: Improve error messaging when calling Setup Device in jTraverser

WhoBrokeTheBuild:
e.getMessage() sometimes returned null, but just e will always print something Add a printStackTrace() for InvocationTargetException exceptions to show the encapsulated error


mwinkel-dev:
Confirmed changes match description above.

Will approve this PR after it passes the Jenkins build test.

mwinkel-dev (approved):
Approving as per previous post.

PR #2745, opened 2024-04-23
Gm thin cpp

GabrieleManduchi:
Added segment support via thin client in mdsobjects/cpp. This implied also a few new TDI funs in addition to those already used for segment management.  The test procedure has been extended to exercise also Segment access via thin client.

Improved MdsFilter in order to properly work also with resampled signals via MinMax resampling (typical use case: jScope visualization of large signals)


GabrieleManduchi:
retest this please

Issue #2741, opened 2024-04-14
Some calling routines do not properly handle SsINTERNAL, C_ERROR, and/or FALSE returned by low-level routines


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
all

**Platform**
all

**Describe the bug**
The entire MDSplus code base uses different standards for error codes.
- on Linux, calls to the operating system functions usually return `-1` to indicate an error (i.e., access `errno` for the code)
- some MDSplus functions can return the macros `C_OK` (=0) and `C_ERROR` (= -1)
- others return `TRUE` (=1) or `FALSE`(=0)
- some return `B_TRUE` (=1) or `B_FALSE` (=0)
- and several return `SsINTERNAL` (= -1)
- plus many return MDSplus codes, such as `MDSplusSUCCESS` and `MDSplusERROR` (there are many more too)

Often the status returned by a function is immediately checked with one of the following "define" macros.

- `IS_OK`
- `STATUS_OK`
- `IS_NOT_OK`
- `STATUS_NOT_OK`

However, the above macros only work with MDSplus codes, which consist of a 32-bit integer containing 3 fields (facility ID, message ID, and severity code).  **Note that the low-order bit is important.  Success = 1 in the bit, Failure = 0 in the bit.**

The problems arise when non-MDSplus codes are passed to the macros.   Specifically, the -1 of `SsINTERNAL`, `C_ERROR` or an operating system error status.   **The value `-1` = `0xFFFFFFFF` has the low-order bit set, thus the macros will incorrectly treat `-1` as a flavor of success.** Which can cause error handling code to produce different results than what was  intended when the code was originally written.

This mix of error code conventions is a systemic source of errors in the entire code base.   It would probably be worthwhile to  at least make sure that all routines that return `SsINTERNAL` or `C_ERROR`, are in turn called by functions that properly detect those return values and then map them into the appropriate MDSplus error codes.

**To Reproduce**
This is an example that was spotted during the investigation of Issue #2731 and PR #2740.   The low-level routine, `send_bytes()` can return `SsINTERNAL`.   And potentially, it can ripple up the call stack all the way to the top level of `mdstcl`.  

The following links are to lines of code starting in `send_bytes()` and then climbing up the call stack.
[...]

PR #2746, opened 2024-04-23
Fix: improve mdstcl's error handling and add comments

mwinkel-dev:
Fixes some problems found during the investigation of Issue #2731.

### DIFF
--- a/mdstcpip/mdsipshr/ConnectToMds.c
+++ b/mdstcpip/mdsipshr/ConnectToMds.c
@@ -82,7 +82,7 @@ static int do_login(Connection *c)
   int status = SendMdsMsgC(c, msend, 0);
   int err;
   free(msend);
-  if (STATUS_NOT_OK)
+  if ((status == SsINTERNAL) || STATUS_NOT_OK)
   {
     perror("Error during login: send");
     err = C_ERROR;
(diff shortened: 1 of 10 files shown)


mdstcpip/mdsipshr/ConnectToMds.c
```diff
@@ -82,7 +82,7 @@ static int do_login(Connection *c)
   int status = SendMdsMsgC(c, msend, 0);
   int err;
   free(msend);
-  if (STATUS_NOT_OK)
+  if ((status == SsINTERNAL) || STATUS_NOT_OK)
```
WhoBrokeTheBuild, line 85: In the other changes, we rewrite `status` to `MDSplusERROR`, but in this one we just check for `SsINTERNAL` as well, is there a reason for the difference?

WhoBrokeTheBuild (approved):

mdstcpip/mdsipshr/ConnectToMds.c
mwinkel-dev: No, I don't have a good reason.   Other than the changes were made over several days, thus I was inconsistent.

The longer answer . . . 

I'm taking a cautious approach until we all reach agreement on the broader SsINTERNAL problem, Issue #2741. 

Timo tells me it is important to keep SsINTERNAL so that it ripples up the call stack.   However, most of the code I've looked at never checks for SsINTERNAL.   This is a systemic error throughout all of MDSplus.   This PR is just a targeted fix for the issues I saw while investigating the dispatcher's leaking sockets.

My preference would be to redefine SsINTERNAL so it is a valid MDSplus error code (i.e., set it to `-2` so that the low order bit is zero, meaning failure).   Timo and I have not reached agreement on whether that is feasible or not.

PR #2749, opened 2024-04-26
Build: Resolve linker error after updating the windows builder to Fedora 39

WhoBrokeTheBuild:
This appeared after updating the mdsplus/builder:windows docker image to Fedora 39, and Wine to 9.0 The newer libxml2 tried to link dynamically unless we explicitly set LIBXML_STATIC


mwinkel-dev (approved):
Thanks for adding the comment about `s2` being a pointer to the middle of the table!

PR #2755, opened 2024-04-30
Fix: MdsFilter for MinMax, TDI error in baseline_segmented.fun

GabrieleManduchi:
- Allow MdsFilter work properly also for signals resampled via MinMax resampling (jScope use case)
- Remove useless thread in jServer
- Fix TDI compile error in baseline_segmented.fun


mwinkel-dev (approved):
Confirmed that the new `checkSignal()` function handles the min-max sampling as per the comments.   And that the code refactoring looks good.

PR #2756, opened 2024-05-01
Build: Fix off-by-one versions produced by Jenkins

WhoBrokeTheBuild:
This fixes the bug where `--os=bootstrap` wasn't receiving the version from `--version=x.y.z` However, confusingly, this also changes the Jenkinsfile to not use that feature, and instead use `git tag` in order to embed the proper git information as well as the proper version information The `--os=bootstrap` and `--version` fix is still included just so that it doesn't break if someone else tries to use it

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,6 +41,7 @@ if (BRANCH_NAME == "stable") {
 }
 
 def new_version = '0.0.0';
+def new_tag = null;
 
 def release_file_list = [];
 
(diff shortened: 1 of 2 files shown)


mwinkel-dev:
Confirmed that the changes match the description given above.   Changing `tag` to `new_tag` is nice -- adds clarity to the `Jenkinsfile`.

Will approve after this PR passes the Jenkins build test.

mwinkel-dev (approved):
Approving as per previous post.

PR #2757, opened 2024-05-01
Build: Increase default test timeout to 1h

WhoBrokeTheBuild:
When the build server(s) are at capacity, it's not unreasonable for a test to take more than 10 seconds, which was the old default timeout
This sets the default to 1h, and removes the overrides in various tests

### DIFF
--- a/mdsobjects/cpp/testing/MdsConnectionTest.cpp
+++ b/mdsobjects/cpp/testing/MdsConnectionTest.cpp
@@ -139,7 +139,7 @@ void test_tree_open(const char *prot, const unsigned short port,
 int main(int argc, char *argv[])
 {
   int ipv6 = (argc > 1 && !strcmp(argv[1], "ipv6"));
-  TEST_TIMEOUT(30);
+  // TEST_TIMEOUT(30);
   setenv("t_connect_path", ".", 1);
 
 #ifdef _WIN32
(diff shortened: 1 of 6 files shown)


santorofer (approved):
Looks good.

mwinkel-dev (approved):
Setting the `TEST_DEFAULT_TIMEOUT` to an hour (3600 seconds) and applying it everywhere is a good idea while we are enhancing the build system.

However, that can also mask performance issues in MDSplus.   It is probable that for some tests we might indeed want to consider it a failure if it takes 5 minutes to execute.

As we increase test coverage in the coming months / years, we will likely want to revisit both the default timeout and some test specific timeouts.

mwinkel-dev:
I concur with everything in your post about the performance tests.

When we eventually get the isolated server set up and all the software infrastructure created to track performance metrics, we could flag any test that is N times slower than expected.   Say that N is set to 2.   Then a test that always completes in 10 seconds would be flagged as failing if it took 20 seconds.   And a test that normally takes 1 minute would be considered failing if it took 2 minutes.   Thus with the performance infrastructure, we could set tight specs for execution time on tests (if we wish).

In the meantime, timeouts are a crude and inadequate substitute.   They can easily generate both false positives and false negatives.   And because of the fluctuating load on the build servers, the timeouts would necessarily be set for looser limits on tests than if we had the performance infrastructure.   Even so, timeouts are the tool that is available now, and thus we might have to revisit the timeouts on a few tests in the coming months.

PR #2758, opened 2024-05-02
Build: Fix 'HEAD' in `show version` and tag error

WhoBrokeTheBuild:
Jenkins builds in a detached HEAD state, which caused bootstrap to use HEAD as the branch name
We pass --branch= to the bootstrap call in Jenkins, but $BRANCH wasn't being passed into the bootstrap docker container 
Also, attempts to build alpha versions with tags that already existed failed

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -128,7 +128,7 @@ pipeline {
                                         checkout scm;
 
                                         if (new_tag) {
-                                            sh "git tag ${new_tag}"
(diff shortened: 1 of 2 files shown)


mwinkel-dev (approved):
The changes are the same ones that we discussed yesterday.   Plus also used this review as an opportunity to study the entire `Jenkinsfile`.   It all makes sense and looks good, so approving this PR.

PR #2761, opened 2024-05-03
Build: Add package overwrites for ubuntu and debian

santorofer:
Overwrite sections for Ubuntu 24 and Debian Bookworm was added.

### DIFF
--- a/deploy/packaging/linux.xml
+++ b/deploy/packaging/linux.xml
@@ -836,6 +836,12 @@ rm -f /etc/ld.so.conf.d/mdsplus.conf 2&gt;/dev/null
  </packages>
 
  <!-- dist first -->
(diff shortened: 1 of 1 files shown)


deploy/packaging/linux.xml
```diff
@@ -836,6 +836,12 @@ rm -f /etc/ld.so.conf.d/mdsplus.conf 2&gt;/dev/null
  </packages>
 
  <!-- dist first -->
+  <external_packages dist="Ubuntu24">
```
WhoBrokeTheBuild, line 839: Looks good, after we drop support for Ubuntu18/DebianBuster we can change the `debian` default below and remove all these overrides

WhoBrokeTheBuild (approved):

Issue #2768, opened 2024-05-17
In `tdi/RfxDevices/DIO2` initialization, remove the flag that triggers a TDI `abort()` on a failed connection


mwinkel-dev:
**Affiliation**
MIT PSFC for LHCD project at GA / DIII-D

**Version(s) Affected**
Found in `alpha-7-141-78`.

**Platform**
Ubuntu 18.04, but surely exists on all platforms.

**Describe the bug**
There is an inconsistency in how the two `DIO2` device drivers handle `MdsConnect()` during device initialization.

The `tdi/RfxDevices/DIO2__init.fun` file calls `MdsConnect(<server>, 1)`.   The `1` is a flag to the `tdi/Remote/MdsConnect.fun` indicating that if there is a connection error that TDI should `abort()`.

However, the comparable `tdi/MitDevices/DIO2__init.fun` file does not use the `1` flag when it calls `MdsConnect()`.

We believe that the `tdi/RfxDevices/DIO2__init.fun` should not use the `1` flag.

**To Reproduce**
1. Compare the `MdsConnect()` statements in the `tdi/RfxDevices/DIO2__init.fun` file with those in the `tdi/MitDevices/DIO2__init.fun` file.

**Expected behavior**
n/a

**Screenshots**
n/a

**Additional context**
On 16-Feb-2023 an edit was made on a MIT PSFC server that removed the `1` flag from the `tdi/RfxDevices/DIO2__init.fun` file.

This discrepancy was noticed when investigating some initialization error messages generated by the `tdi/RfxDevices/DIO2` device.

mwinkel-dev:
Fixed by PR #2769, so closing this issue.

PR #2769, opened 2024-05-17
Fix: remove abort flag from RfxDevices DIO2 initialization

mwinkel-dev:
Fixes issue #2768.

Makes the `tdi/RfxDevices/DIO2` initialization behave the same way that the LHCD team was using the `DIO2` device from 16-Feb-2023 through 9-Mar-2024.


joshStillerman (approved):
I don't really know why this change was put in, but I guess we needed it.

WhoBrokeTheBuild (approved):

PR #2770, opened 2024-05-20
Fix: Missing repo metadata signing

WhoBrokeTheBuild:
This will hopefully fix the lack of signed metadata files that are preventing us from automatically publishing releases


mwinkel-dev (approved):
Concur with your assessment that merging this today is the best path for getting package and repo signing working again.   If this doesn't completely fix the issue, we can always make additional commits in the coming days.

PR #2773, opened 2024-05-23
Fix: sign the RPM package repo's metadata

mwinkel-dev:
CentOS and RHEL use RPM packages.   The metadata for the MDSplus RPM package repository is in the `repodata/repomd.xml` file.   To ensure that the contents of the file have not been altered, it should be protected with a "detached" digital signature.

This fix creates that digital signature.

**Note 1:**  Although this change works OK on a development system using a fake signing key, that environment is not identical to the build system's environment.   And thus this proposed change deserves extra scrutiny before being merged to alpha.

**Note 2:**  This change should also be cherry-picked to the stable branch.

**Note 3:**  This is related to PR #2770 (labeled as "Fix:") and thus this PR should also be labelled with "Fix:" (not "Build:").

### DIFF
--- a/deploy/platform/redhat/redhat_docker_build.sh
+++ b/deploy/platform/redhat/redhat_docker_build.sh
@@ -25,7 +25,7 @@ do_createrepo() {
   checkstatus abort "Failure: Problem creating rpm repository in ${repodir}!" $?
   rm -f ${tmpdir}/repodata/repomd.xml.asc
   export GNUPGHOME=/sign_keys/.gnupg
-  gpg --local-user RPM-GPG-KEY-MDSplus --detach-sign --armor ${tmpdir}/repodata/repomd.xml
+  gpg --local-user MDSplus --detach-sign --armor ${tmpdir}/repodata/repomd.xml
   unset GNUPGHOME
   : && rsync -a ${tmpdir}/repodata ${repodir}/${FLAVOR}/RPMS/
 }
--- a/deploy/platform/redhat/redhat_docker_build.sh
+++ b/deploy/platform/redhat/redhat_docker_build.sh
@@ -24,9 +24,7 @@ do_createrepo() {
   : && createrepo -q $update_args -o ${tmpdir} ${repodir}/${FLAVOR}/RPMS
   checkstatus abort "Failure: Problem creating rpm repository in ${repodir}!" $?
   rm -f ${tmpdir}/repodata/repomd.xml.asc
-  export GNUPGHOME=/sign_keys/.gnupg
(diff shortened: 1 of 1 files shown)


deploy/platform/redhat/redhat_docker_build.sh
```diff
@@ -23,6 +23,10 @@ do_createrepo() {
   fi
   : && createrepo -q $update_args -o ${tmpdir} ${repodir}/${FLAVOR}/RPMS
   checkstatus abort "Failure: Problem creating rpm repository in ${repodir}!" $?
+  rm -f ${tmpdir}/repodata/repomd.xml.asc
+  export GNUPGHOME=/sign_keys/.gnupg
```
mwinkel-dev: Note that when Docker runs, it maps the `--keys=<dir>` to `/sign_keys` in the workspace.   The hard-coded `/sign_keys` occurs in 3 other locations in this file.   If we replace `/sign_keys` with an environment variable, then the change should be made to all occurrences.

deploy/platform/redhat/redhat_docker_build.sh
```diff
@@ -23,6 +23,10 @@ do_createrepo() {
...
+  gpg --local-user MDSplus --detach-sign --armor ${tmpdir}/repodata/repomd.xml
```
mwinkel-dev: Note that there is one key that has two names: `MDSplus` and `RPM-GPG-KEY-MDSplus`.

- `MDSplus` is on the build system's keyring, and thus using it to sign the RPM metadata allows the build to pass.
- `RPM-GPG-KEY-MDSplus` is not on the keyring and thus signing fails (which causes the build to fail the "release" step).   Although the name is different, this is the same key as the "public" `MDSplus` key.  (Confirmed that by importing the `RPM-GPG-KEY-MDSplus` ASCII key-file into GPG, and then doing `--list-keys`.)
- The source repo has an older (obsolete) `RPM-GPG-KEY-MDSplus` in `deploy/platform/redhat/RPM-GPG-KEY-MDSplus`.

deploy/platform/redhat/redhat_docker_build.sh
WhoBrokeTheBuild: Instead of `export` and then `unset`, you can just set that env for just the `gpg` command, e.g.
`GNUPGHOME=/sign_keys/.gnupg gpg --local-user MDSplus...`

WhoBrokeTheBuild (changes requested):

deploy/platform/redhat/redhat_docker_build.sh
mwinkel-dev: A-OK, will make that change right now.

WhoBrokeTheBuild (approved):

PR #2774, opened 2024-05-27
Feature: jScope improvements

GabrieleManduchi:
jScope improvements:

1) Allow using expressions also for signal labels
2) Allow defining a time window for the most recent samples in trend visualization (i.e. absolute times). This is configured by setting in xmin field <expression returning window size in seconds>:


GabrieleManduchi:
Done, the Feature prefix should be also already present in the original commit comment

mwinkel-dev:
Hi @GabrieleManduchi -- Thanks for making the changes that @WhoBrokeTheBuild requested.

Note though that the first commit (hash `8400cac`) has a typo.  It is titled, **ture:jScope improvements**.   So when merging, double-check the default merge-commit message that GitHub creates.

Anyway, we'll review this PR later today.

WhoBrokeTheBuild (approved):

PR #2775, opened 2024-05-29
Docs: Update GitHub Bug Report Template

WhoBrokeTheBuild:
Request both a client and server version
Add instructions for getting the client and server versions
Request the installation methods


mwinkel-dev (approved):
These are useful changes that will help streamline support tasks.

PR #2778, opened 2024-05-30
Build: Fix version bumping

WhoBrokeTheBuild:
`get_new_version.py` was never resetting the minor or patch portions of the version
So 1.2.3 with a feature commit would give you 1.3.3, instead of 1.3.0


santorofer (approved):
Looks correct.

PR #2780, opened 2024-05-31
Build: Multithread IDL Tests

WhoBrokeTheBuild:
First attempt at multithreading the IDL tests

TODO: The write tests still need a way to access mdsip, and now in parallel
We'll probably use different shot numbers or something to that effect

This also moves the `testid` variable out of IDL and turns it into a parameter to each test
This lets us name the threads and output log files for each test individually

TODO: Output a JUnit XML to allow Jenkins to track these tests

### DIFF
--- a/idl/testing/run_tests.py
+++ b/idl/testing/run_tests.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import argparse
 import tempfile
+import threading
 
 # The default values are intended to be used from within the PSFC network
 # If you want to run these tests on your own infrastructure, provide the
@@ -231,7 +232,6 @@ parser.add_argument(
 
 args = parser.parse_args()
 
-
 #---------------------------------------------------------------------------
 # Each IDL write test should start with a clean tree.
 # Write tests use a local tree, but eventually will be upgraded to use mdsip.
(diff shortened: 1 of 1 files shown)


idl/testing/run_tests.py
```diff
@@ -259,98 +259,127 @@ def build_write_tree(tree, shot):
...
+    all_tests.append(t)
 
 
 #-------------------------------------------------------------------------------
 # Tree open / read / close
-idl_test(f'''
-         
-testid = 'IDL-tree-read'
+idl_test('IDL-tree-read',
+
```
mwinkel-dev: There is a blank line after the `idl_test(IDL-tree-read'`.   Is A-OK "as is".   However, if delete that blank line then this test case has the same style as all the other test cases.

mwinkel-dev (dismissed):
Have confirmed that this refactoring of the IDL test suite is suitable for multi-threading of the "read" tests.   So approving this PR.

There is a trivial formatting / style issue at line ~380 which is an extraneous blank line.   It is OK to leave it "as is".

We'll likely have to run this multi-threaded read test for a week or so before we can be confident that the connection stress tests (60 to 100 connections per test case) won't clash when run in parallel.

mwinkel-dev (dismissed):
Re-approving this PR.

heidthecamp:
This is currently in the cmake branch. The gains of adding this with our current jenkins set up is minimal while adding a lot of complexity. This will be merged with cmake.

PR #2784, opened 2024-06-01
fix: make TreeGetRecord call _TreeGetRecord(*TreeCtx(), nid_in, dsc)

zack-vii:
invert calling order of TreeGetRecord and _TreeGetRecord in favor of the thread safe version.
This method was left out originally as the underlying methods where not context ready, i.e tdishr. This should not be the case anymore.


WhoBrokeTheBuild (approved):

zack-vii:
retest this please

PR #2785, opened 2024-06-01
 Fix: XNCI writes corrupting node data (alternative)

zack-vii:
Proposal for a substitute of PR  #2498

### DIFF
--- a/treeshr/TreeGetRecord.c
+++ b/treeshr/TreeGetRecord.c
@@ -39,7 +39,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include "treeshrp.h"
 #include "treethreadstatic.h"
 
-#define align(bytes, size) ((((bytes) + (size)-1) / (size)) * (size))
+#define align(bytes, size) ((((bytes) + (size) - 1) / (size)) * (size))
 
 static int64_t ViewDate = -1;
 
@@ -64,7 +64,9 @@ static inline int read_descriptor(TREE_INFO *info, int length, NCI *nci, int nid
     status = TreeBADRECORD;
   else if (!(nci->flags2 & NciM_NON_VMS) &&
            ((retsize != length) || (nodenum != nidx)))
+  {
     status = TreeBADRECORD;
+  }
   else
     status = (MdsSerializeDscIn(data, dsc) & 1)
                  ? TreeSUCCESS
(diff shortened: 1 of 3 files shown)


WhoBrokeTheBuild (dismissed):
This works as expected and I'm approving, but @mwinkel-dev is still looking so please give him time to review before merging

Could you explain a bit of what was happening here?

zack-vii:
sure, well for what I gathered the attribute's facilities are meant to store (among others) either a standard record or a segment header. I am not exactly sure why the header and standard record are not stored in the same memory although they are exclusive. The nci flags/flags2 store the information about the context as well. Nonetheless, sunce they re exclusive, when writing a segment, the standard record should be removed. In the same way writing a standard record into a segmented node removes the segmented header ref in TreePutRecord. Maybe when refactoring the code to reduce code duplication the begin_attributes method inherited the removal. however, when writing an xnci the standard record should of course be preserved.
Changing the deleted length value to -1 is just to be consistant with the init value of attributes which is defined by memset of -1.

mwinkel-dev:
Hi @zack-vii -- Thanks for the explanation.

I have repeated the experiments that Stephen ran.   And obtained the same results.

I am now studying the changes (mostly to learn about XNCI).   If the review doesn't prompt any additional questions, I will add my approval to this PR later today.

zack-vii:
I happy to hear that. In case it is all good, feel free to merge the pr.

treeshr/TreeGetRecord.c
```diff
@@ -136,22 +138,16 @@ int TreeGetRecord(int nid_in, mdsdsc_xd_t *dsc)
                 status = TreeGetExtendedAttributes(
                     info, RfaToSeek(nci.DATA_INFO.DATA_LOCATION.rfa),
                     &attributes);
-                if (STATUS_OK &&
-                    attributes
-                            .facility_offset[STANDARD_RECORD_FACILITY] !=
-                        -1)
+                RETURN_IF_STATUS_NOT_OK;
```
mwinkel-dev: If `TreeGetExtendedAttributes()` encounters errors, this line probably returns a status of `TreeFAILURE` whereas the original code likely returned `TreeBADRECORD`.   However, the revised `TreeSegmentTest.c` works fine, thus presumably the different status code returned isn't important.

treeshr/TreeSegments.c
```diff
@@ -616,6 +611,8 @@ static int set_xnci(vars_t *vars, mdsdsc_t *value, int is_offset)
             break;
           }
         }
+        vars->attr_update = 1;
+        vars->attr.facility_offset[STANDARD_RECORD_FACILITY] = -1;
         status =
```
mwinkel-dev: Should this `status` variable be checked to see if is not OK?

mwinkel-dev (dismissed):
Have minor questions about two lines (see above posts).

Nonetheless, am approving this fix because it looks good and matches the explanation in a preceding post.  Plus, testing shows that it fixes the problem with XNCI breaking a record (and thus throwing a `TreeBADRECORD` error).

`TreeGetRecord.c` = just has stylistic changes to improve readability.
`TreeSegments.c` = has the real fix, plus stylistic changes to improve readability.
`TreeSegmentTest.c` = now also tests XNCI read / write and effect on the record.

treeshr/TreeGetRecord.c
zack-vii: replaced it with status = TreeBadRecord. this also keeps the program sequence uniform (single return statement)

zack-vii:
> I verified that this clobbers the data on the current alpha, and works as expected on your branch 🎉

for testing it, you dont need to write a segment first. the test extension just tests also if segmented nodes stay intact. as well as the other segment tests regarding multithreading. using more than one thread in the test fails on ntfs filesystems because ofd locks are not supported. i think there was a related issue. I played with a solution to an alternate locking mechanism in case ofd lock are not supported.

mwinkel-dev (approved):
Hi @zack-vii,

Thanks for the two recent commits and resolving my comments.   The two commits look fine, so I am approving this PR.

An even bigger thanks for your continuing contributions to MDSplus.   We are aware that you do so in your spare time.   Your knowledge of the code base and associated history enables you to troubleshoot / fix these complex issues quickly.   I appreciate your help and contributions.

PR #2787, opened 2024-06-04
Build: MATLAB struct tests

f-trx:
This PR adds the tests for the the `mdsFrom/ToMatlab` function, which converts native MATLAB objects into MDSplus objects.

In particular it tests the following recently introduced mappings:
- MATLAB `struct` ↔ `MDSplus.Dictionary`
- MATLAB `struct array` ↔ `MDSplus.List`s of `MDSplus.Dictionary`s
- MATLAB `cell array` ↔ `MDSplus.Apd`

by testing the methods `JavaFrom/ToMatlabStruct` and `JavaFrom/ToMatlabCell` in the `testJavaMatlabStruct.m` and `testJavaMatlabCell.m` respectively.

Tests for the previously existing numeric conversions are also added by testing the method `mdsFrom/ToMatlab` in the `testJavaMatlabNumeric.m` script.

Working principle: the `run_tests.m` script is run by the `run_tests.py` script, and the output of the former is checked by the latter.


mwinkel-dev:
Retest this please

mwinkel-dev (approved):
Hi @f-trx,

This PR is a nice way to extend the existing suite of MATLAB tests to run the ~115 conversion tests that you've added.   The change to `run_tests.py` looks good, as do the four MATLAB script files (`*.m`).   So, approving this PR.

After this PR is merged, the next step is for us to install a node-locked MATLAB license on the build system.   And the final step is to switch our existing tests over to the MATLAB testing framework as per your suggestion.

Issue #2777, opened 2024-05-30
Source repo contains obsolete public key for RPM packages


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
Found in alpha-7-142-80, but surely exists on all other platforms.

**Platform(s)**
Rocky Linux 9, but applies to other RHEL platforms.

**Installation Method(s)**
n/a

**Describe the bug**
The `deploy/platform/redhat/RPM-GPG-KEY-MDSplus` file contains a public key that was created on 26-Oct-2011 and is thus obsolete.   (The current signing key was generated on 6-Dec-2017.)

**To Reproduce**
1. Create a new GPG keyring
2. Import the RPM-GPG-KEY-MDSplus file from the source repository
3. Import the RPM-GPG-KEY-MDSplus file from `http://www.mdsplus.org/dist/RPM-GPG-KEY-MDSplus` (or the build system)
4. run `gpg --list-keys` and notice the different creation dates

**Expected behavior**
There are two options.

1. Remove the RPM-GPG-KEY-MDSplus public key from the source repository and instead distribute it via the MDSplus.org web site.
2. Or update the source repository so it contains the current public key.

**Screenshots**
n/a

**Additional context**
There might be value in saving all public keys used by the MDSplus project.   Although unlikely, there might be scenarios that would require a customer to revive / restore an ancient server running a very old version of MDSplus.   In that case, they might need the public key used to sign the packages in that old version.   Storing the public keys in the source repo is probably more robust than storing them just on `www.mdsplus.org` or the build system.

If use the source repo to archive the MDSplus public signing key for the RPM packages, should probably do the same for the DEB packages.   Perhaps move all public keys to `deploy/public_keys/`?

However, the private keys must not be archived in the source repo.   The private keys must be stored elsewhere (i.e., call it the "vault").   And it would be wise to store the public keys in the "vault" too.   So, perhaps it would be best to remove the public keys from the source repo.

mwinkel-dev:
This has turned out to be more complicated than expected.   Perhaps the better solution is to add some comments to explain why the current RPM packaging scripts use an obsolete public signing key as a placeholder.

mwinkel-dev:
Fixed by #2798.

PR #2798, opened 2024-08-02
Fix: Remove obsolete public RPM package signing key.

mwinkel-dev:
Fixes issue #2777.


mwinkel-dev:
Accidentally closed when rebased to get related PR #2801.   So, starting over.   This is WIP because there are some complexities that will require additional commits.

mwinkel-dev:
The central issue is that the `linux.xml` file that describes the packages and their contents, assumes static content.   However, this is not the case for the RHEL `repo` package -- its contents vary depending on whether the RPM packages are signed or unsigned.

The proposed fix thus adds some new XML tags that allow conditional elements in the XML description of RHEL's `repo` package.   (Plus also cleans up a few other minor issues with RPM signing.)

mwinkel-dev:
The initial PR #2773 was incomplete and caused a cascade of problems: #2777, #2800, and #2801.   This PR fixes all of those issues.

- works with both signed and unsigned builds
- the “repo” package now has different content for signed vs. unsigned
- the “gpg-pubkey” now has the correct fingerprint for the current key
- obsolete public key (used as a placeholder) has been deleted
- only includes a copy of the public key if doing a signed build
- checks the package repo’s metadata on signed builds

mwinkel-dev:
Making the entire `mdsplus*repo.rpm` file conditional simplifies the code.   Thanks to Stephen for suggesting this improvement.

deploy/packaging/redhat/redhat_build_rpms.py
```diff
@@ -92,15 +92,28 @@ def doRequire(info, out, root, require):
...
+    try:
+        os.stat('/sign_keys/.gnupg')
+        return True
+    except:
+        return False
+
+
 def build():
+
+    has_key = checkSigningKey()
```
WhoBrokeTheBuild: You might simplify this with `os.path.exists`

deploy/packaging/redhat/redhat_build_rpms.py
```diff
@@ -219,6 +232,7 @@ def build():
     try:
         os.stat('/sign_keys/.gnupg')
```
WhoBrokeTheBuild: Same comment as above

WhoBrokeTheBuild (changes requested):

deploy/packaging/redhat/redhat_build_rpms.py
mwinkel-dev: Yes, of course.   That is also a good catch.   I'll make the change.

deploy/packaging/redhat/redhat_build_rpms.py
mwinkel-dev: Same reply as above

deploy/packaging/redhat/redhat_build_rpms.py
mwinkel-dev: Change made.

deploy/packaging/redhat/redhat_build_rpms.py
mwinkel-dev: Use of `os.stat()` was the original code that existed prior to the recent RPM signing fixes.   Have simplified this as per suggestion above.

WhoBrokeTheBuild (approved):

mwinkel-dev:
Retest this please

PR #2808, opened 2024-09-12
Fix: String class breaks when using numpy >=2 with npbytes.

santorofer:
In the String class, check for Python version when using npbytes and npunicode 
Fixes #2807

PR #2809, opened 2024-09-16
Changes to support numpy 2+

### LINKED ISSUE #2807: another trouble with numpy 2.x
**Affiliation**
Consorzio RFX

**Version(s) Affected**
alpha-7.145-5

**Platform(s)**
win64

**Installation Method(s)**
windows installer from mdsplus.org repository.

**Describe the bug**
```
  File C:\MDSplus\python\MDSplus\tree.py:910 in getNode
    return TreeNode(int(nid.value), self)

  File C:\MDSplus\python\MDSplus\tree.py:1221 in __new__
    if str(node.usage) == "DEVICE":

  File C:\MDSplus\python\MDSplus\tree.py:1585 in usage
    return _scr.String(str(self.usage_str)[10:])

  File C:\MDSplus\python\MDSplus\mdsscalar.py:423 in __init__
    super(String, self).__init__(value)

  File C:\MDSplus\python\MDSplus\mdsscalar.py:114 in __init__
    self._value = self._ntype(value)

TypeError: string argument without an encoding
```

**To Reproduce**
just try to retrieve any data from a (recent?) pulsefile using a python installation with numpy version >= 2.0.x, e.g.
```
 data = tree.getNode( node ).data()
```

**Additional context**
the root cause (as for bug #2793) is that strings and bytes in numpy 2.x are no longer compatible for direct/implicit conversion, like in the older 1.x versions. 

I circumvent the problem with this trick in mdsscalar.py:
```
class String(Scalar):
    """String"""
    dtype_id = 14
#    _ntype = _ver.npbytes
    _ntype = _ver.npunicode
```
 
In this way I can retrieve data, but other operations, such as actions on devices are being broken.

WhoBrokeTheBuild:
Replace instances of `.array(copy=False)` with `.asarray()` as per numpy errors
Add error handling to scalar and array types to handle numpy 2+ throwing OverflowError exceptions when passed a negative or oversided number for the given type, e.g. `Uint8(-1)`
It will now explicitly calls the equivalent ctypes type to do the integer wrapping for us, which will be quite slow by comparison
Rewrite the `data_scalars` and `data_arrays` test cases to use more sensible numbers that can fit within a Int8, plus some cleanup

This fixes #2807 and replaces #2808

This will rely on https://github.com/MDSplus/Docker/pull/34 being merged to fully test

This will need to be tested on a wider range of python/numpy versions before I'm comfortable merging it, but I wanted to open the PR to let the automated testing work on it

### DIFF
--- a/python/MDSplus/tests/data_case.py
+++ b/python/MDSplus/tests/data_case.py
@@ -116,9 +116,7 @@ class Tests(_common.Tests):
         test('abs(-_a)', abs(-a), res.pop(), **kwargs)
         test('abs1(-_a)', m.ABS1(-a), res.pop(), **kwargs)
         test('abssq(-_a)', m.ABSSQ(-a), res.pop(), **kwargs)
-        rcount = 11
-        acount = 4+rcount
-        if almost:  # some operations only make sence on floating point
+        if almost:  # some operations only make sense on floating point
             test('exp(_a)',     m.EXP(a),    res.pop(), **kwargs)
             test('log(_a)',     m.LOG(a),    res.pop(), **kwargs)
             test('sin(_a)',     m.SIN(a),    res.pop(), **kwargs)
@@ -135,10 +133,6 @@ class Tests(_common.Tests):
                 test('cosd(_a)',    m.COSD(a),   res.pop(), **kwargs)
                 test('tand(_a)',    m.TAND(a),   res.pop(), **kwargs)
                 test('anint(_a/3)', m.ANINT(a/3), res.pop(), **kwargs)
-            else:  # strip results of real operations
-                res = res[:-rcount]
-        else:  # strip results of floating operations
-            res = res[:-acount]
         if real:  # some binary operations are only defined for real numbers
             test('_a mod _b',   a % b,   res.pop(), **kwargs)
             test('_a>_b',       a > b,   res.pop())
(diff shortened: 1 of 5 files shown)


python/MDSplus/tests/data_case.py
```diff
@@ -116,9 +116,7 @@ def _executeTests(test, a, b, res, **kwargs):
         test('abs(-_a)', abs(-a), res.pop(), **kwargs)
         test('abs1(-_a)', m.ABS1(-a), res.pop(), **kwargs)
         test('abssq(-_a)', m.ABSSQ(-a), res.pop(), **kwargs)
-        rcount = 11
-        acount = 4+rcount
-        if almost:  # some operations only make sence on floating point
+        if almost:  # some operations only make sense on floating point
```
mwinkel-dev, line 119: The test harness is using two flags: `almost` and `real`.   What are they supposed to do?   The confusion arises from the flag name versus the associated comment.   It appears that `almost` means floating point?

python/MDSplus/tests/data_case.py
```diff
@@ -135,10 +133,6 @@ def _executeTests(test, a, b, res, **kwargs):
                 test('cosd(_a)',    m.COSD(a),   res.pop(), **kwargs)
                 test('tand(_a)',    m.TAND(a),   res.pop(), **kwargs)
                 test('anint(_a/3)', m.ANINT(a/3), res.pop(), **kwargs)
-            else:  # strip results of real operations
-                res = res[:-rcount]
-        else:  # strip results of floating operations
-            res = res[:-acount]
         if real:  # some binary operations are only defined for real numbers
```
mwinkel-dev, line 136: This `real` flag and the associated comment are also confusing.   Note that "real" numbers is the entire number line from negative infinity to positive infinity (i.e., on a computer, that is all floating point numbers and integers).   Modulo operations are typically only done on the "integers".  (However, in Python one can indeed apply `mod` to floating point numbers.)

python/MDSplus/tests/data_case.py
WhoBrokeTheBuild: Yes, so `almost` eventually gets passed to `assertAlmostEqual(places=almost)` and is the number of decimal places to care about when comparing floating point numbers.

python/MDSplus/tests/data_case.py
WhoBrokeTheBuild: Yes, the `real` flag is always true except for the tests that use complex number types, for which the operations wouldn't be defined.

python/MDSplus/tests/data_case.py
mwinkel-dev: Thanks for the explanation.   That eliminates my confusion.

python/MDSplus/tests/data_case.py
mwinkel-dev: Also makes sense.   Thanks for the answer.

mwinkel-dev (approved):
The change to `asarray()` makes sense.   

The `try` block is appropriate given that numpy2+ doesn't handle overflow the same way as numpy1.x did.   (Unfortunate though that the conversion is likely slower.)

Explanation of the `almost` and `real` flags eliminates my confusion.

Changes to the test case are not a refactoring, but rather a complete rewrite of the test case to avoid overflow.

All looks good, so approving the PR.

PR #2814, opened 2024-10-09
Fix: putMany

### LINKED ISSUE #2813: PutManyExecute fails with thin client
**Affiliation**
PSFC @ MIT

**Version(s) Affected**
Alpha

**Platform(s)**
Any

**Installation Method(s)**
Repositories

**Describe the bug**
While writing a PutMany test for mdsthin, I found it failed to run PutManyExecute on the server

```
  File "mdsthin/test/write_test.py", line 230, in test_putmany
    pm.execute()
  File "mdsthin/connection.py", line 888, in execute
    result = self._connection.get('PutManyExecute($)', self._queries.serialize())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "mdsthin/connection.py", line 493, in get
    raise getException(manswer.status)
mdsthin.exceptions.LibKEYNOTFOU: %LIB-F-KEYNOTFOU, Key not found
```

**To Reproduce**
Follow the steps in:
https://github.com/MDSplus/mdsthin/blob/main/mdsthin/test/write_test.py#L223

**Additional context**
I believe this is caused by the lack of `EXPORT` on PutManyExecute here:
https://github.com/MDSplus/mdsplus/blob/0ffe7d8f0662b3e8da4007437eb250a7b9837355/mdsobjects/cpp/mdsdata.c#L928

I traced the LibKEYNOTFOU to this function:
https://github.com/MDSplus/mdsplus/blob/0ffe7d8f0662b3e8da4007437eb250a7b9837355/mdsshr/librtl.c#L581

And then, while checking the libmdsobjectscppshr.so I was unable to find a PutManyExecute symbol, which I think is the issue

WhoBrokeTheBuild:
There are two issues fixed, the first is that the `EXPORT` on `PutManyExecute` was missing, leading to the `LibKEYNOTFOU` error when calling the method from TDI

The second is in `putManyObj`, where we were passing `argsData->getDscs()` as `mdsdsc_t` pointers, when they were in fact `Data` pointers.

There is now a temporary vector that holds the results of `convertToDsc()` for each `Data`, and then calls `freeDsc` at the end of the function.

This fixes #2813 and unblocks the mdsthin putMany test

As part of the testing rehaul, we need to add GetMany and PutMany tests to catch this sort of thing.

### DIFF
--- a/mdsobjects/cpp/mdsdata.c
+++ b/mdsobjects/cpp/mdsdata.c
@@ -925,7 +925,7 @@ struct descriptor_xd EXPORT *GetManyExecute(char *serializedIn)
   return &xd;
 }
 
-struct descriptor_xd *PutManyExecute(char *serializedIn)
+struct descriptor_xd EXPORT *PutManyExecute(char *serializedIn)
 {
   static EMPTYXD(xd);
   struct descriptor *serResult;
--- a/mdsobjects/cpp/mdsipobjects.cpp
+++ b/mdsobjects/cpp/mdsipobjects.cpp
@@ -302,8 +302,8 @@ void *putManyObj(char *serializedIn)
       result->setItem(nodeNameData.get(), errorData.get());
     }
 
-    for (void * dsc : actualDscList) {
-      freeDsc(dsc);
+    for (size_t i = 0; i < actualDscList.size(); ++i) {
+      freeDsc(actualDscList[i]);
     }
   }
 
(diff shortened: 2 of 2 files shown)


GabrieleManduchi (approved):
Thanks for finding it....curious it was never noted before....

mwinkel-dev (approved):
Changes match the description.   All looks good.

Also, Stephen explained that it is a TDI *.fun that is calling PutManyExecute().   And also explained that I had missed the existing "catch" block (i.e., I forgot to expand the lines that the GitHub diff had hidden).

PR #2815, opened 2024-10-15
Build: Allocate ports to each test

WhoBrokeTheBuild:
This will allow the tests to run in parallel, so long as `$TEST_INDEX` and `$TEST_PORT_OFFSET` are set appropriately.

The list of ports allocated to each test is in `testing/ports.csv`
Each test has been updated to use the specific ports from that CSV

### DIFF
--- a/python/MDSplus/tests/_common.py
+++ b/python/MDSplus/tests/_common.py
@@ -200,13 +200,19 @@ class Tests(TestCase):
 
     @classmethod
     def main(cls):
+        if 'TEST_INDEX' in os.environ:
+            cls.index = int(os.environ['TEST_INDEX'])
+            
         cls.maxDiff = None  # show full diffs
         if cls.__module__ == '__main__':
             cls.__module__ = os.path.basename(sys.argv[0]).split('.', 1)[0]
             if len(sys.argv) == 2 and sys.argv[1] == 'all':
                 cls.runTests()
             elif len(sys.argv) > 1:
-                cls.runTests(sys.argv[1:])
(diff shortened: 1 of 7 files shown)


python/MDSplus/tests/_common.py
```diff
@@ -200,13 +200,19 @@ def runWithObjgraph(cls, tests=None):
 
     @classmethod
     def main(cls):
+        if 'TEST_INDEX' in os.environ:
+            cls.index = int(os.environ['TEST_INDEX'])
```
mwinkel-dev, line 204: Is `cls.index` just a hook for the future?   (I don't see where it is used.).  Regardless, I do like this change.

WhoBrokeTheBuild: Yeah, I think so. It was present in the testing framework, and seemed like it would be used for this exact sort of thing, so I decided to re-use it instead of make another one.

mwinkel-dev: A-OK.  I concur that it is good to have `cls.index` as it might be useful in the future.

mwinkel-dev (approved):
The port assignments in the code match the notes given in `testing/ports.csv`.   

Use of the port offsets allows the build system to avoid port collisions when running multiple instances of the test suite concurrently (i.e., a different port offset is assigned to each instance).

Looks good, so approving this PR.

joshStillerman (approved):
It would have been nice to be able to use the CSV in the tests directly, however I concur with your decision not to.

PR #2816, opened 2024-10-15
Feature: Deprecate TreeNOTOPEN in favor of TreeNOT_OPEN

WhoBrokeTheBuild:

This is not quite as open and shut as I would like, but there's only one file that uses `TreeNOTOPEN` and that's in `TreeSetNci.c`
That and the two names are very confusing

Setting `TreeNOTOPEN` as deprecated will inform how the exception classes are generated, and should clean some things up

~~We could consider removing the `TreeNOTOPEN` returns from `TreeSetNci.c`, however I'm worried this will break someone else's code that was looking for those values.~~

This now includes the changes to `TreeSetNci.c`

### DIFF
--- a/treeshr/TreeSetNci.c
+++ b/treeshr/TreeSetNci.c
@@ -180,7 +180,7 @@ int _TreeSetNci(void *dbid, int nid_in, NCI_ITM *nci_itm_ptr)
   int putnci = 0;
 
   if (!(IS_OPEN(dblist)))
-    return TreeNOTOPEN;
+    return TreeNOT_OPEN;
   if (dblist->open_readonly)
     return TreeREADONLY;
   if (dblist->remote)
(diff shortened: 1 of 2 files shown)


mwinkel-dev (dismissed):
I concur that it is confusing to have both `TreeNOTOPEN` and `TreeNOT_OPEN`.   The latter has 39 occurrences across 20 source files.   However, `TreeNOTOPEN` has just 3 occurrences in the `TreeSetNci.c` file.

Thus agree that `TreeNOTOPEN` should be flagged as "deprecated".   And that it should eventually be removed and replaced with `TreeNOT_OPEN`.

This change is A-OK "as is".   Might be a bit clearer though if it read `deprecated="replace with TreeNOT_OPEN"`

joshStillerman:
I approve, and I think we should change our one use of this in treeshr.  We should make sure this ends up in the Release notes (If we still make release notes)

joshStillerman (dismissed):
See comment I added

mwinkel-dev (approved):
Straightforward change, so approving.

PR #2818, opened 2024-10-24
Feature: Use python to generate files

WhoBrokeTheBuild:
This is being tagged as "Feature:" instead of "Build:" due to how many files it touches, I want to update the minor version number.

Add `opcodes.csv` to be parsed by `gen-*` scripts
Convert `opcbuiltins.h` and `tdishr.h` into generated files, generated by `opcodes.csv`
Update, Rename and Add `gen-*` scripts
Move yacc/lex files out of `yylex/` directories
Update `bootstrap` to call new `gen-*` scripts


deploy/gen-include-tdishr.py
```diff
@@ -0,0 +1,72 @@
...
+print("Generating '{}' from '{}'".format(OUTPUT_FILENAME, OPCODES_FILENAME))
+
+opcodes_file = open(OPCODES_FILENAME) # , newline=''
+reader = csv.DictReader(opcodes_file)
+
+opcode_names = []
+for line in reader:
+    opcode_names.append(line['name'])
+
+output_file = open(OUTPUT_FILENAME, 'wt')
```
zack-vii: you should use a context manager
```python
with open(...) as f:
    f.write(...)
    ...
```
instead of
```python
f = open(...)
f.write(...)
...
f.close()
```

deploy/gen-include-tdishr.py
```diff
@@ -0,0 +1,72 @@
+
+import os
+import csv
+import shutil
+
+from subprocess import *
```
zack-vii: is this import even used? besides you should avoid the usage of `from x import *` its not nice to IDEs and humans ;)

I suggest to autopep8 all gen-*.py files .

deploy/gen-messages-exceptions.py
```diff
@@ -1,4 +1,10 @@
 #!/usr/bin/env python
```
zack-vii: this shebang is missing from other files.. and should probably rather be python3.. well you are calling it explicitly with ${PYTHON} so it does not really matter for the auto generation. But for testing and developing it may be nice to be able to quickly call a single generator. In the end it depends on if you want these to be chmod-executable or not. they should be all alike though.

deploy/gen-include-tdishr.py
WhoBrokeTheBuild: Will do! I actually switched it to `import subprocess` in the other ones, I'll update this one to match.

deploy/gen-messages-exceptions.py
WhoBrokeTheBuild: I generally avoid putting the shebang line, I think this one is from the original file I modified this from. Ideally a human shouldn't be calling this script unless they're updating the script, it should be called by `bootstrap` or from CMake directly after the merge. I'll remove it from this so it's clear you shouldn't be `./` calling these files.

deploy/gen-include-opcbuiltins.py
```diff
@@ -0,0 +1,41 @@
...
+
+print("Generating '{}' from '{}'".format(OUTPUT_FILENAME, OPCODES_FILENAME))
+
+with open(OUTPUT_FILENAME, 'wt') as output_file:
+    with open(OPCODES_FILENAME) as opcodes_file:  # , newline=''
+
+        output_file.write(
+            '\n'
+            '/* <opcbuiltins.h> */\n'
+            '/*   name,                builtin,               f1,                  f2,       f3,               i1,          i2,         o1,         o2,         m1, m2,         token */\n'
```
mwinkel-dev, line 19: The table rows end with a "comment" field.   Should a "comment" label also be included in this file header?

WhoBrokeTheBuild: The columns are for things inside the `OPC()` macro, and the comment is just whatever was left as a comment on each line

mwinkel-dev (approved):
This refactoring looks good.   Some of the improvements . . . 
- opcodes.csv replaces the opcbuiltins.h and its OPC macros (which complicated debugging)
- clarity improved by renaming some files
- gen-messages-exceptions has a nice simplification for extracting the elements from the XML file, plus gets rid of the unused "msglistm" argument
- deleted the opcbuiltins.h file (that was added by the rebase)

Furthermore, prior versions of this PR successfully built and passed the automated tests.  And another core developer has already reviewed this.

For all the above reasons, this PR is low risk.   Thus approving.

Issue #2821, opened 2024-11-08
tags with 24 characters can be created and listed but not used


cenkoloji:
**Affiliation**
EPFL

**Version(s) Affected / Platform(s)**
Rocky Linux 9 - mdsplus-alpha 7.139-60
RHEL7 - mdsplus-stable 7.142.81
CentOS 8 - mdsplus-stable 7.131-5

**Installation Method(s)**
yum/dnf

**Describe the bug**
Tags with 24 characters can be created, or seen by `Tcl('dir /tag')`, but node can't be retrieved by tag

**To Reproduce**
python
```python
#!/usr/bin/env python

import MDSplus
t = MDSplus.Tree('cenktest',-1,"NEW")

nodes_tags = {"mynode{:02d}".format(n):"a"*n for n in range(1,26)}

for node,tag in dict(nodes_tags).items():
    try:
        n = t.addNode(node,usage='SIGNAL')
        n.addTag(tag)
    except:
        print("Can't create node,tag:{},{}".format(node,tag))
        nodes_tags.pop(node)

t.write()
t.close()

print()
t = MDSplus.Tree('cenktest',-1)

for node,tag in nodes_tags.items():
    try:
        t.getNode("\\" + tag)
    except Exception as ex:
        print("{} - {:30s} -> Can't get node from tag".format(node,tag));
        print(ex)
```
output:
```
Can't create node,tag:mynode25,aaaaaaaaaaaaaaaaaaaaaaaaa

mynode24 - aaaaaaaaaaaaaaaaaaaaaaaa       -> Can't get node from tag
%TREE-W-NNF, Node Not Found

```
Here I demonstrate that tag creation works as expected, up to 24 characters, but retrieval doesn't work for 24 characters.

merlea:
It seems the error is when calling `TreeFindNode` with the 24 character tag name. It always returns a "Node Not Found" status in this case.

Using `gdb`, it seems the failing step is `WildParse` where the search term is correctly identified as "TAG_TREE" but the search string has been shortened by one character (see at the end of the snipped below). I couldn't go further into debugging the Parser itself.

```
(gdb) bt
#0  FindTags (dblist=0x5555555c5380, info=0x5555555c5440, treenum=0, tagname=0x5555555aacf7 "ABCDEFGHIJKLMNOPQRSTUVW", tail=0x7ffffffee1f8) at ../../treeshr/TreeFindNode.c:445
[...]

PR #2824, opened 2024-11-09
Fix conversion of floating values using TEXT

### LINKED ISSUE #2823: Conversion of floating point values using TEXT is using the VMS syntax on Linux machines.
**Affiliation**
SPC-EPFL

**Version(s) Affected**
Server Version: Stable 7.142.81

**Platform(s)**
Ubuntu 22.04

**Installation Method(s)**
DEB package installed from MDSplus repository

**Describe the bug**
Text conversion of floating point values using `TEXT` is using the VMS syntax on Linux machines.
`TEXT(1.0)` or `TEXT(FS_FLOAT(1.0))` yield `"       1.00000S0"` while `"       1.00000E0"` was expected (and was the case prior to stable version 7.79.0).

**To Reproduce**
1. Open `tdic`
2. type `TEXT(1.0)`
3. Output if `"       1.00000S0"`

**Expected behavior**
The answer should be `"       1.00000E0"`. The main issue is that "1.00000S0" cannot be converted to a floating point number in other codes/languages.

**Screenshots**
```
$ tdic
TDI> TEXT(1.0)
"       1.00000S0"
```

merlea:
Fixes #2823.

The cause of the issue is that ever since 0d95b94c (which first appeared in the `stable_release-7-79-0`) `DTYPE_*` are not CPP macros anymore but enumerated types. Hence the CPP directive `#if DTYPE_F == DTYPE_NATIVE_FLOAT` was always true and the VMS settings were enforced.

Looking at `mdsdescrip.h` it seems that now the VMS build is not supported anymore and `DTYPE_NATIVE_FLOAT` is always defined to be `DTYPE_FS`, so I just removed the VMS option from `TdiConvert.c` as a fix.

### DIFF
--- a/tdishr/TdiConvert.c
+++ b/tdishr/TdiConvert.c
@@ -1187,32 +1187,6 @@ static void DOUBLEC_TO_TEXT(int itype, char *pa, char *pb, int numb, int lenb, c
 #define Q_T(lena, pa, lenb, pb, numb) TO_TEXT(pa, int64_t, pb, numb, sprintf(text, "%" PRId64, *ip++))
 #define OU_T(lena, pa, lenb, pb, numb) TO_TEXT(pa, uint128_t, pb, numb, sprintf(text, "%s", uint128_deco(ip++, int128_buf)))
 #define O_T(lena, pa, lenb, pb, numb) TO_TEXT(pa, int128_t, pb, numb, sprintf(text, "%s", int128_deco(ip++, int128_buf)))
-#if DTYPE_F == DTYPE_NATIVE_FLOAT
-#define F_T(lena, pa, lenb, pb, numb)              \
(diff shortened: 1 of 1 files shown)


mwinkel-dev:
Retest this please.

mwinkel-dev:
Preceding "retest" post is a request to Jenkins, not a request to merlea.

merlea:
Hi @mwinkel-dev, thanks I have submitted a few PRs to MDSplus now so am used to this.

By the way, let me know if you think some tests should also be added for this fix. I might need some guidance on how to implement them though. (Same for #2821 ).

joshStillerman (approved):

joshStillerman:
A test for this would be great.  I have approved it regardless.  Thanks.

Issue #2829, opened 2024-11-20
Actmon crashes the moment any action is dispatched.


cenkoloji:
**Affiliation**
EPFL/Swiss Plasma Center

**Version(s) Affected/Platform(s)**
RHEL9 - mdsplus-alpha 7.148.2
RHEL7 - mdsplus-stable 7.142.81

**Installation Method(s)**
yum/dnf

**Describe the bug**
Actmon crashes the moment any dispatch command is run.

**To Reproduce**
* On 4 separate terminals, start 3 servers and actmon (Change ports if these ones are not available)
```sh
mdsip -p 9999 -s -h ./mdsip.hosts  -c 9   # Dispatch server
mdsip -p 9998 -s -h ./mdsip.hosts  -c 9   # Action server
mdsip -p 9997 -s -h ./mdsip.hosts  -c 9   # Monitor server
actmon -monitor localhost:9997
```

Run following program which will create a tree with a single action node.
```python
import os
import time
import MDSplus

# Create tree path dir
user = os.environ['USER']
distest_path="/tmp/{}/distest".format(user)
os.system('mkdir -p {}'.format(distest_path))
os.environ['distest_path'] = distest_path

# Set environment vars for tree paths in the servers
for port in ["9997","9998","9999"]:
    c = MDSplus.Connection('localhost:' + port)
    c.get('setenv("distest_path=/tmp/{}/distest/")'.format(user))
    c.disconnect()

# Create a single action node
template = 'Build_Action(Build_Dispatch(2, "{}", "INIT", 10, *), {}, *, *, *)'
server = "localhost:9998"
function_expr = 'write(*,"Test action")'
expr = template.format(server, function_expr)

# Create main tree
t = MDSplus.Tree('distest',1,'NEW')
node = t.addNode('ACT01',usage='ACTION')
node.record = t.tdiCompile(expr)
t.write()
t.close()

# Run actions
MDSplus.tcl('dispatch /command /server=localhost:9999 set tree distest /shot=1')
time.sleep(1)
MDSplus.tcl('dispatch /command /server=localhost:9999 dispatch/build/monitor=localhost:9997')
time.sleep(1)
MDSplus.tcl('dispatch /command /server=localhost:9999 dispatch/phase/monitor=localhost:9997 init')

```

**Expected behavior**
Actmon doesn't crash

**Screenshots**

**Additional context**
Output of `gdb --args actmon -monitor localhost:9997 ` , then `run`, then `backtrace` after the crash

[...]

PR #2835, opened 2024-11-25
Fix: dispatching any action causes actmon/actlog to crash with SIGSEGV

merlea:
* Rename Svr -> Srv
* Always use 2 arguments for callback_done signature

A possible fix for #2829 basically setting all the `job_done`, `action_done` callbacks to have a signature ` void (*ast)(void *astprm, char *msg)` and setting `msg` to `NULL` when applicable. Using this fix with the case described in #2829 leads to no error.


zack-vii:
I think we need to add a test maybe using actlog (terminal version of actmon).
What i found is that actmon/actlog dont handle null either .. further they need to be protected in case the action monitor server (mdsip) becomes offline otherwise it may generate a stack overflow.

zack-vii:
If the ACTION_MONITOR server becomes unreachabel/offline The CheckIN will eventually call MessageAst with a NULL for message. This will cause if (!parseMsg(reply, event)) to skip and call CheckIN recursively until the program stack overflows.

We need to be able to detect this from the callback msg somehow, possibly by returning a special message instead. alternatively the CheckIN must be called from a async outside of MessageAst with some wait time to prevent a runaway process at 100% as well as the stack overflow due to recursion.

merlea:
OK. I removed the recursive calls to `CheckIn` since it already had a loop with pauses.

I also put some stubs for tests of `actlog` based on the original issue. We could compare the output of `actlog` with the expected one. I am not too familiar with the tests setup so let me know how best to integrate this.

joshStillerman:
retest this please

zack-vii:
sorry i totally missed this. for your tests comparing output with expected output i think the tditest tests are a good candidate to copy from.

merlea:
@WhoBrokeTheBuild I should have fixed the compilation issue.
I still have to work on the integration of tests, but perhaps we can rerun the Jenkins build to check that the compilation indeed goes through.

merlea:
OK. I have integrated the tests with the test suite.

I think this is now ready for a final review. @WhoBrokeTheBuild @zack-vii

merlea:
> OK. I have integrated the tests with the test suite.
> 
> I think this is now ready for a final review. @WhoBrokeTheBuild @zack-vii

@WhoBrokeTheBuild can we trigger the tests again please?

merlea:
Ping @WhoBrokeTheBuild , just making sure we don't forget this PR...

heidthecamp:
Retest this please

actions/testing/ServerShrTest.sh
```diff
@@ -0,0 +1,30 @@
...
+
+mdstcl @${srcdir}/test_action.tcl
+
+pkill -P $$
+
+# Cut time info
+cat actlog.log | cut -d' ' -f2- > actlog_cut.log
+
+# Check output of actlog against reference
+if diff actlog_cut.log ${srcdir}/actlog.log_ref; then
```
WhoBrokeTheBuild: Just like the other PR, can we rename .log_ref to .ans

actions/testing/ServerShrTest.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+srcdir=$(readlink -f $(dirname ${0}))
+
+export distest_path=$PWD
+
+mkdir -p $distest_path
+
+mdsip -p 9997 -s > server_9997.log 2> server_9997.err & # Monitor server
```
WhoBrokeTheBuild: Like you did in the other PR, can you allocate ports for this and use the $TEST_PORT_OFFSET? This is currently failing on all platforms due to conflicting ports.

WhoBrokeTheBuild (changes requested):
Once these are done, and the tests pass, I think we should merge this

heidthecamp:
Retest this please

actions/testing/ServerShrTest.sh
merlea: Done.

actions/testing/ServerShrTest.sh
merlea: Done.

merlea:
I implemented the requested changes but I expect some similar problems to #2906 in the tests for windows for example.

merlea:
> Oh and I forgot to mention / include it in the diff, but please move the ports close to the range of other allocated ports in `ports.csv`, your other PR uses up through 8020 so I would use 8021-8023, cheers

OK. Done. I think I covered everything, some parts I had worked on before you sent the diff, so they are slightly different.

heidthecamp:
Retest this please

WhoBrokeTheBuild (approved):

PR #2843, opened 2024-12-12
Fix imp deprecation

gtrevisan:
- fix the `imp` deprecation -- since it has been handled -- by suppressing the `DeprecationWarning`,
- specify the `ModuleNotFoundError` in the `except` clause,
- add a useful link next to the comment.


WhoBrokeTheBuild (approved):
Passed all tests and I manually tested it on python 2.7 to be sure, this looks good to me

Issue #2844, opened 2024-12-15
The transient mdsip server process always returns -1 which causes problems with systemd, leading to mdsip crash


mwinkel-dev:
**Affiliation**
General Atomics / DIII-D
(filed by @mwinkel-dev on behalf of @ModestMC for an issue reported via Discord)

**Version(s) Affected**
Client and Server versions are the same.
Found in alpha-7-139-59, but also occurs in several other versions.

**Platform(s)**
Found on RHEL9, but also occurs in Rocky9 and Ubuntu22

**Installation Method(s)**
On RHEL9, installation used the RPM packages.   Rocky9 and Ubuntu22 testing used MDSplus built from source.

**Describe the bug**
The mdsip server process always returns a -1 (aka shell exit status 255), which causes systemd to mark the service unit as "failed".  For transient mdsip services (`mdsip@.service` and `mdsip.socket`) this caused the mdsip server process to fail on RHEL9 when a simple stress test opened/closed ~60K mdsip connections.

Although the problem exists with persistent mdsip services too (`mdsip.service`), the process never stops which masks the -1 (255) exit issue.

In a client / server architecture, there are three basic causes for the server process to terminate:
1) the server crashes because of a bug in its code -- this is an error (but likely rare),
2) the client crashes and leaves a dangling socket that causes the server to timeout -- this indicates an error,
3) the client disconnects and the server times out -- this is normal operation and is OK.
The issue is the existing source code erroneously treats 3) as an error, so returns an exit code of -1.   The shell process that called the mdsip server converts that to 255.   So 255 gets returned to systemd which then marks the service unit as "failed" and apparently keeps the failed unit loaded into memory and/or recorded in various systemd data structures.

These are the error messages from GA's RHEL9 server when the mdsip server process crashed because systemd has too many "failed units".

```
$
$ # Edited error messages to remove sever names, etc.
$ # Items in <<brackets>> are placeholders.
$
$ python3 <<test_program.py>>
<MDSplus.connection.Connection object at 0x145a342ac910>
E, 1734043444.642:  get_bytes_to()                Connection(id=-1, state=0x00, protocol='tcp', info_name='tcp', version=0, user='(null)') error 0/48: Connection reset by peer
Error during login: recv NULL
Traceback (most recent call last):
  File "<<test_program.py>>", line 4, in <module>
    <<conn>> = MDSplus.Connection('<<mdsip_server>>')
[...]

PR #2846, opened 2024-12-23
Fix: systemd incorrectly reporting services failed

heidthecamp:


WhoBrokeTheBuild (approved):
LGTM, This should help your woes @ModestMC

mwinkel-dev:
This PR addresses Issue #2844.   Note that although the `mdsip@service` file is in the `rpm/` directory, it applies to all Linux distros.   

Also note that by declaring an exit code of 255 (aka -1) as "success", this PR masks true errors that might arise during client-server communication.   The root cause of the issue is that `mdsip` was not architected to distinguish between communication problems (an error condition) and clients disconnecting (an OK and normal condition).

At present, there is little value in fixing the architectural issue because there are no MDSplus downstream processes that would behave differently on success than on failure.   Stated another way, this PR is a simple and effective fix for Issue #2844.

sflanagan:
> Also note that by declaring an exit code of 255 (aka -1) as "success", this PR masks true errors that might arise during client-server communication. The root cause of the issue is that `mdsip` was not architected to distinguish between communication problems (an error condition) and clients disconnecting (an OK and normal condition).

Hi @mwinkel-dev 

Can you elaborate on what scenarios constitute said "communication problems"?

mwinkel-dev:
Hi @sflanagan -- They are just the usual scenarios that arise with client-server software.

- Network link between the client and the server could fail.
- Client program could crash (or the user could accidentally turn off their client computer).
- Unlikely, but a severely overloaded server dealing with resource starvation might encounter weird problems that kills a portion of its network stack.

In all of those scenarios, the mdsip process on the server times out.   Those timeouts are true error conditions.

Ideally, a client that disconnects normally, should be treated as a normal condition.   Unfortunately, the architecture of mdsip does not realize that a normal disconnect is OK; mdsip instead times out.   Which of course causes it to report an error.

Stated more succinctly, mdsip always returns -1 (aka 255) upon exit.   And has always done so.

However, it would be risky to rearchitect mdsip at this late date.   This architectural feature / flaw has been present in all versions of MDSplus.   Given the size of the installed base, plus the years that the existing code has been in service, rearchitecting mdsip now would likely introduce bugs and disrupt customer workflows.

If any of the communication problems listed above arise, it will be possible to troubleshoot them using logs from the client computer, logs from the network switches, and logs from the server.

ModestMC:
@mwinkel-dev does this mean that the client side disconnect command does not actually issue any messages to the server? I assume then that it is simply closing the socket on the client side.

One important question: if the mdsip process times out, what happens to any tree modifications in progress?

PR #2847, opened 2025-01-02
Fix: Close socket in MDSUdpEventCan

### LINKED ISSUE #2830: wfevent leaves UDP sockets after the it returns due to event or timeout
**Affiliation**
EPFL/Swiss Plasma Center

**Version(s) Affected/Platform(s)**
RHEL9 - mdsplus-alpha 7.148.2
RHEL7 - mdsplus-stable 7.142.81

**Installation Method(s)**
yum/dnf

**Describe the bug**
* After wfevent/wfeventRaw returns (due to an event or timeout), the UDP socket remains, causing number of sockets/open file descriptors to increase.
* This is not an issue for short lived programs, but for long acquisition programs that stay alive for long periods listening events, it can cause hitting limits of number of open files.
* For instance in one host I was hit with error
```
Cannot bind socket        
: Address already in use  
invalid eventid -1
```

**To Reproduce**
One one terminal
```sh
 echo "for (_i=0;_i<1000;_i++) {wfevent('sockettest',,1);}" | tditest
```
On other terminal show summary of open sockets
```sh
watch -n 1 ss -s
```
or (Assuming there is no other tditest process  running)
```sh
ls -l /proc/$(pgrep tditest)/fd
```

**Expected behavior**
After wfevent returns, UDP socket is destroyed

santorofer:
Fixes #2830 
When a UDP event listener is cancelled, the thread was killed but the socket was not closed.

### DIFF
--- a/mdsshr/UdpEvents.c
+++ b/mdsshr/UdpEvents.c
@@ -324,6 +324,7 @@ int MDSUdpEventCan(int eventid)
   closesocket(ev->socket);
 #else
   pthread_cancel(ev->thread);
+  close(ev->socket);
 #endif
   pthread_join(ev->thread, NULL);
   free(ev);
--- a/mdsshr/UdpEvents.c
+++ b/mdsshr/UdpEvents.c
@@ -324,9 +324,13 @@ int MDSUdpEventCan(int eventid)
   closesocket(ev->socket);
 #else
   pthread_cancel(ev->thread);
-  close(ev->socket);
 #endif
   pthread_join(ev->thread, NULL);
+
+#ifndef _WIN32
+  close(ev->socket);
+#endif
+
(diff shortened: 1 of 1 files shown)


mdsshr/UdpEvents.c
```diff
@@ -326,6 +326,11 @@ int MDSUdpEventCan(int eventid)
   pthread_cancel(ev->thread);
 #endif
   pthread_join(ev->thread, NULL);
+
+#ifndef _WIN32
+  close(ev->socket);
```
WhoBrokeTheBuild, line 331: This looks good, my only questions is if we should move the windows `closesocket()` call to this location as well.

WhoBrokeTheBuild (approved):

PR #2849, opened 2025-01-08
Build: Update Linux dependencies

### LINKED ISSUE #2819: Debian / Ubuntu package needs libMrm4 dependency.
**Affiliation**
OpenStar Technologies

**Version(s) Affected**
  Release:  alpha_release-7-145-7

**Platform(s)**
Ubuntu 22.04.5 LTS

**Installation Method(s)**
apt install mdsplus-alpha-motif

**Describe the bug**
traverser will crash with missing reference to libMrm4

**To Reproduce**
Start with system that doesn't have libMrm4 installed, and install or update mdsplus.

**Expected behavior**
It should run without crashing for missing library.  It is fixed by:
`apt install libMrm4`

**Screenshots**
```
darren@mediumboy:~$ traverser &
[1] 234351
darren@mediumboy:~$ traverser: error while loading shared libraries: libMrm.so.4: cannot open shared object file: No such file or directory
```

heidthecamp:
Change linux_build_packages.py to correctly fallback to platform if distribution is not set.
Update Linux xml to default to platform for common packages.
Add a dependencies for libxml2 and motif/libmrm4.

Fixes #2819

### DIFF
--- a/deploy/packaging/linux.xml
+++ b/deploy/packaging/linux.xml
@@ -843,63 +843,63 @@ rm -f /etc/ld.so.conf.d/mdsplus.conf 2&gt;/dev/null
    <numpy package="python3-numpy"/>
    <java package="java-runtime"/>
    <readline package="libreadline8"/>
-   <libmotif package="libmrm4">
+   <libmotif package="libmrm4"/>
  </external_packages>
 
  <external_packages dist="Ubuntu22">
    <numpy package="python3-numpy"/>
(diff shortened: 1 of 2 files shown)


joshStillerman:
retest this please

santorofer (approved):
I noticed that `<package="java-runtime">` have been removed from the linux.xml file for all the debian distributions, but just added to the debian platform, that makes sense.

PR #2851, opened 2025-01-09
Fix: the Fortran MacOS aliases now declare the remote calls

mwinkel-dev:
On MacOS, the `mdslib_fremotetest.f` program did not build because of linker issues.   Turns out that  `mdslib/fortran_aliases_mac.h` was missing declarations for all "remote" calls.   (Which implies that this test never ran on MacOS.)

This PR adds those declarations.


WhoBrokeTheBuild (approved):
Seems very likely this hasn't worked in a long time, LGTM

PR #2852, opened 2025-01-13
Build: Add include/opcbuiltins.h to the .gitignore

WhoBrokeTheBuild:
This is due to PR #2818 that generates opcbuiltins.h during bootstrapping


heidthecamp (approved):

mwinkel-dev (approved):
Good that Git will no longer complain about this generated file.

PR #2853, opened 2025-01-13
Feature: Python API to return variable length NCIs

### LINKED ISSUE #2788: Wrong number of child nodes in python TreeNode.getChildren()
**Affiliation**
RFX?

**Version(s) Affected**
The MDSplus Version(s) affected, if any.
e.g. Client Version: Alpha 

**Platform(s)**
Linux

**Installation Method(s)**
Compiled from sources

**Describe the bug**
If a Tree node has more than 4096 children, python TreeNode.getNumChildren() method returns the correct number of children, but the length of the array of nodes returned by method TreeNode.getChildren() is 4096

**To Reproduce**
Create a tree where a node has a number of child node larger than 4096
Access it with python

WhoBrokeTheBuild:
Allow the Python API to return variable length node lists for:
* MEMBER_NIDS
* CHILDREN_NIDS
* CONGLOMERATE_NIDS

This works by checking the NCI code and getting the corresponding 'number_of_{x}' property to determine the size of the buffer

This will, however, add an extra C function call whenever those NCIs are accessed, but this is more correct.

Fixes #2788

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -288,12 +288,12 @@ class Nci(object):
     MEMBER = (18, _C.c_int32, 4, None)
     CHILD = (19, _C.c_int32, 4, None)
     PARENT_RELATIONSHIP = (20, _C.c_uint32, 4, int)
-    CONGLOMERATE_NIDS = (21, _C.c_int32*1024, 1024*4, None)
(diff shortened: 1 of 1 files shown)


mwinkel-dev (approved):
The code changes match the text description.   Straightforward change so approving.

santorofer (approved):
Looks good, straight forward solution.

PR #2855, opened 2025-01-15
Build: Always use --dockernetwork for Jenkinsfile

WhoBrokeTheBuild:
Currently we only use a unique --dockernetwork for the Test stage
This will add it to all stages to ensure a unique network is always available


santorofer (approved):
This makes sense to me.

PR #2857, opened 2025-02-12
Feature: Add python copyTo method to Tree and TreeNode

WhoBrokeTheBuild:
This adds the copyTo method to Tree and TreeNode in the python API copyTo allows you to copy a portion of a tree to another tree, optionally:
* Filtering unwanted nodes
* Copying data
* Copying tags
* Copying XNCIs
* Copying NCIs

Example:
```py
t1 = Tree('testsrc', -1, 'READONLY')
t2 = Tree('testdst', -1, 'EDIT')

def skip_devices(node):
    return (node.conglomerate_elt == 0)

t1.BRANCH.copyTo(t2.SECTION, node_filter=skip_devices, copy_xnci=False)

t1.close()
t2.write()
t2.close()
```

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -3278,10 +3278,35 @@ class TreeNode(_dat.TreeRef, _dat.Data):
                         dst_node.makeSegment(start, end, dim, data)
                         index += seg_len
                 else:
-                    # Regular data is decompiled/recompiled to preserve node references
+                    def _update_tree_paths(data, new_tree):
+                        # TreePath is a subclass of TreeNode so we need to do this first
+                        if isinstance(data, TreePath):
+                            return data
+                        
+                        elif isinstance(data, TreeNode):
+                            try:
+                                return new_tree.getNode(data.minpath)
+                            except _exc.TreeNNF:
+                                return TreePath(data.path, new_tree)
+
+                        elif isinstance(data, _cmp.Compound):
+                            for i in range(data.getNumDescs()):
+                                data.setDescAt(i, _update_tree_paths(data.getDescAt(i), new_tree))
+                            return data
+
+                        elif isinstance(data, _apd.List):
+                            return _apd.List([ _update_tree_paths(v, new_tree) for v in data.value ])
+
+                        elif isinstance(data, _apd.Dictionary):
+                            return _apd.Dictionary({ _update_tree_paths(k, new_tree): _update_tree_paths(v, new_tree) for k, v in data.value })
+
+                        return data
+
                     try:
-                        deco = src_node.record.decompile()
-                        dst_node.record = dst.tree.tdiCompile(deco)
+                        # Decompiling/Compiling the data loses floating point precision
+                        # so we manually traverse the data to update node references
+                        dst_node.record = _update_tree_paths(src_node.record, dst_node.tree)
+
                     except _exc.TreeNODATA:
                         pass
                     except Exception as e:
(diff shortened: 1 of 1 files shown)


mwinkel-dev (commented):
Just making sure that I understand how devices are handled . . . 

1. I gather that if the `node_filter` removes the head node of a "device", that the entire device is skipped (as per the example given in the feature's description).
2. However, if devices are copied, the latter steps of the procedure apply a sanity check to make sure that all of the associated conglomerate_elt nodes have been copied and are in the proper order.

Is my understanding correct?

mwinkel-dev (dismissed):
The code is easy to understand and matches the feature description.   So, approving this PR.

python/MDSplus/tree.py
```diff
@@ -1182,6 +1182,10 @@ def tdiData(self, arg, **kwargs):
         kwargs['tree'] = self.tree
         return _dat.TdiData(arg, **kwargs)
 
+    def copyTo(src, dst, **kwargs):
```
heidthecamp: Why are there two definitions for copyTo?

python/MDSplus/tree.py
```diff
@@ -3102,6 +3106,201 @@ def updateSegmentLimits(self, start, end, idx):
...
+
+                # Skip this node and all children
+                if not copy_this_node:
+                    skip_nodes.extend(list(src_node.getNodeWild('***')))
+
+        # Handle existing nodes with discrepancies
+        for src_node in src_nodes:
+            dst_path = src_node.fullpath.replace(root_src, root_dst)
+
+            try:
```
heidthecamp, line 3189: Why are we not logging the exceptions?

python/MDSplus/tree.py
WhoBrokeTheBuild: Good question, the one in `Tree` is a helper function for people who want to copy an entire tree, e.g. `t1.copyTo(t2)` or `t1.copyTo(t2.SOMEWHERE)`

python/MDSplus/tree.py
WhoBrokeTheBuild: Not logging, but the whole function will fail

python/MDSplus/tree.py
```diff
@@ -1182,6 +1182,10 @@ def tdiData(self, arg, **kwargs):
         kwargs['tree'] = self.tree
         return _dat.TdiData(arg, **kwargs)
 
+    def copyTo(self, src, dst, **kwargs):
```
gtrevisan: shouldn't you remove `src` from here?
I'd imagine our first (and only) non-kw arg, that is, the destination, would get assigned to `src` rather than `dst`...
is the alias going to solve it?

python/MDSplus/tree.py
```diff
@@ -3102,6 +3106,204 @@ def updateSegmentLimits(self, start, end, idx):
                                         None,
                                         _C.c_int32(int(idx))))
 
+    def copyTo(self, dst, node_filter=None, copy_tags=True, copy_data=True, copy_nci=True, copy_xnci=True):
+        """Recursively copy a portion of one tree to another.
+        @param src: The source TreeNode to copy nodes from
```
santorofer: Maybe this line in the comments (i.e @param src: ...) should be removed and just add a comment below?

santorofer (dismissed):
Looks very good now.

mwinkel-dev (dismissed):
The `_update_tree_paths()` function looks fine.   As do all previous commits.

heidthecamp (approved):

PR #2860, opened 2025-02-19
Fix: the find_fun() function now checks the status variable

mwinkel-dev:
**Note: See next post for a better analysis.**

While troubleshooting an issue with a TDI *.fun file, noticed that the `findfile_fun()` function has a typo and some unnecessary code.

1. The `if (isext == EXT_PY)` is a typo because the associated block of code deals with *.fun files, not *.py files.
2. There is no need for the block to set the value of "isext" because it is already set prior to entering the `if()` test.
3. The search for the file uses two functions:`LibFindFileRecurseCaseBlind()` and `LibFindFileCaseBlind()`.   The debugger shows that the first function finds the full path to the *.fun file or *.py file (including the extent).  That full path is then passed in `bufd` to the second function which of course also finds the same file.   The second function is thus unnecessary.
4. After the second function is called, `pyfile` and `funfile` are left with their original paths, thus there is no need to uppercase the extents as is done after the first function.
5. Furthermore, there is no need to resize the `bufd` buffer, because it is only used for the unnecessary second function.

Also, the `find_fun()` function should test the status returned by its call of `findfile_fun()`.   That meant wrapping a block of code in an `if()` statement and indenting the block.   (Unfortunately, GitHub displays the indented code block in a visually cluttered way.)

The problems mentioned above apparently are from PR #2293 (which affected 625 files).

This PR simplifies the code and passes all the automated regression tests.

**_Note: Please scrutinize this PR closely.   In particular, look for any subtle side effects that would affect devices._**

### DIFF
--- a/tdishr/TdiVar.c
+++ b/tdishr/TdiVar.c
@@ -726,7 +726,7 @@ static inline int findfile_fun(const mdsdsc_t *const entry,
     }
     else if (isext == EXT_PY)
     {
-      *pyfile = file;  //XMW was funfile
+      *pyfile = file;
     } else {
       status = TdiUNKNOWN_VAR;
     }
--- a/tdishr/TdiVar.c
+++ b/tdishr/TdiVar.c
@@ -688,6 +688,9 @@ static inline ext_t matchext(const mdsdsc_d_t *const file)
     return EXT_PY;
   return EXT_NONE;
 }
+
+// If successful, will have either a *.fun or *.py file, and also compiles.
+// On error, must return TdiUNKOWN_VAR so that _TreeAddConglom() works correctly.
 static inline int findfile_fun(const mdsdsc_t *const entry,
                                char **const funfile, char **const pyfile)
 {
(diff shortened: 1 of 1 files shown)


mwinkel-dev:
It just occurred to me why the original code is probably correct.

1) Suppose that we have a TDI function, `stuff.fun`, that was written years ago.
2) But recently, it was rewritten as `stuff.py` (in preparation for eventually deleting `stuff.fun`).
3) Thus during the transition period the file system contains two functions: `stuff.fun` and `stuff.py`.

In that scenario, it probably would be useful for `findfile_fun()` to return both `stuff.fun` and `stuff.py`.   The original code will do that, but only if both files are in the same directory in the file system.

If that is indeed the intent of the `findfile_fun()` function, then most of this PR (but not all) should be scrapped.  (It would be good though to use the debugger to confirm that the original function does return both files.)  However, as a courtesy to future maintenance programmers it would be useful to add a one-line comment explaining that it returns both files.

The change to `find_fun` to check the status variable is still needed.   The rest of the PR though should be undone because it only returns one of the files (whichever one it finds first).

mwinkel-dev:
Experiments with the debugger confirm that the intent of `findfile_fun()` is to return both files.

- If `stuff.fun` and `stuff.py` are both in the same directory, then both will be returned.
- If both are returned, then the Python file is the one that TDI will execute.
- If the two files are in different directories, then only one will be returned (i.e., whichever one is found first).
- If only `stuff.fun` is found, the calling routine, `find_fun()`, also compiles it into TDI's byte code.

mwinkel-dev:
This PR has been revised / simplified as a result of the debugger experiments.

mwinkel-dev:
Retest this please.

mwinkel-dev:
Retest this please.

tdishr/TdiVar.c
```diff
@@ -820,30 +823,34 @@ static int find_fun(const mdsdsc_t *const ident_ptr, node_type **const node_ptr,
...
     INIT_AND_FREE_ON_EXIT(char *, funfile);
     // check if we can find method as either .py or .fun
+    // An error status must be remapped so that _TreeAddConglom() works OK.
     status = findfile_fun(ident_ptr, &funfile, &pyfile);
-    if (pyfile)
-    {
-      char *funname;
-      status = tdi_load_python_fun(pyfile, &funname);
-      if (STATUS_OK)
+    if (STATUS_NOT_OK) status = TdiUNKNOWN_VAR;
```
WhoBrokeTheBuild, line 828: We should wrap this in `{ }`, and just in general try to avoid one-line if statements and the like

tdishr/TdiVar.c
```diff
@@ -820,30 +823,34 @@ static int find_fun(const mdsdsc_t *const ident_ptr, node_type **const node_ptr,
...
+          if (STATUS_OK)
+          {
+            status =
+                put_ident((mdsdsc_r_t *)ident_ptr, &tmp, TDITHREADSTATIC_VAR);
+            MdsFree1Dx(&tmp, NULL);
+          }
         }
+        if (STATUS_NOT_OK)
+          // unable to load python method try tdi alternative
+          status = compile_fun(ident_ptr, funfile);
```
WhoBrokeTheBuild, line 849: Same comment as above

WhoBrokeTheBuild (approved):
A couple comments about code style, but overall the code looks good and does match the code around it, so it LGTM

tdishr/TdiVar.c
mwinkel-dev: I agree wholeheartedly.   I dislike the multitude of one line if statements in the code for two reasons:  easy to misread, and makes it a hassle to add temporary debug statements.

Only reason I left it without `{ }` was to conform to the existing code.

There are likely more single line ifs in the upcoming Apple Silicon PRs.   I probably won't catch all of them.   However, when doing maintenance work in the future will add the braces.

tdishr/TdiVar.c
mwinkel-dev: ditto as above

Issue #2859, opened 2025-02-18
The makeSegment() function fails causing a crash in a multi-threaded application


mwinkel-dev:
**Affiliation**
EPFL (submitted via email by Alex. S.)

**Version(s) Affected**
server = alpha-7-150-0
client = alpha-7-139-24

**Platform(s)**
client and server = Rocky Linux 9 (5.14.0.-362)

**Installation Method(s)**
assuming was installed from rpm packages

**Describe the bug**
The makeSegment() function is leaving descriptors open, thus causing a crash in the MARTe2 MDSWriterNode() function.  This problem apparently arose with PR #2288.

**To Reproduce**

1. Refer to the associated tar file which contains test code and a tree.

**Expected behavior**
MDSplus should not crash, but instead should behave the same way it did prior to PR #2288.

**Screenshots**
n/a

**Additional context**
The emailed bug report follows . . . 

We’ve made some extra investigation, here you’ve some context and findings:

1- Versions involved:
   Server: 
mdsplus: alpha_release-7-150-0
os: Rocky Linux 9 - 5.14.0-362.8.1.el9_3.x86_64 #1 SMP PREEMPT_DYNAMIC x86_64 x86_64 x86_64 GNU/Linux

   Client:
mdsplus: alpha_release_7.139.24
os: Rocky Linux 9 - 5.14.0-362.18.1.el9_3.0.1.x86_64+rt #1 SMP PREEMPT_RT x86_64 x86_64 x86_64 GNU/Linux

2- We have a client multi-thread application that performs node operations. Specifically, MDSWriter running in a main thread opens a tree, later it executes several calls to makeSegment() in another thread, so, two threads involved during the execution time.

3- As described previously, we started to have a crash in the execution of MDSWriter after a call to makeSegment() exactly at the 170th node (432 in total required for our three bolometers).

4- Investigating the previous limitation on the server side, we’ve found that after every call to makeSegment(), a set of file descriptors were left opened thus reaching the operating system limit (1024). 170 nodes, 3 file descriptors per channel, and for some reason it opens every file twice: 170 * 3 * 2 = 1024. To check this:

```
# display/count number of open file descriptors belonging to process
# This assumes last mdsip connection is the one started by the writer process
cd /proc/$(pgrep -f 'mdsip'|tail -n 1)/fd
ls -l
ls -l |wc -l

See the trace:

[...]

PR #2861, opened 2025-03-03
Fix: C++ TreeNode remove unnecessary Tree duplication

GabrieleManduchi:
When instantiating a new  TreeNode object, the passed Tree reference was not copied, and a new Tree instance was created instead. This caused a useless proliferation of tree connection, affecting system performance in applications dealing with a large number of nodes, such as the MARTe2-based Plasma Control System at TCV.   The Tree reference is now shared among TreeNodes created from it, thread consistency being ensured by the already present management internal to TreeNode by creating a new Tree connection for new threads.

### DIFF
--- a/mdsobjects/cpp/mdstreeobjects.cpp
+++ b/mdsobjects/cpp/mdstreeobjects.cpp
@@ -1058,6 +1058,7 @@ TreeNode::~TreeNode()
 {
 //The Tree instance is not dallocated since it may be shared with other TreeNode objects. 
 //It is responsibilitu of the user program deallocate the tree object when it is no more required.
+}
 
 std::string TreeNode::getNciString(int itm)
 {
--- a/mdsobjects/cpp/mdstreeobjects.cpp
+++ b/mdsobjects/cpp/mdstreeobjects.cpp
@@ -328,6 +328,7 @@ void *Tree::getCtx()
             return threadContextV[i].ctx;
         }
     }
+    printf("NEW CONTEXT ALLOCATED\n");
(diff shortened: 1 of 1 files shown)


WhoBrokeTheBuild (approved):

mdsobjects/cpp/mdstreeobjects.cpp
```diff
@@ -328,6 +328,7 @@ void *Tree::getCtx()
             return threadContextV[i].ctx;
         }
     }
+    printf("NEW CONTEXT ALLOCATED\n");
```
mwinkel-dev, line 331: Shouldn't this `printf()` be commented?   Or activated only if `DEBUG` is defined?

mwinkel-dev:
This PR addresses Issue #2859.

mdsobjects/cpp/mdstreeobjects.cpp
GabrieleManduchi: Right! This is a left debug message. I must have missed it when checking the differences

GabrieleManduchi: But I don't see it in the updated alpha....did you remove it?

WhoBrokeTheBuild: Yep, I merged a followup PR
https://github.com/MDSplus/mdsplus/pull/2862

GabrieleManduchi: Thanks!

PR #2863, opened 2025-03-03
Fix: Apple Silicon - use the correct function declarations

mwinkel-dev:
MDSplus has many variadic functions, including all of TDI's "intrinsic" functions.   Using the correct function declarations throughout the source code eliminates most of the segfaults that occur on Apple Silicon.   Furthermore, using the correct declarations has no impact on other platforms.


heidthecamp (dismissed):

WhoBrokeTheBuild (dismissed):

mwinkel-dev:
It is unclear why the `(structure descriptor *)` casts were initially not needed on the `cmake` branch.   However, they now are required on `cmake`.   Will investigate.

mwinkel-dev:
The use of `mdsdsc_t` is a good suggestion.   I will make that change.

The only reason I used `(struct descriptor *)` is because that is how `TdiDefFunction.c` declares the functions.   However, `TdiDefFunction.c` can of course also be switched to the synonym, `mdsdsc_t`.

I will make the changes in my local repos and if they pass the test suite (as is expected), will then add the `mdsdsc_t` change to this PR.

heidthecamp (approved):
It looks like the update since my last approval is to how we're actually calling the functions. Is that right? This looks good to me and seems to be passing all tests.

mwinkel-dev:
Hi @heidthecamp -- The TDI intrinsic functions were originally declared with the first parameter being `structure descriptor *`.   Stephen noted that MDSplus already has a type defined, `mdsdsc_t`, which is a synonym for the structure stuff.   (I don't know the early history of MDSplus, so don't know why `mdsdsc_t` was not used originally.)  Stylistically, it is better to use the type in the declaration and the casts, because that provides flexibility if we ever have to change the underlying data structure.

PR #2864, opened 2025-03-03
Fix: delete references to deprecated data types

mwinkel-dev:
MDSplus has several deprecated data types inherited from VAX VMS (including a COBOL data type).   These obsolete data types caused compilation errors when activated the debug print statements in the TdiShrExt.c file.

However, did not delete them because they are referenced in these other MDSplus source files: `DTYPE.java` and `dtypedef.h`.   Thus, just commented out the lines referencing the obsolete data types.

An alternative approach would have been to rename them (e.g., `CIT` becomes `CIT_deprectated` as per `dtypedef.h`).

### DIFF
--- a/tdic/TdiShrExt.c
+++ b/tdic/TdiShrExt.c
@@ -803,32 +803,33 @@ static void bTest(struct descriptor *desc)
       break;
     case DTYPE_HC: /*      30              H_floating complex */
       break;
-    case DTYPE_CIT: /*      31              COBOL Intermediate Temporary */
-      break;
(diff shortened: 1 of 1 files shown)


mwinkel-dev:
Correct, these are only in a debug function that produces verbose output.   As per your suggestion, I will delete the deprecated defines.

mwinkel-dev:
Retest this please.

mwinkel-dev:
Retest this please.

WhoBrokeTheBuild (approved):

PR #2865, opened 2025-03-03
Fix: Apple Silicon - add a define

mwinkel-dev:
The `MACOS_ARM64` define is used to isolate Apple Silicon specific code from all other platforms.   It is essential that this PR be merged before the rest of the Apple Silicon PRs.

In the `cmake` branch, this define was placed in the `mdsconfig.h.in` file.

When reviewing this PR, please decide if the name should be changed.   Note that Apple Silicon is based on the Arm Ltd. 64-bit architecture which Arm refers to as `aarch64`.   Some other possible names:
- MACOS_AARCH64
- APPLE_SILICON
- MACOS_M_CPUS

My preference is `MACOS_ARM64` to distinguish from the older MacOS (Intel) computers.

This PR is a partial fix for Issue #2597.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1225,6 +1225,13 @@ AH_VERBATIM([_WIN32], [
 #endif
 ])
 
+AH_VERBATIM([MACOS_ARM64],[
+// This define is used to bracket Apple Silicon specific code.
+#if defined(__APPLE__) && defined(__aarch64__)
+#define MACOS_ARM64
+#endif
+])
+
 dnl add here what goes on bottom of mdsconfig.h
 AH_BOTTOM(
 #endif // _MSC_VER else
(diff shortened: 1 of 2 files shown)


WhoBrokeTheBuild (dismissed):
I like `MACOS_ARM64` for the name, this LGTM

mwinkel-dev:
Temporarily converted to a "Draft" because investigating some differences between `alpha` and `cmake`.

mwinkel-dev:
The define of `MACOS_ARM64` must be accessible via the many includes of the `<mdsplus/mdsconfig.h>` file.   In the `cmake` branch, this is done by defining it in the `mdsconfig.h.in` file.   For the `alpha` branch, the define must be placed in the `configure.ac` (which then generates the `mdsconfig.h` file).

WhoBrokeTheBuild (approved):

Issue #1326, opened 2018-06-01
Documentation: TreeCreatePulseFile.fun


alkhwarizmi:
Hello all,

trying to create a pulse file remotely via the thin client mechanism (as a workaround to #1322) I found some difficulties calling to `TreeCreatePulseFile.fun`.

The Python user expects to call that function with one parameter only, the new shot number, because so he/she does in the documented mdsobjects API.

The fun instead has three parameters and their function is not documented. One seems to be a node ID but it is not clear to me what it is used for. The third parameter is totally obscure.

I had to look at the Python source code to understand that it is safe to call it with the last two parameters set to 0.

Sincerely,

tfredian:
The TreeCreatePulseFile.fun was implemented for internal use (mdsip thin client). It should have included documentation in the fun to indicate that intention. The additional arguments provide the added capability of the TCL> CREATE PULSE command which enables you to only include specific subtrees or exclude specific subtrees. With the trailing to arguments set to 0 it behaves like CREATE PULSE without those qualifiers. I suspect a more traditional way to accomplish what you were trying to do would be to use the tcl() function. I will look into adding "internal use only" comments to these treeshr fun's. When thinking about using the tcl() function for creating pulses I thought perhaps the tcl() fun could be enhanced to support arrays of strings for the commands: tcl(['set tree mytree','create pulse 1']) which might be useful for some applications. This feature might be useful for other languages such as python too.

PR #2866, opened 2025-03-04
Feature: Enable optional args in python wrapper createPulse

vadim-at-te:
This commit enables passing optional arguments to…_TreeCreatePulseFile via python Tree wrapper method createPulse. In particular, it allows one to copy the top tree but not the subtree files. It's a bit related to issue #1326

**Context**:
* Certain data, like calibration shots, feed into version-controlled workflows and necessitate flexible data provenance (version control of the data stored in MDSplus). This data is outside of the normal pulse sequence collected from "normal" Tokamak experiments. 
* We do not use MDSplus native version control features.
* For the better or worse, we find it simple, robust and convenient to use dedicated pulse ranges to enable data provenance and security. A monotonically-increasing pulse number, within the designated pulse range, becomes equivalent to data version number. The file-system operations, such as permissions management, digital signature (md5sum), etc, further supplement this approach to data provencance and security.
* essentially, version number of the MDSplus-stored data = pulse number in the special reserved pulse range.
* copying an MDSplus tree from such a designated pulse range, with pulse==version, into the model -1 pulse becomes equivalent to "git checkout certain-version". Also, local copies of codes/workflows can use the pulse matching the version of the code/workflow rather than using data in -1 pulse.
* changing version of the code / workflow is accompanied by changing the model tree to match the workflow / code. This is done by copying the correct version(==pulse number) of the tree from a reserve pulse range into -1.

**Motivation for this commit**:
Given our approach to data provenance, we want to be able to copy files of the top tree without copying its subtrees. This is possible in MDSplus using TDI function TreeCreatePulseFile but the user-friendly python wrapper createPulse doesn't allow passing on the optional arguments.

More generally, a top tree and its subtree(s) in general have distinct user permissions. There are situations where the user is allowed to copy and overwrite the top tree but not its subtrees, reflecting the domain of responsibilities and competence of that user. 

**potential shortcomings of the commit**
* wanted to check if the 4th optional argument, node_or_nid, is of type "MDSplus.tree.TreeNode" but wasn't able to find the correct way to do it. So, as a work-around, I compare input argument to self.getNode('\\TOP') since I know the latter is of node type
* I allowed fourth input argument to be node id itself or a node object. Arguably this makes it more convenient. I did not allow it to be node path. These choices are subjective
* the third, optional, argument, "copy_only_this", is simply an integer 0 or 1. Arguably, a more user-friendly approach is to use kwargs** so that the user can specify "only_copy_this": True. However, in my opinion most users will not use these options anyways, and only "power" users or admins will use it. Hence I left it as is.

Thank you in advance for considering this pull request and for any feedback.

### DIFF
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -777,7 +777,7 @@ class Tree(object):
                     _TreeShr._TreeSetSubtree(self.ctx, nid))
         return TreeNode(nid.value, self)
 
-    def createPulse(self, shot, copy_only_this=0, node_or_nid=0):
+    def createPulse(self, shot, copy_only_this=False, node_or_nid=0):
         """Create pulse.
     
         @param shot: Shot number to create
@@ -789,10 +789,10 @@ class Tree(object):
         @rtype: None
         """
 
-        if isinstance(node_or_nid, type(self.getNode('\\TOP'))):
+        if isinstance(node_or_nid, TreeNode):
             node_or_nid = node_or_nid.getNid()  # Extract node ID
     
-        nid_pointer=_C.cast(_C.pointer(_C.c_int32(node_or_nid)),_C.c_void_p)
+        nid_pointer = _C.cast(_C.pointer(_C.c_int32(int(node_or_nid))),_C.c_void_p)
 
         _exc.checkStatus(
             _TreeShr._TreeCreatePulseFile(
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -777,17 +777,31 @@ class Tree(object):
                     _TreeShr._TreeSetSubtree(self.ctx, nid))
         return TreeNode(nid.value, self)
 
-    def createPulse(self, shot):
+    def createPulse(self, shot, copy_only_this=0, node_or_nid=0):
         """Create pulse.
+    
         @param shot: Shot number to create
         @type shot: int
+        @param copy_only_this: Logical flag (0 or 1), defaults to 0
+        @type copy_only_this: int
+        @param node_or_nid: Either an integer (node ID) or a TreeNode object (defaults to 0)
+        @type node_or_nid: int or MDSplus.tree.TreeNode
         @rtype: None
         """
+
+        if isinstance(node_or_nid, type(self.getNode('\\TOP'))):
+            node_or_nid = node_or_nid.getNid()  # Extract node ID
+    
+        nid_pointer=_C.cast(_C.pointer(_C.c_int32(node_or_nid)),_C.c_void_p)
+
         _exc.checkStatus(
-            _TreeShr._TreeCreatePulseFile(self.ctx,
-                                          _C.c_int32(int(shot)),
-                                          _C.c_int32(0),
-                                          _C.c_void_p(0)))
+            _TreeShr._TreeCreatePulseFile(
+                self.ctx,
+                _C.c_int32(int(shot)),
+                _C.c_int32(int(copy_only_this)),
+                nid_pointer
+            )
+        )
 
     def deleteNode(self, wild):
         """Delete nodes (and all their descendants) from the tree. Note: If node is a member of a device,
--- a/python/MDSplus/tree.py
+++ b/python/MDSplus/tree.py
@@ -782,7 +782,7 @@ class Tree(object):
     
         @param shot: Shot number to create
         @type shot: int
-        @param copy_only_this: Logical flag (0 or 1), defaults to 0
+        @param copy_only_this: Logical flag, defaults to False
         @type copy_only_this: int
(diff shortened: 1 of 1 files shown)


vadim-at-te:
@WhoBrokeTheBuild thanks for a prompt reply.

Could you elaborate please? You want me to retest because it's not working for you ? Or is that a message left for your colleagues? 

It seems to work OK for me with MDSplus versions 7.49-3 and 7.131-6

Apologies, first time with a pull request here, so not familiar with the process

python/MDSplus/tree.py
```diff
@@ -777,17 +777,31 @@ def addNode(self, nodename, usage='ANY'):
...
         @param shot: Shot number to create
         @type shot: int
+        @param copy_only_this: Logical flag (0 or 1), defaults to 0
+        @type copy_only_this: int
+        @param node_or_nid: Either an integer (node ID) or a TreeNode object (defaults to 0)
+        @type node_or_nid: int or MDSplus.tree.TreeNode
         @rtype: None
         """
+
+        if isinstance(node_or_nid, type(self.getNode('\\TOP'))):
```
WhoBrokeTheBuild: You should be able to just do `isinstance(node_or_nid, TreeNode)`, unless that throws an error?

python/MDSplus/tree.py
```diff
@@ -777,17 +777,31 @@ def addNode(self, nodename, usage='ANY'):
                     _TreeShr._TreeSetSubtree(self.ctx, nid))
         return TreeNode(nid.value, self)
 
-    def createPulse(self, shot):
+    def createPulse(self, shot, copy_only_this=0, node_or_nid=0):
```
WhoBrokeTheBuild: What about making `copy_only_this` a boolean defaulted to `False`, and then keeping the `int()` conversion down below?
Also, we might consider making `node_or_nid` default to `None`, and then converting it to 0 inside the function, to make it clear that no node is being passed, but I could go either way on that one.

python/MDSplus/tree.py
```diff
@@ -777,17 +777,31 @@ def addNode(self, nodename, usage='ANY'):
...
+        @type copy_only_this: int
+        @param node_or_nid: Either an integer (node ID) or a TreeNode object (defaults to 0)
+        @type node_or_nid: int or MDSplus.tree.TreeNode
         @rtype: None
         """
+
+        if isinstance(node_or_nid, type(self.getNode('\\TOP'))):
+            node_or_nid = node_or_nid.getNid()  # Extract node ID
+    
+        nid_pointer=_C.cast(_C.pointer(_C.c_int32(node_or_nid)),_C.c_void_p)
```
WhoBrokeTheBuild: I would add spaces around the `=` and possibly add an `int()` cast inside the `_C.c_int32(node_or_nid)` to match the calls down below.

WhoBrokeTheBuild (changes requested):

python/MDSplus/tree.py
WhoBrokeTheBuild: Alternatively we could check `isinstance(node_or_nid, int)`, I suppose

python/MDSplus/tree.py
vadim-at-te: I've implemented your suggestions, thanks. Except I left node_or_nid as 0 by default, which is iequivalent to t.getNode('\\TOP'). 

My reason for keeping node_or_nid as 0 by default is that mdsplus/tdi/treeshr/TreeCreatePulseFile.fun sets it 0 if no optional argument is supplied to the wrapper and, also, 0 is not "nothing", it is the top of the open tree. So 0 is kind of a "true" default value while "None" is misleading, in my opinion.

python/MDSplus/tree.py
vadim-at-te: implemented - thanks.

python/MDSplus/tree.py
vadim-at-te: TreeNode works - thanks. I should've guessed it

WhoBrokeTheBuild (approved):

vadim-at-te:
hi @heidthecamp , @joshStillerman and @mwinkel-dev 

Sorry to bug you, is this still on your radar for carrying out a code review? 

I am just going through my first MDSplus source code pull request and pretty keen to tick this off and build an experience for the future pull requests. If the typical timescale of these review is, say a week, or a month or 3 months or something, I am happy to put a reminder and check in on this later. I do appreciate the fact that you are busy and this is a low priority. I just don't want this to drop off everyone's to-do list.

Thanks in advance!

joshStillerman:
Sorry for the delay.  We are having trouble getting it to pass the regression tests.  Unrelated to its content.

joshStillerman:
retest this please

PR #2867, opened 2025-03-05
Fix: Apple Silicon - defines for use with LibCallgFfi()

mwinkel-dev:
The Apple Silicon port uses the `libffi` library for calling variadic functions from libraries.    These defines are only used for Apple Silicon and have no impact on Linux and Windows.   This PR must be merged before the remaining Apple Silicon PRs are submitted.

### DIFF
--- a/include/libroutines.h
+++ b/include/libroutines.h
@@ -16,6 +16,19 @@
 /// defined in MdsShr.so
 
 extern void *LibCallg();
+#ifdef MACOS_ARM64
+extern void *LibCallgFfi();
+#define NOT_VARIADIC 0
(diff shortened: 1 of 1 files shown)


mwinkel-dev:
Perhaps the various `VARIADIC_*` defines should have been implemented as an enum.

heidthecamp (approved):

joshStillerman (approved):

WhoBrokeTheBuild (commented):
I would not have approved this PR as is, these defines should have been merged with the actual functions they add and the places where the `#defines` are used.

PR #2868, opened 2025-03-05
Fix: add missing va_end() statements

mwinkel-dev:
Variadic functions use the `va_start()`, `va_arg()` and `va_end()` macros to process the list of arguments.   Although the `gcc` compiler does nothing for `va_end()`, nonetheless the standard is to always call `va_end()`.   Doing so, makes it possible to port the code to other architectures / compilers that actually do emit code for `va_end()`.

Presently, about 50% of the MDSplus source code uses `va_end()`.

This PR supplies the missing `va_end()` calls so that all the source code conforms to the standard.

For more details refer to this `gcc` documentation:
[gcc arg macros](https://www.gnu.org/software/libc/manual/html_node/Argument-Macros.html)

### DIFF
--- a/mdslib/MdsLib.c
+++ b/mdslib/MdsLib.c
@@ -135,6 +135,7 @@ extern EXPORT int descr(int *dtype, void *data, int *dim1, ...)
 
     va_start(incrmtr, dim1);
     dim = va_arg(incrmtr, int *);
+    va_end(incrmtr);
     if (*dim == 0)
     {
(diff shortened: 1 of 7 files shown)


joshStillerman (approved):

mwinkel-dev:
Retest this please.

mwinkel-dev:
Retest this please.

mwinkel-dev:
Retest this please.

PR #2869, opened 2025-03-05
Fix: Apple Silicon -- traverser now calls TDI functions directly

mwinkel-dev:
The `traverser` tool that is part of MDSplus is a Motif application that is used to examine trees.   (It is similar to `jTraverser` and `jTraverser2`.)   This PR eliminates a segfault that occurs when the `traverser` is run on Apple Silicon.   This change has no impact on Linux or Windows.

The issue is that `traverser` creates a function pointer to some TDI intrinsic functions that are variadic.   On Apple Silicon, the compiler gets confused so emits the wrong calling code, causing a segfault.   The remedy is to skip the function pointer and just directly call the TDI functions.

### DIFF
--- a/xmdsshr/XmdsExpr.c
+++ b/xmdsshr/XmdsExpr.c
@@ -372,7 +372,7 @@ EXPORT struct descriptor *XmdsExprGetXd(Widget w)
       TreeGetDefaultNid(&old_def);
       TreeSetDefaultNid(def_nid);
     }
-    status = TdiCompile(&text_dsc, ans MDS_END_ARG);
+    status = TdiCompile((mdsdsc_t *)&text_dsc, ans MDS_END_ARG);
     if ((STATUS_OK) == 0)
     {
       TdiComplain(w);
@@ -755,7 +755,7 @@ static void LoadExpr(XmdsExprWidget w, struct descriptor *dsc)
         TreeGetDefaultNid(&old_def);
         TreeSetDefaultNid(def_nid);
       }
-      status = TdiDecompile(xd, &text MDS_END_ARG);
+      status = TdiDecompile((mdsdsc_t *)xd, &text MDS_END_ARG);
       w->expr.is_text = 0;
       if (STATUS_OK)
       {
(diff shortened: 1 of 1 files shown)


heidthecamp (dismissed):
Is `(*ew->expr.compile)` just pointing to essentially the same function? That's an _interesting_ way of doing that.

santorofer (dismissed):
I like the solution, simple and straightforward. I hope there will be no hidden side effects of doing that?. It doesn't seem to be the case, because it looks very self-contained to me.

mwinkel-dev:
Hi @heidthecamp -- Yes, the `expr` stuff is pointing to the TDI function.   For example, this line in the source code binds `TdiCompile()` to `expr.compile`.   
[XmdsExpr.c, line 232](https://github.com/MDSplus/mdsplus/blob/ad9b55ffecc25b20ad16b9610251f79e32440429/xmdsshr/XmdsExpr.c#L232)

Hi @santorofer -- This fix worked fine when I tested it on Apple Silicon.   However, I noticed that this change just failed Jenkins (sigh), so perhaps there are some side effects.   I will investigate this issue later today.

santorofer:
I see, it seems that there is a conflict of data types: TdiCompile() expects 'mdsdsc_t' while the one passed is 'struct descriptor_s'.

mwinkel-dev:
Hi @santorofer -- Thanks for investigating for me.   That is an easy fix with a cast.   I will make the edit and update this PR.

This glitch is the result of the massive change to PR #2863 to switch to `mdsdsc_t` instead of `structure descriptor`.

santorofer:
Retest this please.

santorofer (approved):

xmdsshr/XmdsExpr.c
```diff
@@ -372,7 +372,7 @@ EXPORT struct descriptor *XmdsExprGetXd(Widget w)
       TreeGetDefaultNid(&old_def);
       TreeSetDefaultNid(def_nid);
     }
-    status = (*ew->expr.compile)(&text_dsc, ans MDS_END_ARG);
+    status = TdiCompile((mdsdsc_t *)&text_dsc, ans MDS_END_ARG);
```
WhoBrokeTheBuild, line 375: I would rather solve this with a cast, rather than avoiding using the function pointers at all, this could have larger implications we can't see.

PR #2870, opened 2025-03-05
Fix: IDL tests now have correct paths

mwinkel-dev:
The Python program that tests the MDSplus API for IDL references several nodes in trees.   Some of the node paths were missing `\\`, which was causing error messages to be displayed when the test was run.   This PR fixes those typos.

When the Jenkins build system is configured to automatically run the IDL tests, it might be necessary to revisit this PR.   (A vague recollection is that the output from running the test varied a bit depending on which MDSplus archive server was used.)

### DIFF
--- a/idl/mdsconnect.pro
+++ b/idl/mdsconnect.pro
@@ -6,7 +6,7 @@ endcase
 end
 
 function sockmin
-  if !version.os eq 'linux' then return, 0 else return, 11-(!version.os eq 'MacOS')
+  if ((!version.os eq 'linux') or (!version.os eq 'darwin')) then return, 0 else return, 11-(!version.os eq 'MacOS')
 end
 
 function mds$socket,quiet=quiet,status=status,socket=socket
--- a/idl/mdsconnect.pro
+++ b/idl/mdsconnect.pro
@@ -6,7 +6,7 @@ endcase
 end
 
 function sockmin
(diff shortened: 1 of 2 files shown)


mwinkel-dev:
Retest this please.

mwinkel-dev:
Conjecture is that this PR is A-OK, and that the Jenkins system has a flaky `test-asan`.   Sometimes this PR builds fine, other times it fails just in `test-asan`.

mwinkel-dev:
Retest this please.

idl/mdsconnect.pro
```diff
@@ -6,7 +6,7 @@ endcase
 end
 
 function sockmin
-  if !version.os eq 'linux' then return, 0 else return, 11-(!version.os eq 'MacOS')
```
WhoBrokeTheBuild: This should either have it's own PR, or should be called out more explicitly as part of this one

WhoBrokeTheBuild (changes requested):

idl/mdsconnect.pro
mwinkel-dev: I will split this into two PRs.   One for just the path changes; the other to add `darwin` to the `!version.os` stuff.

The `MacOS` test appears to be obsolete code.   My recommendation is that we deal with it in a multi-step approach:
- leave `MacOS` test for now
- add `darwin` test (which is what IDL now returns for current MacOS systems)
- after the Apple Silicon branch has been released and used by customers, then do a new PR to remove the obsolete `MacOS` test

WhoBrokeTheBuild (approved):

PR #2877, opened 2025-03-09
Fix: Apple Silicon - add libffi for calling variadic functions

mwinkel-dev:
Apple Silicon requires the `libffi` library for calling variadic functions.   There are three execution flows that use `libffi`.   The simplest is when the source code directly calls TDI intrinsic functions.   (All TDI intrinsic functions are variadic.)

This PR is the complete vertical slice for the first execution flow.   A new function, `LibCallgFfi()`, is used on Apple Silicon to call the TDI intrinsic functions.

The use of the `#ifdef MACOS_ARM64` blocks isolate the Apple Silicon code.   All other platforms continue to execute the same code that they always have.   Thus this PR should have no impact on Linux, Windows or Intel MacOS platforms.

Some notes . . . 

- The `MdsLib.c` file now includes `<libroutines.h>` which is why the declaration of `LibCallg()` was deleted.
- The `TdiExtPython.c` file calls two variadic Python functions.
- The `TreeDoMethod.c` has a comment about MDSplus "devices" calling variadic "methods".

**Addendum:**  As of 12-Mar-2025, this PR contains all the remaining Apple Silicon specific code.   It has two vertical slices, changes to `lex`, changes to tests and so forth.   (The third vertical slice is in WIP #2882 and likely won't be implemented.)  However, this PR does not contain any of the changes needed for the build system and `cmake` branch.


include/libroutines.h
```diff
@@ -16,6 +16,21 @@
 /// defined in MdsShr.so
 
 extern void *LibCallg();
+#ifdef MACOS_ARM64
+extern void *LibCallgFfi();
+enum LibCallgFfiArgs {
```
WhoBrokeTheBuild: Is there a reason for the enum rather than just `int fixedArgs` as the parameter?

javamds/mdsobjects.c
```diff
@@ -1441,7 +1441,11 @@ JNIEXPORT jobject JNICALL Java_MDSplus_Data_compile(JNIEnv *env,
   arglist[varIdx++] = &outXd;
   arglist[varIdx++] = MdsEND_ARG;
   *(int *)&arglist[0] = varIdx - 1;
+#ifdef MACOS_ARM64
```
WhoBrokeTheBuild: We could consider wrapping these in a macro, or calling a macro that correctly calls `LibCallg` or `LibCallgFfi` based on the platform, instead of having the #ifndef block each time

include/libroutines.h
```diff
@@ -16,6 +16,21 @@
...
+    NOT_VARIADIC = 0,
+    VA_1_FIXED_ARG = 1,
+    VA_2_FIXED_ARGS = 2,
+    VA_3_FIXED_ARGS = 3
+};
+enum LibCallgFfiRtype { 
+    RTN_NONE = 0,
```
WhoBrokeTheBuild: We should probably prefix these with something like `MDS_FFI_` to make sure we don't conflict with other symbols. The same goes for the above enum if we keep it.

WhoBrokeTheBuild: We could also consider using dtype definitions for this, but that's probably not a good idea.

include/libroutines.h
```diff
@@ -16,6 +16,21 @@
 /// defined in MdsShr.so
 
 extern void *LibCallg();
+#ifdef MACOS_ARM64
```
WhoBrokeTheBuild: We probably want to have a broader define of `MDSPLUS_USE_FFI` or something similar, and it can be enabled automatically for `MACOS_ARM64` systems. This would let us easily test it on other systems that support libffi.

mdsshr/librtl.c
```diff
@@ -280,6 +284,72 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
   return 0;
 }
 
+
+#ifdef MACOS_ARM64
+EXPORT void *LibCallgFfi(void **const a, void *(*const routine)(), int num_fixed_args, int rtype)
+{
+  if (!routine)
```
WhoBrokeTheBuild: Should this instead be `assert(routine)`?

mdsshr/librtl.c
```diff
@@ -280,6 +284,72 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
...
+    abort(); // intercept definite stack corruption
+
+  int num_args = *(int *)a & 0xff;
```
WhoBrokeTheBuild, line 306: Can you add a comment explaining what's happening here? Is the number of arguments always passed as the first argument?

mdsshr/librtl.c
```diff
@@ -280,6 +284,72 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
...
+  
+  if (num_fixed_args > 0) {
+    enum { SIZE = 32 };
```
WhoBrokeTheBuild: This should probably be moved and named `MDSPLUS_FFI_MAX_FIXED_ARGS` or something

mdsshr/librtl.c
```diff
@@ -280,6 +284,72 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
...
+    ffi_cif cif;
+    ffi_type *arg_types[SIZE];
+    void *values[SIZE];
+    void *myresult;
+
+    if ((num_args > SIZE) || (num_fixed_args > SIZE)) {
+      printf("Error - currently no more than 32 arguments supported on external calls\n");
```
WhoBrokeTheBuild: And the constant should be used in the printf, in case it someday changes

include/libroutines.h
WhoBrokeTheBuild: Wouldn't it be `VOID` instead of `NONE?

mdsshr/librtl.c
```diff
@@ -280,6 +284,72 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
...
+      ffi_call(&cif, (void (*)(void))routine, &myresult, values);
+      if (rtype != RTN_NONE) {
+        return myresult;
+      }
+    }
+
+  // Num_fixed_args = 0 denotes a non-variadic function.   Usually, the check 
+  // for non-variadic is done prior to calling this routine.  
+  } else {
+    return LibCallg(a, routine);
```
WhoBrokeTheBuild: We could consider reorganizing the function to be
```
if num_fixed_args == 0 {
    return LibCallg(...)
} 

everything else ...
```

mdsshr/librtl.c
```diff
@@ -280,6 +284,72 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
...
+  }
+  return 0; // should never reach this
```
WhoBrokeTheBuild: We could add an `assert` to make sure it is never hit

mitdevices/t4012.c
```diff
@@ -418,6 +418,7 @@ static int ReadChannel(InStoreStruct * setup, int chunk, int samples, unsigned s
   return status;
 }
 
+// Usually called without a function, but once is called with TdiData().
```
WhoBrokeTheBuild, line 422: We could possibly even add `assert(routine == &TdiData)` if that's the assertion we're making

WhoBrokeTheBuild (changes requested):

include/libroutines.h
mwinkel-dev: A-OK, I will add this.   (I had something similar at one time, but stripped it out when partitioning into multiple PRs.)

include/libroutines.h
mwinkel-dev: Just a matter of personal preference.   I like the first statement below over the second statement.

```
status = (int)(intptr_t)LibCallgFfi(arglist, TdiCompile, VA_1_FIXED_ARG, RTN_INT32);
status = (int)(intptr_t)LibCallgFfi(arglist, TdiCompile, 1, RTN_INT32);
```

However, it is a weak preference and I can easily replace all VA_<N>_FIXED_ARGS  with integers if you prefer.

If we retain the VA_<N>_FIXED_ARGS, the question is whether they should be defines or enums.   I prefer enums for a set of related items.

Regardless, if there is a programming style guide for MDSplus, then we should do whatever the style guide requires.

include/libroutines.h
mwinkel-dev: I concur that `MDS_FFI_2_FIXED_ARGS` is a better name than `VA_2_FIXED_ARGS`.

If we retain the enum (or the defines), it will be easy to change the prefix of the member names.

These enums were only intended for use with the new `LibCallgFfi()` function, hence the enums were named accordingly.

The simplest solution is to eliminate the enums and using integer values.

include/libroutines.h
WhoBrokeTheBuild: But are the only options 1, 2, or 3 fixed args? Doesn't it go up to 32? Why would we limit it with an enum that just maps to the numbers they describe? If we're going off of what already exists in the code base, we would just pass an `int`.

include/libroutines.h
WhoBrokeTheBuild: For the FIXED_ARGS ones yes, but for the RTN ones we should add the prefix. We may only intend them for `LibCallgFfi` but they'll be in the global namespace and they're in the public `include/` folder.

mwinkel-dev: Yes, `RTN_VOID` is a better name than `RTN_NONE`.   I will make that change.

javamds/mdsobjects.c
mwinkel-dev: Yes, if you feel strongly about this it can be converted to a macro.

A digression . . . 

It is impressive how much is accomplished with macros in the MDSplus source code.   Some of that is necessary because C lacks the template constructs of other languages.   However, the extensive use of macros can also complicate debugging and maintenance work.

During the early prototypes of the Apple Silicon port, it was useful to have these #ifndef blocks.   Now that the port is stable, there isn't much risk in condensing them to a macro.

mdsshr/librtl.c
mwinkel-dev: The new `LibCallgFfi()` routine follows the original `LibCallg()` routine as closely as possible.

The `if (!routine)` exists in the original `LibCallg()` routine.

There are three options for dealing with this (all of which are easy to implement).

1.  Only convert the new routine to `assert()`.
2. Convert both the new and old routines to `assert()`.
3. Leave as is.

mdsshr/librtl.c
mwinkel-dev: Correct, the first argument is the number of arguments.   The `num_args` line is patterned after a similar construct in a `switch` statement at the top of the original `LibCallg()` function.

There is a later PR that adds clarifying comments to a few routines, including this one.   If needed, I can extract that comment and add it to this PR.

However, the more churn there is on PRs, the greater the risk that errors will creep in during this initial check-in of the Apple Silicon code.

mdsshr/librtl.c
mwinkel-dev: Yes, `SIZE` is a lousy name.   I will change it.

The 32 is the maximum number of arguments handled by the original `LibCallg()` function.   Because the new function, `LibCallgFfi()` is patterned after the old, it too has a limit of 32 total arguments.

Because this restriction is not related to FFI, the new name will be generic.    Perhaps, `MAX_LIBCALLG_ARGS`.

mdsshr/librtl.c
mwinkel-dev: The `printf()` is from the original `LibCallg()`.

There are three options for dealing with this situation (all of which are easy to implement).

1. Fix the new function's print statement.
2. Fix both the new and old function's print statements.
3. Leave as is.

mdsshr/librtl.c
mwinkel-dev: Yes, the reorganization adds clarity.   I will make this change.

mdsshr/librtl.c
mwinkel-dev: Alternatively, we could return `MDSplusFATAL`.   Which do you prefer?

mitdevices/t4012.c
mwinkel-dev: Yes, we should either use an `assert` or through a `MDSplusFATAL` error.   Which do you prefer?

include/libroutines.h
mwinkel-dev: There are many calls of `LibCallg()` scattered throughout the original source code to call various variadic functions.   It just so happens that those variadic functions essential to the operation of MDSplus only involve 1, 2 or 3 fixed arguments.

1 fixed arg = all TDI intrinsics
2 fixed args = all _Tdi* functions (these are helper functions associated with TDI intrinsics)
3 fixed args = just one function (that I don't remember at the moment).

So an enum (or defines) with just the four values (0 = not variadic, 1 fixed arg, 2, 3) covers everything associated with the source code that directly calls LibCallgFfi().   Indeed, because that is such a limited set of variadic functions, I was able to port all that code to Apple Silicon without using libffi.   [However, it was a hassle to do so and resulted in three additional flavors of `LibCallg()`.   When the new LibCallgFfi() function was written at the tail end of the project and proven to be stable, then it was used to replace the three variants of `LibCallg()`].

The scenario of up to 32 total arguments only applies when TDI is using the `->` (aka `BUILD_CALL`) construct to call user supplied variadic functions in external libraries.   The user can supply any variadic function they wish, which is why the `lex` change is required to allow the user to specify the number of fixed args to use for that variadic call.

include/libroutines.h
mwinkel-dev: Very good point about the `RTN` names.   They definitely need the `MDS_FFI_` prefix.   I will make that change.

include/libroutines.h
mwinkel-dev: *CLARIFICATION:*

For Apple Silicon, the only requirement for using `libffi` is to pass in the number of fixed arguments on variadic functions.

However, it is conceivable that future platforms might require additional steps.   For example, suppose that MDSplus is ported to an unusual embedded processor.   In that case, in addition to specifying the number of fixed variables on variadic functions, there might also be some associated memory management tasks that are required before or after calling `libffi`.

We should only implement `MDSPLUS_USE_FFI` if we are confident that the Apple Silicon code (in the present `#ifdef MACOS_ARM64` blocks) can apply to multiple platforms (i.e., that the scenario described above is unlikely to arise).

include/libroutines.h
mwinkel-dev: Aside from the above digression, we are in agreement that the enum will be removed and the VA_<N>_FIXED_ARGS will be replaced with integers.   That change will be added to this PR.

include/libroutines.h
mwinkel-dev: As discussed on 11-Mar-2025 at the MDSplus software team meeting, we are all in agreement to switch to the `MDSPLUS_USE_FFI` define.   (We consider the scenario described in the "clarification" above to be unlikely to arise.).  This PR will be revised accordingly.

mdsshr/librtl.c
mwinkel-dev: As per the 11-Mar-2025 meeting of the MDSplus software team, the `assert()` will be added to the new routine, `LibCallgFfi()`.   This PR will be revised accordingly.

A new PR will be created to make similar changes to the old `LibCallg()` routine.

mdsshr/librtl.c
mwinkel-dev: The comment will be added to this PR.

mdsshr/librtl.c
mwinkel-dev: This PR will be revised as suggested.

And as per the 11-Mar-2025 meeting of the MDSplus software team, a new PR will be created to make similar changes to the old LibCallg() routine.

mdsshr/librtl.c
mwinkel-dev: This PR will be revised as suggested.

And as per the 11-Mar-2025 meeting of the MDSplus software team, a new PR will be created to make similar changes to the old LibCallg() routine.

mdsshr/librtl.c
mwinkel-dev: As per the 11-Mar-2025 meeting of the MDSplus software team, the assert() will be added to the new routine, LibCallgFfi().  This PR will be revised accordingly.

mitdevices/t4012.c
mwinkel-dev: As per the 11-Mar-2025 meeting of the MDSplus software team, the assert() will be added to this PR.

javamds/mdsobjects.c
WhoBrokeTheBuild: I want to give it a try to see what it looks like, but I could easily be swayed either way. I think with the modern clangd tooltips and such, the macros are less confusing while debugging, but they're still a layer of indirection. I'll play with it after this is merged, and might put in a follow up PR that we can look at

mwinkel-dev: Yes, using macros is a better solution.   This PR will be revised accordingly.

mwinkel-dev:
All suggested revisions have been made to this PR.   (It is expected that upon re-review, some additional changes might be needed.)

Regarding the `mitdevices/t4012.c` file, refer to the comment added to the first `LIB_CALL_G()` statement.

mitdevices/t4012.c
```diff
@@ -434,15 +436,18 @@ static int AccessTraq(InStoreStruct * setup, int data, int memsize, void *arglis
   for (try = 0; (try < 30) && (!(CamQ(0) & 1)) && (STATUS_OK); try++) {
     if (arglist && !called) {
       called = 1;
-      LibCallg(arglist, routine);
+      assert(routine == &TdiData);
+      // Only called with TdiData() which is an intrinsic thus MDS_FFI_RTN_INT32
+      status = LIB_CALL_G(arglist, routine, 1, MDS_FFI_RTN_INT32);
```
mwinkel-dev, line 441: The original code did not store the result of `LibCallg()` in a variable.   However, when replaced that with the `LIB_CALL_G()` macro, the compiler on Ubuntu22 x86-64 alpha stopped on that line because it treated that warning as an error.   (The cmake branch didn't have any compiler issues with this file.).  So added `status = ` to the statement.

Note that this is a change to the original code and might affect the status returned by the `AccessTraq()` method of the device.

mwinkel-dev:
Retest this please

mwinkel-dev:
Retest this please

mwinkel-dev:
Retest this please

include/libroutines.h
```diff
@@ -16,6 +16,24 @@
 /// defined in MdsShr.so
 
 extern void *LibCallg();
+
+// The following define must be available on all platforms, thus cannot be inside the MDSPLUS_USE_FFI ifdef
+#define MDS_BYPASS_FFI 0
```
WhoBrokeTheBuild: I don't know if this needs to be a #define'd constant

include/libroutines.h
```diff
@@ -16,6 +16,24 @@
...
+extern void *LibCallgFfi();
+enum MDS_FFI_Rtype { 
+    MDS_FFI_RTN_VOID = 0, 
+    MDS_FFI_RTN_POINTER = 1, 
+    MDS_FFI_RTN_INT32 = 2,
+    MDS_FFI_RTN_INT64 = 3
+};
+#endif
+#ifdef MDSPLUS_USE_FFI
+#define LIB_CALL_G(args, routine, fixed, rtype) (int)(intptr_t)LibCallgFfi(args, routine, fixed, rtype)
```
WhoBrokeTheBuild, line 30: The wrapper macro looks good :slightly_smiling_face:

mdsshr/librtl.c
```diff
@@ -280,6 +287,77 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
...
+  // arglist[0]        = N (total number of args excluding first and last elements of vector)
+  // arglist[1]        = expression
+  // arglist[2 .. N-1] = variable args (pointer to descriptors)
+  // arglist[N]        = result xd1 descriptor
+  // arglist[N+1]      = NULL 
+  int num_args = *(int *)a & 0xff;
+
+  // Usually the check for non-variadic routines,and the associated call of LibCallg(), is
+  // done prior to calling this function.
+  if (num_fixed_args == MDS_BYPASS_FFI) {
```
WhoBrokeTheBuild: Yeah I think `0` here would be perfectly fine

mitdevices/t4012.c
WhoBrokeTheBuild: Oh interesting, it was probably mad that we were casting it with `(int)(intptr_t)` and then discarding the value.

WhoBrokeTheBuild (dismissed):

tdishr/TdiCall.c
```diff
@@ -65,63 +65,108 @@ extern int tdi_put_ident();
 _Pragma("GCC diagnostic ignored \"-Wcast-function-type\"")
```
mwinkel-dev, line 65: **SCRUTINIZE:** The changes to this file must be scrutinized because ALL platforms use the TdiCall intrinsic. Changes in this file, and in particular to the `interlude()` function, thus have the potential to affect ALL platforms.

tdishr/TdiCall.c
```diff
@@ -65,63 +65,108 @@ extern int tdi_put_ident();
...
+    *max = 0;
+#ifdef MDSPLUS_USE_FFI
+    if (num_fixed_args == MDS_BYPASS_FFI) {
+      LibCallg(newdsc, routine);
+    } else {
+      LibCallgFfi(newdsc, routine, num_fixed_args, MDS_FFI_RTN_VOID);
+    }
+#else
     LibCallg(newdsc, routine);
+#endif
```
mwinkel-dev, line 81: **NOTE:** In the above `#ifdef` construct, it is reasonable to ask why there are two calls to `LibCallg()`. The above code could be reorganized.

However, some things to keep in mind:
- The `LibCallgFfi()` routine does not exist in the source code if `MDSPLUS_USE_FFI` is not defined.
- On Apple Silicon, the `BUILD_CALL` (aka `->`) calls to functions in external libraries only need the `#<num_fixed_args>` suffix added if `<routine>` is variadic. Stated another way, many external functions are NOT variadic, in which case it is best just to call the plain `LibCallg()`.

tdishr/TdiCall.c
```diff
@@ -65,63 +65,108 @@ extern int tdi_put_ident();
 _Pragma("GCC diagnostic ignored \"-Wcast-function-type\"")
 #endif
 
-    static inline int interlude(dtype_t rtype, mdsdsc_t **newdsc,
+    static inline int interlude(dtype_t rtype, int num_fixed_args, mdsdsc_t **newdsc,
                                 int (*routine)(), void **result, int *max)
 {
```
mwinkel-dev, line 71: **SCRUTINIZE:**   All platforms use the `interlude()` function.   Thus all changes (or introduced bugs) this PR makes to this function could affect ALL platforms.

tdishr/TdiCall.c
```diff
@@ -65,63 +65,108 @@ extern int tdi_put_ident();
...
+    } else {
+      char *dup_entry = strdup(strtok(c_entry, "#"));
+      new_entry.length = strlen(dup_entry);
+      new_entry.pointer = dup_entry;
+      status = TdiFindImageSymbol(image.pointer, &new_entry, proutine);
+      free(dup_entry);
+    }
+    free(c_entry);
+  }
+
```
mwinkel-dev, line 161: **SCRUTINIZE:** The above code must be scrutinized.   If it contains bugs, it can affect ALL platforms.

tdishr/TdiCall.c
```diff
@@ -162,8 +207,22 @@ int tdi_call(dtype_t rtype, int narg, mdsdsc_t *list[], mdsdsc_xd_t *out_ptr, cl
...
+  int num_fixed_args = MDS_BYPASS_FFI;
+  if (token != NULL) {
+    num_fixed_args = atoi(token);
+  }
+  free(dup);
+  #ifndef MDSPLUS_USE_FFI
+  num_fixed_args = MDS_BYPASS_FFI;  // bypasses libFFI for all other platforms
+  #endif
+  
   *(int *)&newdsc[0] = narg - 2;
```
mwinkel-dev, line 224: **SCRUTINIZE:**   The above code must be scrutinized.   If it contains bugs, it can affect ALL platforms.

tdishr/TdiCall.c
```diff
@@ -245,7 +304,7 @@ int tdi_call(dtype_t rtype, int narg, mdsdsc_t *list[], mdsdsc_xd_t *out_ptr, cl
     }
   }
   if (STATUS_OK)
-    status = interlude(rtype, newdsc, routine, (void **)result, &max);
+    status = interlude(rtype, num_fixed_args, newdsc, routine, (void **)result, &max);
   if (!out_ptr)
```
mwinkel-dev, line 306: **SCRUTINIZE:**   All platforms use the `interlude()` function.   Adding `num_fixed_args` to it thus has the potential to affect ALL platforms.

tdishr/TdiLex.l
```diff
@@ -159,7 +159,7 @@ int	([[:digit:]][[:alnum:]]*)
 float	({f0}|{f1}|{f2}|{f3})
 ident	([[:alpha:]$_]{anum}*)
 path	(((("\\"({ident}"::")?|[.:])?{ident})|(".-"("."?"-")*))([.:]{ident})*)
-libcall	("->"{anum}+(":"|"..")?)
+libcall	("->"{anum}+("#"{int})?(":"|"..")?)
 comment ("/*"(\n|[^\*]+|\*+[^/\*])*\*+"/")
```
mwinkel-dev, line 163: **NOTE:**  When this change is made to alpha, the potential exists for it to affect ALL platforms.   Stated another way, the changes to `TdiCall.c` should be free of bugs prior to activating this `lex` change.

mdsshr/librtl.c
mwinkel-dev: As we discussed via Slack, look at the changes in `TdiCall.c`.   We can then decide whether to keep the `MDS_BYPASS_FFI` or not.   If we keep it, then we can decide whether its value should be 0, -1, MAX_INT, or something else.

tdishr/TdiCall.c
mwinkel-dev: Not a surprise, but Jenkins also choked on this code.   So refactoring it right now so has just one `LibCallg()` and so the compiler doesn't complain about `num_fixed_args` being unused.

mwinkel-dev: The code has been refactored.

Porting MDSplus to Apple Silicon involved many phases.   The two `LibCallg()` functions per block were an unintended consequence of that process.

mwinkel-dev:
Retest this please

tdishr/TdiCall.c
mwinkel-dev: Note that `num_fixed_args` is an "overloaded" variable.   It can contain two types of information:
- the `MDS_BYPASS_FFI` flag (presently is 0, but could be changed to anything outside of the valid range).
- for variadic functions, is the number of fixed arguments (i.e., the range 1 .. `LIBCALLG_MAX_ARGS`).

An alternate design would be to add a `<bypass_ffi>` parameter to the `interlude()` function.   If `TRUE` it calls `LibCallg()`; if `FALSE` it calls `LibCallgFfi()`.   This avoids overloading the `num_fixed_args` parameter.   However, the valid range for `num_fixed_args` remains the same: 1 .. `LIBCALLG_MAX_ARGS`.

**EDGE CASE:**   If `num_fixed_args` is equal to `LIBCALLG_MAX_ARGS` (or more likely `LIBCALLG_MAX_ARGS - 4`), then all arguments to the function are fixed args.  Because there are no varargs, the function is not variadic.   In this scenario, the code should bypass `LibCallgFfi()` and instead just call the ordinary `LibCallg()`.   This edge case was not considered or tested during development of the Apple Silicon port.   Thus, this edge case is likely a latent bug in this PR.

*DIGRESSION:*   The number of arguments available to the external function being called is actually less than `LIBCALLG_MAX_ARGS` because of the four overhead arguments included in that count.   Refer to the comments about the "arglist[]" added at the top of `LibCallgFfi()`.

mdsshr/librtl.c
mwinkel-dev: **NOTE:**  The above comments describe the argument lists for calling TDI intrinsic functions.   For external third-party functions, the `expression` is just another argument.   Same is probably true of the "result xd descriptor.

mdsshr/librtl.c
```diff
@@ -280,6 +287,77 @@ EXPORT void *LibCallg(void **const a, void *(*const routine)())
   return 0;
 }
 
+
+#ifdef MDSPLUS_USE_FFI
+EXPORT void *LibCallgFfi(void **const a, void *(*const routine)(), int num_fixed_args, int rtype)
+{
```
mwinkel-dev, line 297: **NOTE:** This function assumes all arguments are pointers.   That approach works well for the TDI intrinsic functions.  

However, it is possible / probable that many third-party external functions will have different requirements.   Thus this `LibCallgFfi()` is not a general purpose solution.   Depending on what the external function requires, it might be necessary for the user to write their own custom shim.

mwinkel-dev:
Retest this please

mdsshr/librtl.c
mwinkel-dev: As part of this PR, I will add comments to this routine to explain its limitations.

mdsshr/librtl.c
mwinkel-dev: As mentioned in the GitHub comment regarding the `interlude()` function of `TdiCall.c`, the `MDS_BYPASS_FFI` could also be moved to a separate parameter (instead of overloading `num_fixed_args` as shown above).

mdsshr/librtl.c
mwinkel-dev: The original `LibCallg()` also shares many (all?) of these limitations.  In particular if TDI's `VAL()` function is used to pass a parameter by value, the resulting value must not exceed the size of a pointer.   Refer to the `OPC_VAL`  section of `tdi_call()` in `TdiCall.c`.

tdishr/TdiCall.c
```diff
@@ -65,63 +65,100 @@ extern int tdi_put_ident();
...
-    int64_t *result_q = (int64_t *)result;
+    // 8 bytes
     *max = sizeof(int64_t);
-    *result_q = called_q(newdsc, routine);
+    int64_t *result_q = (int64_t *)result;
+    int64_t (*called_q)() = (int64_t(*)())LibCallg; 
+    if (num_fixed_args == MDS_BYPASS_FFI) {
+      *result_q = called_q(newdsc, routine);
+    } else {
+#ifdef MDSPLUS_USE_FFI
```
WhoBrokeTheBuild, line 112: I think this #ifdef needs to be restructured, because I believe it's possible to end up in this function with `num_fixed_args != MDS_BYPASS_FFI` and as is I think it would fail to call the routine

tdishr/TdiCall.c
```diff
@@ -65,63 +65,100 @@ extern int tdi_put_ident();
...
+    // 4 bytes
     *max = sizeof(int32_t);
-    *result_int = called_int(newdsc, routine);
-  }
+    int32_t *result_int = (int32_t *)result;
+    int32_t (*called_int)() = (int32_t(*)())LibCallg; 
+    if (num_fixed_args == MDS_BYPASS_FFI) {
+      *result_int = called_int(newdsc, routine);
+    } else {
+#ifdef MDSPLUS_USE_FFI
```
WhoBrokeTheBuild, line 127: Same comment as above

tdishr/TdiCall.c
```diff
@@ -65,63 +65,100 @@ extern int tdi_put_ident();
...
-    status = TdiFindImageSymbol(image.pointer, entry.pointer, proutine);
+ 
+  // Given "<function_name>#<num_fixed_args>" extract just the function name
+  if (STATUS_OK) {
+    char *c_entry = MdsDescrToCstring(entry.pointer);
+    char *hash_ptr =strrchr(c_entry, '#');
+    if (hash_ptr == NULL) {
+      status = TdiFindImageSymbol(image.pointer, entry.pointer, proutine);
+    } else {
+      char *dup_entry = strdup(strtok(c_entry, "#"));
```
WhoBrokeTheBuild: I believe we have "ownership" of `c_entry` here, so we can just use `*hash_ptr = '\0'` to truncate the name, and avoid the calls to `strdup/strtok`

tdishr/TdiCall.c
```diff
@@ -162,8 +199,22 @@ int tdi_call(dtype_t rtype, int narg, mdsdsc_t *list[], mdsdsc_xd_t *out_ptr, cl
   char result[8] = {0}; // we need up to 8 bytes
   unsigned short code;
   mdsdsc_t dx = {0, rtype == DTYPE_C ? DTYPE_T : rtype, CLASS_S, result};
-  unsigned char origin[255];
+  unsigned char origin[255];   // (MW) TODO: Should this be 256?
   mdsdsc_t *newdsc[256] = {0};
+
+  // Given "<function_name>#<num_fixed_args>" extract just the number of fixed args
+  char *dup = strdup(list[1]->pointer);
+  char *token = strtok(dup, "#");
```
WhoBrokeTheBuild: We should either avoid calls to `strtok` (my preference) or switch to using `strtok_r`

tdishr/TdiCall.c
```diff
@@ -162,8 +199,22 @@ int tdi_call(dtype_t rtype, int narg, mdsdsc_t *list[], mdsdsc_xd_t *out_ptr, cl
   char result[8] = {0}; // we need up to 8 bytes
   unsigned short code;
   mdsdsc_t dx = {0, rtype == DTYPE_C ? DTYPE_T : rtype, CLASS_S, result};
-  unsigned char origin[255];
+  unsigned char origin[255];   // (MW) TODO: Should this be 256?
```
WhoBrokeTheBuild: Great question, I'm not sure how we would confirm either way

WhoBrokeTheBuild (changes requested):

tdishr/TdiCall.c
mwinkel-dev: There is an interlock with the `MDSPLUS_USE_FFI` section at the beginning of the `tdi_call()` function in `TdiCall.c`.

If `MDSPLUS_USE_FFI` is undefined, the code in `tdi_call()` forces `num_fixed_args` to be `MDS_BYPASS_FFI`.   Thus the original `called_q()` routine is used.

However, if `MDSPLUS_USE_FFI` is defined, then `tdi_call()` sets it to the `#<number>` suffix from the `BUILD_CALL` (aka `->`).   If there is no `#<number>`, then `tdi_call()` defaults it to `MDS_BYPASS_FFI`, thus it still goes through the original `called_q()` routine.

However, if there is a `#<number>` and `MDSPLUS_USE_FFI` is defined, then the `else` block will be expanded with the `#ifdef` and thus the `LibCallgFfi()` variant will be called.

The above code passes 100% of the test suite on 4 configurations:
- MacOS-Silicon / cmake branch
- MacOS-Intel / cmake
- Ubuntu22-Intel / cmake
- Ubuntu22-Intel / alpha

*CAVEAT:*   The `tdi_call()` assumes that users will not have `<function>#0` or `<function>#<negative_number>` in their `BUILD_CALL` (aka `->`) calls to external variadic functions.   I will upgrade this PR to check for those edge cases.

*HOWEVER:*  Please pressure-test all the above assertions.   The scenario you described might exist.   If you have an example, post it here.   Thanks.

tdishr/TdiCall.c
mwinkel-dev: A-OK, I will make that change to this PR.

tdishr/TdiCall.c
mwinkel-dev: A-OK, I'll make this change to the PR.

mdsshr/librtl.c
mwinkel-dev: Should I add a second parameter to the `interlude()` function so that `num_fixed_args` is not overloaded?

I have a weak preference to leave it "as is", but I can be easily persuaded either way (overloaded single parameter, or two separate parameters).

include/libroutines.h
mwinkel-dev: The `MDS_BYPASS_FFI` is discussed in other comments in this PR.   So marking this as resolved.

tdishr/TdiCall.c
mwinkel-dev: See reply to above post.

tdishr/TdiCall.c
mwinkel-dev: I found the answer in the `TdiDefFunction.c` code.   That code limits the argument list to 255 actual parameters plus one result (aka `OUTPTR`) parameter.

Because `TdiDefFunction.c` limits the argument list to 255 or less, that also explains why this code doesn't check for exceeding the bounds of the two arrays: `origin[]` and `newdsc[]`.

mwinkel-dev:
Retest this please

mdsshr/librtl.c
mwinkel-dev: I have decided to add a `bypass_ffi` boolean parameter to the `interlude()` function.   Doing so has the following benefits: adds clarity, eliminates the `MDS_BYPASS_FFI` define, and makes it a bit easier to check for edge cases involving `num_fixed_args` containing bogus values.   

I will add that change to this PR.

mwinkel-dev:
Retest this please

tdishr/TdiCall.c
mwinkel-dev: Implemented the alternate design described above, thus `num_fixed_args` is no longer overloaded.

tdishr/TdiCall.c
mwinkel-dev: Replaced the `MDS_BYPASS_FFI` define with a boolean parameter, `bypass_ffi`.   That not only made it easier for `LibCallgFfi()` to test for the edge cases described in the caveat, it also adds clarity to this section of code.

tdishr/TdiCall.c
```diff
@@ -65,63 +65,101 @@ extern int tdi_put_ident();
 _Pragma("GCC diagnostic ignored \"-Wcast-function-type\"")
 #endif
 
-    static inline int interlude(dtype_t rtype, mdsdsc_t **newdsc,
+    // num_fixed_args is unused on all platforms except those that use libffi
```
mwinkel-dev, line 68: The compiler aborted the build on `alpha` for Ubuntu22 x86-64 because the `num_fixed_args` parameter was unused.   So added the `__attribute__((unused))` to the following line.   However, that should probably be replaced with a `_Pragma()`.

mdsshr/librtl.c
mwinkel-dev: The change is now in this PR.

tdishr/TdiCall.c
mwinkel-dev: The "edge case" analysis is not quite correct.

A better analysis is that `num_fixed_args` must always be in the range 1 to `nargs`.   And that `LibCallgFfi()` is not designed to handle `nargs` greater than `LIBCALLG_MAX_ARGS`.   (The original `LibCallg()` function has the same limitation.)   The `LibCallgFfi()` function now has better validity checks.

It is possible that a variadic function might only be called with its fixed arguments.   In which case, `num_fixed_args` = `nargs`.   Presumably, in that scenario the call of the external function could be handled by the original `LibCallg()` without causing a segfault.   And also can be handled by the new `LibCallgFfi()`, but with a bit more overhead.   This scenario will be tested soon.

mdsshr/librtl.c
mwinkel-dev: This PR now has comments explaining the limitation.

tdishr/TdiCall.c
mwinkel-dev: Change made to this PR.

During the porting process, the biggest concern was inadvertently altering existing data structures thus resulting in more debugging effort.   Unfortunately, that very conservative strategy also meant that some inefficient and unnecessary code was left in the Apple Silicon port.

tdishr/TdiCall.c
mwinkel-dev: The change has been made in this PR.

tdishr/TdiCall.c
mwinkel-dev: Added a comment to the code to explain why it has `origin[255]` and `newdsc[256]`.

mwinkel-dev:
The jScope I/O problem should be fixed before this PR is merged to alpha.

mdsshr/librtl.c
mwinkel-dev: Edited the comment accordingly.

mwinkel-dev:
jScope on the `cmake` branch requires the `MDSIP_CLIENT_LOCAL_LOGFILE` environment variable to be set.

jTraverser2's `File->Connect` menu item throws an error because the `mdsip_server` script has the compression level set to `-c 9`.   However, works fine with the compression off, `-c 0`.

WhoBrokeTheBuild (dismissed):

mwinkel-dev:
The `TdiExtFunction.c` now only executes TDI `*.fun` files.   The `MdsExpressionCompileTest.cpp` now uses `->` instead of `ext_function()`.

If users attempt to use the removed `EXT_FUNCTION(<library>, <function>, ...)` feature, they will see a message like the following one.

```
TDI> ext_function('mylib','myfunction')

** The EXT_FUNCTION(<library>, <function>, ...) feature has been removed.  **
** Please rewrite using the BUILD_CALL (aka ->) function. **

%TDI Error in EXT_FUNCTION("mylib", "myfunction")
%TDI Error in EXECUTE("ext_function('mylib','myfunction')")

TDI> 
```

mwinkel-dev:
The `TdiExtFunction.c` changes were made in this PR just to ensure that they are compatible with the Apple Silicon changes.

However, if there is a chance that the `TdiExtFunction.c` changes will have to be reverted at some future date, it would be easier to do so if the changes are instead submitted as part of PR #2882.   In that case, this PR #2877 should be merged first and then PR #2882.

mwinkel-dev:
Retest this please

mwinkel-dev:
Retest this please

mwinkel-dev:
Retest this please

mwinkel-dev:
Retest this please

WhoBrokeTheBuild (approved):

PR #2878, opened 2025-03-09
Fix: Apple Silicon - specify default for PyLib for MacOS

mwinkel-dev:
On 14-Mar-2017, commit `433cb19` added hardcoded default paths to be used if the user forgets to set the `PyLib` environment variable.

This PR, updates the hardcoded path used by MacOS (both Apple Silicon and Intel).

Note that portions of MDSplus require Python 2.7 and a compatible Numpy to be installed.   This is problematic on Apple Silicon because Apple now ships Python 3.x with MacOS (pretty sure Apple no longer ships Python 2.7).   Thus one must use package managers, such as Brew or MacPorts, to install the older Python and Numpy.   The location of the Python library therefore depends on which package manager is installed.

This PR uses the path for Python 2.7 installed by MacPorts.

There are two strategies for dealing with the hardcoded paths:

1. Revise this PR to include a new default path mechanism if the PyLib environment variable is missing.
2. Or approve this PR "as is", get the rest of the Apple Silicon port checked-in, and revise the paths when configuring the Jenkins build system to generate Apple Silicon releases of MDSplus.

### DIFF
--- a/tdishr/TdiExtPython.c
+++ b/tdishr/TdiExtPython.c
@@ -129,7 +129,11 @@ inline static void initialize()
     _putenv_s("PyLib", envsym);
 #else
     envsym = "python2.7";
+#ifdef MACOS_ARM64
+    const char *aspath = "/opt/local/lib/libpython2.7.dylib";   // (MW) TODO: for MacPorts version
+#else 
     const char *aspath = "/usr/lib/python2.7.so.1";
+#endif
     setenv("PyLib", envsym, B_FALSE);
 #endif
     fprintf(stderr,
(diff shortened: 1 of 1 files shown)


WhoBrokeTheBuild (approved):
This is certainly a weird one.

We by no way require python 2.7, that was just the hardcoded default when the python integration was written and it's stayed that way. If there was a reliable path on macosx for some libpython, regardless of the version, I would say we use that.

In reality, you *need* to define `$PyLib` on modern systems or nothing works properly, so I don't think we need to worry too much about the default.

I'm approving this PR mostly for the second half that adds the `.dylib`. The first half that sets a default path for python 2.7 will probably never be used, and if it is it is likely by mistake.

PR #2881, opened 2025-03-12
Fix: errors with asan and tsan

WhoBrokeTheBuild:
Update `strtok` to `strtok_r` in `ReplaceAliasTrees` to prevent issues when multithreading. The internal `saveptr` was being clobbered, so this moves it into the function, and therefore safely in the thread.
Reorder the `io_disconnect` function for the thread backend. The cleanup of the last request/response messages were being interrupted by the `pthread_cancel`, which is fixed by moving the `pipe_close` calls above it, matching the windows implementation.


heidthecamp (approved):
Nice catch with the close_pipe

mwinkel-dev (approved):
Spotting that strtok() is not thread safe and switching to reentrant strtok_r() is excellent.   As is moving the close_pipe() statements.

PR #2886, opened 2025-03-18
Feature: Add arm64 arch to the build system

santorofer:
Replaces #2885 

Instead of adding a new block to the $host switch statement, this just extends the x86/x86_64 section. This is a good enough solution given that cmake is coming soon.

This does not package LabView for arm64.


deploy/os/ubuntu24-arm64.opts
```diff
@@ -0,0 +1 @@
+--platform=debian --dockerimage=mdsplus/builder:ubuntu-24.04-arm64 --distname=Ubuntu24 --arch=arm64 --disable-labview
```
WhoBrokeTheBuild: Just a note, this is also using the CMake docker image for Ubuntu 24.04 arm64

heidthecamp:
Is this in the cmake branch already? If so we may want to just wait for cmake to merge this since our build system will be largely overhauled.

heidthecamp:
This is blocked by jenkins. arch64 does not build on builder 2 currently and we cannot add our Orin builder to the current jenkins config. This feature will roll out with cmake.

Issue #2887, opened 2025-03-20
The remote_submit_helper.fun file needs to be updated to allow zero connection id


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
MDSplus version: 7.142.80
  Release:  alpha_release-7-142-80
  Date:     Mon May 20 22:56:09 UTC 2024

**Platform(s)**
Ubuntu18

**Installation Method(s)**
From the *.deb packages

**Describe the bug**
PR #2288 changed the `INVALID_CONNECTION_ID` to `-1` (it used to be `0`).   This means that `0` can now be returned as a valid connection ID.   Unfortunately, the `remote_submit_helper.fun` was not updated to reflect the PR #2288 change, and thus it got stuck in a loop that caused the `mdsip_submit` log file to grow to 128 GB and fill the disk volume.

**To Reproduce**
1. Either reboot the server or restart the mdsip service on port 8000.   This ensures that there are no prior mdsip connections and that the connection table is empty.
2. Run a shot that will "submit" a task to the server.   It will be assigned connection ID = 0, which will be returned by the `remote_submit_queues.fun` to `remote_submit_helper.fun`.
3. Do a `tail -f mdsip_submit` to monitor the associated log file.   It will have an endless stream of `Trying <server>` messages.
4. Inspect the `remote_submit_helper.fun` file and note that it has a `while (_stat == 0)` loop.   The intent of the loop is to repeat until a "valid" connection ID is obtained.   However because of PR 2288, the loop should actually be `while (_stat == -1)`.

**Expected behavior**
Running a shot doesn't cause a log file to fill the disk.

**Screenshots**
n/a

**Additional context**
Regarding PR #2258, refer to these files to see the change.
- `mdstcpip/mdsipshr/Connections.c` has an `AddConnection()` function that used to declare `static int id = 0` prior to the PR.   However, that statement changes to `static int id = INVALID_CONNNECTION_ID` with the PR.
- `mdstcpip/mdsip_connections.h` defines `INVALID_CONNECTION_ID` to be `-1`

mwinkel-dev:
This fix is needed for General Atomics, thus is "US Priority".

mwinkel-dev:
These additional changes would be useful.
- Have the loop terminate after T seconds or N loops.   Perhaps 300 seconds (5 minutes) or 10,000 iterations would be good values.
[...]

PR #2889, opened 2025-03-20
Fix: remote_submit_helper.fun now accepts zero id and has a timeout

mwinkel-dev:
Fixes Issue #2887.

PR #2288 redefined `INVALID_CONNECTION_ID` to be `-1`, which means that `0` is a valid connection id.   Changed `remote_submit_helper.fun` accordingly.

Also added a timeout to the loop, so that it doesn't run forever and fill up the disk with a 128 GB log file.

The timeout value is 300 seconds (5 minutes) which is likely too generous.   (Experiments on a M2 MacBook Pro show that the loop executes ~100K iterations in ~5 seconds.)

Did not rename the `_stat` variable because the `return(_stat)` statement also can be the status value of the `mdsvalue()` statement at the bottom of the file.

### DIFF
--- a/tdi/remote/MdsConnect.fun
+++ b/tdi/remote/MdsConnect.fun
@@ -13,6 +13,7 @@ call:	_host  = Host name eg elpp1.epfl.ch or elpp1.epfl.ch:9000
       write(*,"Host taken from MDS_HOST ["//_host//"]");
    }
    _status = TdiShrExt->rMdsConnect(ref(_host//char(0)));
-   if (present(_abort) && _status == 0) abort();
+   _invalid_id = -1
+   if (present(_abort) && _status == _invalid_id) abort();
    return(_status);
 }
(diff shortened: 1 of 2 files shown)


mwinkel-dev:
Consider cross-version compatibility when reviewing this PR.

Some sites might have this configuration:
- archive server running old version of MDSplus that has `0` as the invalid shot id
- client computers running new version of MDSplus that has `-1` as the invalid shot id

However, the switch to `-1` happened on 19-May-2021, thus sites with the above configuration would have hit various issues associated with the change in definition of `INVALID_CONNECTION_ID` by now, so perhaps have already upgraded their archive servers to a new version of MDSplus.


ADDENDUM:   The opposite scenario should also be considered, namely archive server running new MDSplus and client computers running older MDSplus.   This is the configuration at one site that months ago reported API problems associated with the change in definition of `INVALID_CONNECTION_ID`.

WhoBrokeTheBuild (dismissed):

mwinkel-dev:
Retest this please

WhoBrokeTheBuild (approved):

PR #2892, opened 2025-03-27
Fix: LibCallg() now has similar edits to LibCallgFfi()

mwinkel-dev:
The original `LibCallg()` function has been updated with the same improvements that the reviewer requested on the new `LibCallgFfi()` function (which was added on PR #2877).

### DIFF
--- a/mdsshr/librtl.c
+++ b/mdsshr/librtl.c
@@ -168,9 +168,17 @@ EXPORT char *Now32(char *const buf)
 ///
 EXPORT void *LibCallg(void **const a, void *(*const routine)())
 {
-  if (!routine)
-    abort(); // intercept definite stack corruption
-  switch (*(int *)a & 0xff)
+  assert(routine);  // in DEBUG mode checks for stack corruption
+
+  // The "a" parameter is the arglist for the "routine" and contains the following:
+  // arglist[0]        = N (total number of args excluding first and last elements of vector)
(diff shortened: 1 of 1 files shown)


joshStillerman (commented):
I know that we have discussed this previously and there was some disagreement.  I do not think that the assert necessarily makes the check at run-time.  What I read about them is that they do this 'if enabled'.  They are more for assertions about things the programmer has control over.  So I would at a minimum also check routine for non zero, and I am not sure about having the assert.

heidthecamp (dismissed):
I like the convert from abort to assert. 

Adding the variable also makes this easier to read.

joshStillerman:
google says:
```
No, assert code does not always execute; its execution depends on the programming language and its configuration, often being disabled in production or optimized builds. 
```

WhoBrokeTheBuild (changes requested):
The assert should be changed back to an abort, or perhaps an `exit(1)`.

mwinkel-dev:
In light of the above, then the new `LibCallgFfi()` also has to be fixed.

mwinkel-dev:
Used `abort()` instead of `exit(1)` simply because the original `LibCallg()` routine used `abort()`.

With this PR, `abort()` will be used with both `LibCallg()` and the new `LibCallgFfi()`.

WhoBrokeTheBuild (approved):

PR #2893, opened 2025-03-27
Fix: ensure SsINTERNAL not treated as OK status

mwinkel-dev:
This is a partial fix for Issue #2741.  (There are other occurrences of `SsINTERNAL` in the source code that still need to be fixed.)

Note that `SsINTERNAL` is NOT an MDSplus error code.  Because `SsINTERNAL` has a value of `-1`, its low-order bit is set.   Therefore, if `SsINTERNAL` is passed to a macro that tests MDSplus error codes, the macro will erroneously treat `SsINTERNAL` as a flavor of "success".

The long-term fix is to eventually change `SsINTERNAL` to be `-2` or some other value that will be a valid MDSplus error code.   However, extensive cross-version testing will be needed before that switch can be made.

### DIFF
--- a/mdstcpip/mdsipshr/ConnectToMds.c
+++ b/mdstcpip/mdsipshr/ConnectToMds.c
@@ -90,6 +90,10 @@ static int do_login(Connection *c)
   else
   {
     Message *mrecv = GetMdsMsgTOC(c, &status, 10000);
+    // SsINTERNAL has low order bit set so is erroneously treated as OK.
+    if (status == SsINTERNAL) {
+      status = MDSplusERROR;
+    }
     if (STATUS_NOT_OK)
     {
       perror("Error during login: recv");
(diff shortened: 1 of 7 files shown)


mwinkel-dev:
Note that `SsINTERNAL` is used mostly in the "server" code.   Thus, the following routines can return a status of `SsINTERNAL`.

- send_bytes()
- get_bytes_to()
- SendMdsMsgC()
- GetMdsMsgTO()
- GetMdsMsgTOC()

After extensive cross-version testing is done and `SsINTERNAL` is upgraded to a true MDSplus error code, then all the special handling of `SsINTERNAL` in the "server" code can be cleaned up.

Purpose of this PR is just to make sure that communication errors are not swallowed and hidden by the status macros that treat `SsINTERNAL` as a "success".

WhoBrokeTheBuild (approved):

PR #2900, opened 2025-04-08
Fix: mdsip compression

santorofer:
- Set `dlen` after potentially flipping the bytes in `msglen`
- Correctly pass the compressed length to `uncompress()`
- Use the correct type (unsigned) for the compressed message length

### DIFF
--- a/mdstcpip/mdsipshr/GetMdsMsg.c
+++ b/mdstcpip/mdsipshr/GetMdsMsg.c
@@ -121,13 +121,15 @@ Message *GetMdsMsgTOC(Connection *c, int *status, int to_msec)
     {
       Message *m;
       memcpy(&msglen, msg->bytes, 4);
-      dlen = msglen - sizeof(MsgHdr);
       if (Endian(header.client_type) != Endian(ClientType()))
+      {
         FlipBytes(4, (char *)&msglen);
+      }
+      dlen = msglen - sizeof(MsgHdr);
       m = malloc(msglen);
       m->h = header;
       *status = uncompress((unsigned char *)m->bytes, &dlen,
-                           (unsigned char *)msg->bytes + 4, dlen - 4) == Z_OK;
+                           (unsigned char *)msg->bytes + 4, msg->h.msglen - 4 - sizeof(MsgHdr)) == Z_OK;
       if (IS_OK(*status))
       {
         m->h.msglen = msglen;
--- a/mdstcpip/mdsipshr/SendMdsMsg.c
+++ b/mdstcpip/mdsipshr/SendMdsMsg.c
@@ -112,7 +112,7 @@ int SendMdsMsgC(Connection *c, Message *m, int msg_options)
     cm->h = m->h;
     cm->h.client_type |= COMPRESSED;
     memcpy(cm->bytes, &cm->h.msglen, 4);
-    int msglen = cm->h.msglen = clength + 4 + sizeof(MsgHdr);
+    unsigned int msglen = cm->h.msglen = clength + 4 + sizeof(MsgHdr);
     MDSDBG(MESSAGE_PRI, MESSAGE_VAR(cm));
     if (do_swap)
       FlipBytes(4, (char *)&cm->h.msglen);


mdstcpip/mdsipshr/GetMdsMsg.c
```diff
@@ -121,13 +121,15 @@ Message *GetMdsMsgTOC(Connection *c, int *status, int to_msec)
...
       if (Endian(header.client_type) != Endian(ClientType()))
+      {
         FlipBytes(4, (char *)&msglen);
+      }
+      dlen = msglen - sizeof(MsgHdr);
       m = malloc(msglen);
       m->h = header;
       *status = uncompress((unsigned char *)m->bytes, &dlen,
-                           (unsigned char *)msg->bytes + 4, dlen - 4) == Z_OK;
+                           (unsigned char *)msg->bytes + 4, msg->h.msglen - 4 - sizeof(MsgHdr)) == Z_OK;
```
mwinkel-dev, line 132: Note that `msglen` is set to the first 4 bytes of the `MsgHdr`, which is synonymous with `msg->h.msglen`.   Thus, the original code `dlen - 4` is fine.

However, the new code, `msg->h.msglen - 4 - sizeof(MsgHdr))` is equivalent, so is also OK.

mwinkel-dev (approved):
Very good catch to notice that `dlen` has to be moved to after the potential `FlipBytes()` call if the client and server have different "endian" architectures.

The changes look good, so approving this PR.

mdstcpip/mdsipshr/GetMdsMsg.c
zack-vii: Actually, after looking at it for a while, only the new version is correct.
In short:
 because `msglen != msg->h.msglen`.

In long:
`int ZEXPORT uncompress(dest, destLen, source, sourceLen)`
`dlen` is extracted from the compressed data part and holds the `msglen` of the uncompressed message, i.e. `destLen`.
The original header still has the `msglen` of the compressed message, i.e. `sourceLen`.

Possibly the code would be easier to understand if one would use other variable names.
```c
unsigned int uncompressed_size = msglen - sizeof(MsgHdr);
unsigned int compressed_size = header.msglen - sizeof(MsgHdr) - 4;
```

Also would it be helpful to keep msg to be the designated output and always return NULL on error?
```c
...
 if (IS_OK(*status) && IsCompressed(header.client_type))
 {
  Message * const cmsg = msg;
  memcpy(&msglen, cmsg->bytes, 4);
  if (Endian(header.client_type) != Endian(ClientType()))
      FlipBytes(4, (char *)&msglen);
  // new message with uncompressed msglen
  msg = malloc(msglen);
  msg.h = header;
  msg.h.msglen = msglen;
  // decompress bytes
  const unsigned char *ubuffer = unsigned char *)msg->bytes;
  unsigned int usize = msg.h.msglen - sizeof(MsgHdr);
  const unsigned char *cbuffer = unsigned char *)cmsg->bytes + 4; 
  const unsigned int csize = cmsg.h.msglen - sizeof(MsgHdr) - 4;
  *status = uncompress(ubuffer, &usize, cbuffer, csize) == Z_OK;
  free(cmsg);
 }
 if (IS_NOT_OK(*status)) {
  // cleanup
  free(msg);
  return NULL;
 } else if (Endian(header.client_type) != Endian(ClientType()))
   FlipData(msg);
...
```
Furthermore, do we want to keep old style, i.e. `int` and `unsigned int`, around or do we want to switch to the more specific variants, i.e. `int32_t` and `uint32_t`.

zack-vii (approved):
However, urging for code cleanup wherever we touch it so it will gradually become better and easier to understand.

mdstcpip/mdsipshr/GetMdsMsg.c
santorofer: Hi Timo,
Yes, we thought that we could re-write many of the variable names to reflect a bit better what they are, but we were thinking that at this point we should just fix the code only so that it works correctly. Later on we will be go through and improve the names, newer specific data type, etc

mdstcpip/mdsipshr/GetMdsMsg.c
```diff
@@ -121,13 +121,15 @@ Message *GetMdsMsgTOC(Connection *c, int *status, int to_msec)
     {
       Message *m;
       memcpy(&msglen, msg->bytes, 4);
-      dlen = msglen - sizeof(MsgHdr);
```
joshStillerman, line 124: What does moving this change ?

joshStillerman (approved):

mdstcpip/mdsipshr/GetMdsMsg.c
mwinkel-dev: Hi @zack-vii -- Thanks for the explanation and the thorough review.   (Obviously, I jumped to the wrong conclusion based on the similar variable names.)

Note To All:   If there are performance benefits to using compression, then the `mdsip_server` script could be switched back to use `-c 9`.   For details, see PR #2894.

mdstcpip/mdsipshr/GetMdsMsg.c
zack-vii: `msglen` (as in **uncompressed** msglen) is potentially byte swapped, so it should not be used before byte swapping is handled.

santorofer: Yes, as Timo said, there is the chance that there could be a byte swapped before it is used.

PR #2904, opened 2025-05-05
Fix:remove race conditions leading to deadlock

GabrieleManduchi:
When interacting with jScope during automatic update, jScope often freezes. 
This is due to deadlock conditions that occurred when  an update event was received during interaction operations (e.g. zooming). 
This issue has been fixed by removing the useless synchronized option to several methods, and ensuring that critical concurrent operations were carried out by the main Swing thread. 

In addition, a fix has been carried out in the resampling management, mainly due to the usage of float instead of double for the specification of time windows. This led to missing resampling for high frequencies. For this purpose, a new routine (GetXYSignalDoubleLimits) has been added to mdsmisc/ScopeUtilities.c. This routine is called by the new jScope version. The older routine (GetXYSignal) has been retained for compatibility with the older jScope versions.

### DIFF
--- a/java/jscope/src/main/java/mds/jscope/MdsWaveInterface.java
+++ b/java/jscope/src/main/java/mds/jscope/MdsWaveInterface.java
@@ -918,7 +918,7 @@ class MdsWaveInterface extends WaveInterface
 		return 0;
 	}
 
-	public synchronized void refresh() throws Exception
+	public  void refresh() throws Exception
 	{
 		try
 		{
@@ -941,7 +941,7 @@ class MdsWaveInterface extends WaveInterface
 		}
 	}
 
-	public synchronized boolean refreshOnEvent() throws Exception
+	public boolean refreshOnEvent() throws Exception
 	{
 		final long shots[] = GetShots();
 		if (shots == null)
@@ -957,7 +957,7 @@ class MdsWaveInterface extends WaveInterface
 			{
 				if (!signals[sigIdx].supportsStreaming() || !signals[sigIdx].updateSignal())
 					return false;
-				signals[sigIdx].mergeRegions();
+				//signals[sigIdx].mergeRegions();
 			}
 		}
 		return true;
--- a/java/jscope/src/main/java/mds/jscope/MdsWaveInterface.java
+++ b/java/jscope/src/main/java/mds/jscope/MdsWaveInterface.java
@@ -957,7 +957,6 @@ class MdsWaveInterface extends WaveInterface
 			{
 				if (!signals[sigIdx].supportsStreaming() || !signals[sigIdx].updateSignal())
 					return false;
-				//signals[sigIdx].mergeRegions();
 			}
 		}
 		return true;
--- a/java/jscope/src/main/java/mds/provider/MdsDataProvider.java
+++ b/java/jscope/src/main/java/mds/provider/MdsDataProvider.java
@@ -819,7 +819,7 @@ public class MdsDataProvider implements DataProvider
 		@Override
 		public int getNumDimension() throws IOException
 		{
-                   			if (numDimensions != UNKNOWN)
+            if (numDimensions != UNKNOWN)
 				return numDimensions;
 			String expr;
 			if (_jscope_set)
--- a/java/jscope/src/main/java/mds/provider/MdsDataProvider.java
+++ b/java/jscope/src/main/java/mds/provider/MdsDataProvider.java
@@ -271,7 +271,6 @@ public class MdsDataProvider implements DataProvider
 		private float times[] = null;
 		private Dimension dim = null;
 		private int header_size = 0;
-
 		public SimpleFrameData(String in_y, String in_x, float time_min, float time_max) throws Exception
 		{
 			int i;
@@ -635,14 +634,14 @@ public class MdsDataProvider implements DataProvider
 			}
 			else
 			{
-				args.addElement(new Descriptor(null, new float[]
-				{ (float) xmin }));
-				args.addElement(new Descriptor(null, new float[]
-				{ (float) xmax }));
+				args.addElement(new Descriptor(null, new double[]
+				{ (double) xmin }));
+				args.addElement(new Descriptor(null, new double[]
+				{ (double) xmax }));
 			}
 			args.addElement(new Descriptor(null, new int[]
 			{ numPoints }));
-			byte[] retData;
+			byte[] retData = null;
 			int nSamples;
 			try
 			{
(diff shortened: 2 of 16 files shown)


GabrieleManduchi:
forgot to say.....the new jScope version is also compatible with older mdsip servers that do not implement GetXYSignalDoulbeLimits(). Of course in that case the problem remains (it happens only for very high sampling frequencies)

java/jscope/src/main/java/mds/provider/MdsDataProvider.java
```diff
@@ -795,17 +819,19 @@ public void getDataAsync(double lowerBound, double upperBound, int numPoints)
 		@Override
 		public int getNumDimension() throws IOException
 		{
-			if (numDimensions != UNKNOWN)
+                   			if (numDimensions != UNKNOWN)
```
heidthecamp: I think this is a formatting mistake

java/jscope/src/main/java/mds/provider/MdsDataProvider.java
```diff
@@ -651,11 +650,30 @@ public XYData getData(double xmin, double xmax, int numPoints, boolean isLong) t
...
 					retData = GetByteArray(" MdsMisc->GetXYSignalLongTimes:DSC", args);
 				else
-//                      retData = GetByteArray(setTimeContext+" MdsMisc->GetXYSignal:DSC", args);
-					retData = GetByteArray(" MdsMisc->GetXYSignal:DSC", args);
+                                {
+                                    try {
+ 					retData = GetByteArray(" MdsMisc->GetXYSignalDoubleLimits:DSC", args);
+                                    }catch(Exception exc)
+                                    {
+                                        //Try old method in case the mdsip server is not up-to-date
```
heidthecamp, line 660: Good call out! I was confused why we were attempting to call the old method but it makes sense that the server may not have the same be on the same version.

mdsmisc/ScopeUtilities.c
```diff
@@ -1116,8 +1116,9 @@ static mdsdsc_xd_t *getPackedDsc(mdsdsc_xd_t *retXd)
   return retXd;
 }
 
-EXPORT mdsdsc_xd_t *GetXYSignal(char *inY, char *inX, float *inXMin,
```
heidthecamp, line 1119: Can you run a find all and make sure we're not missing any references? It looks like there is a reference to it here: https://github.com/MDSplus/mdsplus/blob/alpha/java/mdsplus-api/src/main/java/mds/MdsMisc.java#L227

java/jscope/src/main/java/mds/wave/WaveformContainer.java
```diff
@@ -1009,8 +1009,8 @@ public void updatePoints(double x, Waveform curr_w)
 				w.UpdatePoint(x, y);
 		}
 	}
-
-	synchronized public void updateWaveforms()
+/*
```
heidthecamp: Should this be removed entirely? If not I think we should add `TODO:` as the first line of the comment for future find-ability.

java/jscope/src/main/java/mds/wave/WaveformContainer.java
```diff
@@ -142,8 +142,8 @@ public void allSameYScale(Waveform curr_w)
 					w.SetYScale(curr_w);
 		}
 	}
-
-	synchronized public void appendUpdateWaveforms()
+/*
```
heidthecamp: Should this be removed entirely? If not I think we should add TODO: as the first line of the comment for future find-ability.

java/jscope/src/main/java/mds/jscope/MdsWaveInterface.java
```diff
@@ -957,7 +957,7 @@ public synchronized boolean refreshOnEvent() throws Exception
 			{
 				if (!signals[sigIdx].supportsStreaming() || !signals[sigIdx].updateSignal())
 					return false;
-				signals[sigIdx].mergeRegions();
```
heidthecamp, line 960: Why are we removing mergeRegions? If we're removing it can we delete the line rather than just comment it?

heidthecamp: It also looks like this is the only place, in our code, that mergeRegions is used.

java/jscope/src/main/java/mds/wave/Grid.java
```diff
@@ -125,8 +125,11 @@ private int BuildGrid(double val[], int mode, double xmax, double ymax, double x
 		curr = (long) (curr_min / step) * step;
 		if (curr > curr_min)
 			curr -= (long) ((curr - curr_min) / step) * step;
-		while (curr >= curr_min)
+                int maxSteps = 1000;
```
heidthecamp: Would you mind leaving a comment on why maxSteps is 1000?

java/jscope/src/main/java/mds/wave/Signal.java
```diff
@@ -53,7 +56,13 @@ void addRegion(RegionDescriptor newReg)
...
 			if (newReg.upperBound < newReg.lowerBound)
 			{
-				System.err.println("INTERNAL ERROR: LOWER BOUND > UPPER BOUND!!!!!");
+				System.err.println("INTERNAL ERROR : LOWER BOUND "+ newReg.lowerBound +" > UPPER BOUND ("+newReg.upperBound+")");
+                                try {
+                                    throw(new Exception());
+                                }catch(Exception exc)
+                                {
+                                    exc.printStackTrace();
+                                }
```
heidthecamp, line 65: Why are we trying to throw an exception? Would this not just hide errors?

heidthecamp (changes requested):
I have a few questions and comments but mostly this looks good to me.

java/jscope/src/main/java/mds/provider/MdsDataProvider.java
GabrieleManduchi: True, fixed

java/jscope/src/main/java/mds/provider/MdsDataProvider.java
GabrieleManduchi: This happens e.g. when updating jScope on a client node, still retaining the old mdsip on the (less dynamic) server node.

mdsmisc/ScopeUtilities.c
GabrieleManduchi: mdsplus-api is a separate library. Honestly, I was unaware of it, and I am not sure what its purpose is. In any case, for consistency, I will add there _miscGetXYSignalDoubleLimits()

java/jscope/src/main/java/mds/wave/WaveformContainer.java
GabrieleManduchi: Yes, removed.

java/jscope/src/main/java/mds/wave/WaveformContainer.java
GabrieleManduchi: removed also this one

java/jscope/src/main/java/mds/jscope/MdsWaveInterface.java
GabrieleManduchi: Right, the way regions (i.e. different sections of the signal at different resolutions) are handled has been changed and that method makes no sense now. Removed the commeneted line.

java/jscope/src/main/java/mds/wave/Signal.java
GabrieleManduchi: This internal error sometimes (very rarely and under particular circumstances) arises due to approximation errors in the data packaging process at the mdsip server side. Normally it does not harm, and previously, there was only the printed message INTERNAL ERROR : LOWER BOUND .... I have added the printStackTrace() call in order to rely on more information when this error occurs and be able to better diagnose it in the future.

java/jscope/src/main/java/mds/wave/Grid.java
GabrieleManduchi: Right, it is ugly. Now the constant MAX_STEPS = 1000 has been defined. (i.e., the maximum number of steps in the tick selection algorithm)

GabrieleManduchi:
retest this please

GabrieleManduchi:
retest this please

GabrieleManduchi:
For some reason jenkins refuses to build, no apparent reason....................

heidthecamp:
Jenkins appears to be down currently. It seems to be all branches since yesterday. We intend on looking into this further today.

heidthecamp:
retest this please

heidthecamp:
@GabrieleManduchi, before I approve do you have any recommendations on how to test this?

GabrieleManduchi:
Being a user interface, it is hard to test in an automated way. In any case, before issuing the pull request, I gave it to our users, and in three weeks, I did not get complaints......

heidthecamp (approved):
Thank you for the updates looks good to me.

PR #2906, opened 2025-05-05
Fix: copy of event data for remote events

### LINKED ISSUE #2905: jScope connection crashes when some wave container has an update_event and this event is set with data
**Affiliation**
SPC-EPFL

**Version(s) Affected**
Client Version: all, Server Version: Alpha>=7.131.0, Stable>=7.131.1

**Platform(s)**
Ubuntu 24.04

**Installation Method(s)**
DEB packages

**Describe the bug**
When a jscope panel has the "update_event" property set and the corresponding event is set with non-empty data, the MDSip connection crashes.

**To Reproduce**
1. Open jScope
2. Set up a connection to some server (Local is enough)
3. In some wave panel set up the "update_event" property to "test_remote_event"
4. Open a separate connection to the same server and set this event with some data (`setevent test_remote_event hello`)
5. A popup opens up in jScope saying the connection has crashed. If not try refreshing, this should trigger the error.

**Expected behavior**
Connection should stay alive and event should be received.

**Screenshots**

**Additional context**
Offending commit is 4e355a7 introduced in #2288.

merlea:
Fix bug in treatment of event data on remote servers in `mdstcpip/mdsipshr/ProcessMessage.c`, the destination of the `memcpy` operation in `_client_event_ast` was wrong and this had 2 side effects:
- for the java client, the destination was the null pointer and thus this caused a SIGSEGV crash (see #2905)
- for other clients, the event data was never copied.

2 tests for these errors were added:
- one for the java client using the `MdsConnection` class
- one for the other clients using the `setevent/wfevent` binaries

Additionally some typo in the treatment of events in JAVA was fixed (`AddEvent->RemoveEvent`).

This closes #2905.


heidthecamp:
Retest this please

merlea:
I updated the tests and they now pass using `deploy/build.sh --os=ubuntu24 --test`. Happy to get some feedback ...

heidthecamp:
Retest this please

heidthecamp:
I haven't had a chance to fully go through this yet but I wanted  to quickly say thank you for the updates and tests. We'll review this as soon as we can.

wfevent/testing/RemoteEventTest.sh
```diff
@@ -0,0 +1,61 @@
...
+
+mds_event_server=localhost:$port wfevent $event_name --timeout=5 > $name.log 2> $name.err &
+pid=$!
+sleep .1
+setevent $event_name
+
+wait $pid
+
+# Check output of actlog against reference
+if diff $name.log ${srcdir}/$name.log_ref; then
```
WhoBrokeTheBuild: Not that we exactly have a standard, but our other answer files are called `.ans`

wfevent/testing/RemoteEventTest.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+stat=0
```
WhoBrokeTheBuild: Please rename stat to something like status or result, I don't want it confused with the `stat` command

wfevent/testing/RemoteEventTest.sh
```diff
@@ -0,0 +1,61 @@
...
+# Check output of actlog against reference
+if diff $name.log ${srcdir}/$name.log_ref; then
+  echo "PASS: $name"
+else
+  echo "FAIL: $name"
+  stat=1
+fi
+
+# Cleanup
+pkill -P $$
```
WhoBrokeTheBuild: Unfortunately, some of the platforms we build for don't ship with pkill
If it's possible to replace with something else, please do, if not we can work on adding it to our test docker images

wfevent/testing/RemoteEventTest.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Start event server
+
+mdsip --multi --protocol=tcp --port=$port --hostfile=$MDSPLUS_DIR/testing/mdsip.hosts > mdsip.log 2> mdsip.err &
+
+### TEST 1: local event no data
+
+name=wfeventRemote
+
+mds_event_server=localhost:$port wfevent $event_name --timeout=5 > $name.log 2> $name.err &
```
WhoBrokeTheBuild: Several of the distributions are failing with:
```
1,3d0
< 
< Error connecting to localhost:8020
< Event test_event timed out.
FAIL: wfeventRemote
PASS: wfeventRemoteData
FAIL RemoteEventTest.sh (exit status: 1)
```

This test might be flaky? We'll run another build and take a look

wfevent/testing/RemoteEventTest.sh
```diff
@@ -0,0 +1,61 @@
...
+fi
+
+### TEST 2: local event with data
+
+name=wfeventRemoteData
+
+mds_event_server=localhost:$port wfevent $event_name --data --timeout=$timeout > $name.log 2> $name.err &
+pid=$!
+sleep .1
+setevent $event_name "hello1world2" # Data from remote event is always 12 byte long
```
WhoBrokeTheBuild: This is currently failing in our windows build with `setevent: command not found`
We'll look into this one as well, but you should be able to replicate it with `./deploy/build.sh --os=windows --test`

testing/ports.csv
```diff
@@ -6,4 +6,6 @@ python/MDSplus/tests/connection-thick, 8013
 python/MDSplus/tests/connection-tcp, 8014
 python/MDSplus/tests/connection-write, 8015
 python/MDSplus/tests/dcl-dispatcher, 8016-8017
-python/MDSplus/tests/dcl-timeout, 8018
\ No newline at end of file
+python/MDSplus/tests/dcl-timeout, 8018
+java/mdsobjects/test/MdsRemoteEventTest, 8019
```
WhoBrokeTheBuild, line 10: I really appreciate you allocating the ports! This could be easily looked over, and it's a big help to keep things organized

WhoBrokeTheBuild (changes requested):
Overall this looks really good! Thank you for all of your work fixing bugs and more importantly, adding tests. I've called out a couple of small things to change, and where the build system is currently failing. We'll work to get these resolved and merged as soon as we can.
Thank you again for all your contributions! :tada:

wfevent/testing/RemoteEventTest.sh
WhoBrokeTheBuild: The address and thread sanitizers are hanging somewhere, it times out when running the tests. You can test those with `--os=test-asan` or `--os=test-tsan`

wfevent/testing/RemoteEventTest.sh
merlea: For Linux I had a similar issue. This is why I modified the test environment in the m4 macro. Maybe there are platform specific settings?

Also the main cause of this is that some binaries are generated in `$MAKEBINDIR` and others such as `wfevent` are generated in their own directory.

wfevent/testing/RemoteEventTest.sh
merlea: Based on the error message, maybe the `mdsip` server takes a while to be established. We can give it more time.

wfevent/testing/RemoteEventTest.sh
merlea: This is just to make sure that background jobs get killed. If `bash` is fully supported, we should be able to find an alternative.

wfevent/testing/RemoteEventTest.sh
merlea: Done

wfevent/testing/RemoteEventTest.sh
merlea: Done

wfevent/testing/RemoteEventTest.sh
merlea: I updated the tests setup to resolve the `command not found` issue for windows. Not sure the tests pass though.

wfevent/testing/RemoteEventTest.sh
merlea: I added some more delays.

For `test-asan`, it seems there are at least some existing memory leaks in the code handling remote events (not this PR). And on my machine many of the other tests also fail...

wfevent/testing/RemoteEventTest.sh
merlea: I now use plain `kill` commands.

merlea:
Following the latest `alpha` update, my new java test breaks in the docker image because it relies on some Swing components that require a valid X11 display... 
Namely this change: https://github.com/MDSplus/mdsplus/pull/2904/files#diff-b1a2e9ca11d4c9eecc827197e6af36e0e7875b8e56126d541eec7e4e94272ba6

heidthecamp:
@GabrieleManduchi, It looks like including swing in `MdsConnection.java` is causing some issues. Can you find a replacement for this?
https://github.com/MDSplus/mdsplus/pull/2904/files#diff-b1a2e9ca11d4c9eecc827197e6af36e0e7875b8e56126d541eec7e4e94272ba6

heidthecamp:
Retest this please

heidthecamp:
Retest this please

merlea:
Here is my current status:
- for all linux options in Jenkinsfile the tests pass
- for windows:
  -  Part of the tests were false positives (marked as PASS when the test had failed) when the reference output file (`*.ans`) was empty. I updated `wfevent` to print a message in cases where the event data was not required to solve this.
  - The remote event tests fail as there is an error in [a select call](https://github.com/MDSplus/mdsplus/blob/11f4e9c8c36f317623e0e665e81b86ee36156fb5/mdsshr/MdsEvents.c#L448). The error message is ironically: "select error: Success". Could it be that Windows and UNIX should be treated differently there?
 - for sanitizers: Running the deploy script for these cause segmentation faults on my machine, even on the `alpha` branch so this is a bit hard to diagnose. However there are at least 2 memory leaks from code that was previously not tested before:
   - The list of server URIs in `mdsshr/MdsEvents.c` are not freed (`receive_servers` and probably `send_servers`).
   - The event info associated to the AST request (also created in `mdsshr/MdsEvents.c`) is not freed.

Related `valgrind` output:
```
==2681185== 15 bytes in 1 blocks are definitely lost in loss record 4 of 17
==2681185==    at 0x4846828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2681185==    by 0x4AE435E: strdup (strdup.c:42)
==2681185==    by 0x49EBF6B: getServerDefinition (MdsEvents.c:339)
==2681185==    by 0x49EC758: initializeRemote (MdsEvents.c:508)
==2681185==    by 0x49ED8DF: RemoteMDSEventAst (MdsEvents.c:870)
==2681185==    by 0x49EDD9E: MDSEventAst (MdsEvents.c:987)
==2681185==    by 0x49ECE25: MDSWfeventTimed (MdsEvents.c:653)
==2681185==    by 0x109567: main (wfevent.c:151)
==2681185==
==2681185== 536 bytes in 1 blocks are definitely lost in loss record 13 of 17
==2681185==    at 0x4846828: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2681185==    by 0x49ECB7A: eventAstRemote (MdsEvents.c:582)
==2681185==    by 0x49ED8F7: RemoteMDSEventAst (MdsEvents.c:871)
==2681185==    by 0x49EDD9E: MDSEventAst (MdsEvents.c:987)
==2681185==    by 0x49ECE25: MDSWfeventTimed (MdsEvents.c:653)
==2681185==    by 0x109567: main (wfevent.c:151)
```

I would appreciate your help in solving both the windows select error issue and the 2 memory leaks. Maybe one possiblity would be to skip the affected tests for windows and the address sanitizers ... Let me know what you think.

heidthecamp:
Retest this please

merlea:
Ok. In this case I thought the java swing mods could be reintroduced as @GabrieleManduchi deemed them useful. What do you think?

WhoBrokeTheBuild (dismissed):

merlea:
OK. Skipped the tests.

heidthecamp:
Retest this please

mwinkel-dev:
I have just done developer builds with and without this pending PR.   That experiment confirms that this PR fixes related issue #2905 as @merlea asserted.   My hunch is that this will PR will also fix issue #2915 at PPPL (which is likely a duplicate of #2905).

The tests that @merlea wrote also work well.   Many thanks for those tests!

wfevent/testing/RemoteEventTest.sh
```diff
@@ -0,0 +1,94 @@
...
+fi
+
+# Start event server
+
+$MDSIP --multi --protocol=tcp --port=$port --hostfile=$MDSPLUS_DIR/testing/mdsip.hosts > mdsip.log 2> mdsip.err &
+pidmdsip=$!
+
+sleep $sleepdur
+
+### TEST 1: local event no data
```
mwinkel-dev: A-OK as is.  What follows is a trivial cosmetic change that can be ignored.

This is the `RemoteEventTest.sh` file, so should the comment for "TEST 1" and "TEST 2" instead say `remote event` instead of `local event`?

merlea: Done

merlea:
OK. Skipped the java tests and fixed the issues flagged by @mwinkel-dev

WhoBrokeTheBuild (approved):

PR #2923, opened 2025-06-18
Fix: mdsconnect's use of $MDS_HOST

WhoBrokeTheBuild:
Fix (or possibly implement) the use of `$MDS_HOST` in `mdsconnect()`.
This will allow users or administrators to define `MDS_HOST` to point at a default thin client host, allowing users to easily type `mdsconnect()` and not worry about where it needs to connect to.

e.g.
```
$ export MDS_HOST=server
$ tdic
TDI> mdsconnect()
Host taken from MDS_HOST [server]
0
TDI> 
```


mwinkel-dev (approved):
These are straightforward changes.   No questions or concerns.

PR #2927, opened 2025-06-30
Release: Update stable to alpha 7.153.3

WhoBrokeTheBuild:


joshStillerman (dismissed):
I trust that the thorough review that you all did when assembling this PR was good and sufficient! 

New Stable!  Yea!

mwinkel-dev (approved):
The only change after Josh's approval was the patch for NumPy 2.3 and greater.

I visually compared that patch with PR #2928 and confirmed that the changes are identical.

Furthermore, in recent days I have installed and used the current "alpha" on various development systems without any problem.   Because this PR promotes that snapshot of "alpha" to the new "stable", all should be OK.

Thus, am approving this PR.

PR #2928, opened 2025-06-30
Fix: Resolve issues with numpy >= 2.3

WhoBrokeTheBuild:
Conditionally replace calls to tostring() with tobytes()
Conditionally replace calls to fromstring() with frombuffer()
Pass the numpy object directly to hashlib, instead of calling tostring()


mwinkel-dev (approved):
These changes are straightforward and match the description, so approving this change.

PR #2930, opened 2025-07-02
Fix: optional port parameter in IDL's mdsconnect()

WhoBrokeTheBuild:
Also fixed in mds$connect()
This now takes the port and concatenates it with the hostname as `host+':'+port` if specified
Previously this set an enviornment variable named `$mdsip` which is not used anywhere


heidthecamp (approved):
Looks good to me.

mwinkel-dev (approved):
This is a good change, so approving it.

Issue #2925, opened 2025-06-30
MATLAB 2024b is losing its connection to the MDSplus server


mwinkel-dev:
**Affiliation**
DIII-D / GA (submitted on behalf of Andrey L.)

**Version(s) Affected**
client: alpha-7-139-59  (6-Nov-2023)
server: ga_atlas-1-0-0 (16-Apr-2024)

**Platform(s)**
client: RHEL8 (specifically 8.8)
server: TBD

**Installation Method(s)**
TBD

**Describe the bug**
Running MATLAB 2024b with the Java bridge on the Omega cluster initially works but then loses its mdsip connection to the MDSplus server.   When this occurs, the following errors are seen.   The only way to fix it is to restart MATLAB.

```
Error during login: recv NULL
Error during login: recv NULL
Error during login: recv NULL
Error during login: recv NULL
Error during login: recv NULL
Error during login: recv NULL
Error during login: recv NULL
D, 1750805219.876:  buffer_free() Connection(id=44, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=43, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=42, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=41, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=40, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=39, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=38, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1750805219.877:  buffer_free() Connection(id=37, state=0x80,
protocol='tcp', info_name='tcp', version=3, user='(null)')
```

**To Reproduce**
TBD

**Expected behavior**
The mdsip connection should remain active.

**Screenshots**
n/a

**Additional context**
n/a

lvovskiy:
[...]

PR #2935, opened 2025-07-10
Feature: Allow using mdsthin for the MATLAB-Python bridge

WhoBrokeTheBuild:
This is a first pass at supporting mdsthin as a backend for connecting MATLAB to MDSplus The mechanism for enabling it is an extension of how you enable python
```matlab
% Use Java
mdsInfo()
% Use Python
mdsInfo(true)
% Use Python with mdsthin
mdsInfo(true, true)
```

Of course, it will throw an error if you attempt to `mdsconnect('local')` while using mdsthin

### DIFF
--- a/matlab/mdsInfo.m
+++ b/matlab/mdsInfo.m
@@ -14,7 +14,10 @@ else
 end
 if nargin > 0
     MDSINFO.usePython = logical(varargin{1});
-    if nargin > 1
+    if ~MDSINFO.usePython
+        MDSINFO.usemdsthin = false;
+    end 
+    if (nargin > 1) && MDSINFO.usePython
         MDSINFO.usemdsthin = logical(varargin{2});
     end
 end
@@ -47,12 +50,14 @@ if ~all(ok)
         disp(strcat(' Java error: ', err{2}.message))
         disp(strcat(' Python error: ', err{1}.message))
         info = [];
-        MDSINFO = info;
+        MDSINFO = info; 
     elseif ~ok(2)
         disp('Unable to connect to MDSplus using java bridge, using python bridge instead')
         disp(strcat(' Java error: ', err{2}.message))
     else % ~ok(1)
-        disp('Unable to connect to MDSplus using python bridge')
+        disp('Unable to connect to MDSplus using python bridge, using java bridge instead')
         disp(strcat(' Python error: ', err{1}.message))
     end
 end
+
+info = MDSINFO; % success, so return updated structure
--- a/matlab/mdsInfo.m
+++ b/matlab/mdsInfo.m
@@ -8,11 +8,15 @@ else
     info.connection = [];
     info.connectedHost = '';
     info.usePython = false;
+    info.usemdsthin = false;
     info.isPythonConnection = false;
     MDSINFO = info;
 end
 if nargin > 0
     MDSINFO.usePython = logical(varargin{1});
+    if nargin > 1
+        MDSINFO.usemdsthin = logical(varargin{2});
+    end
 end
 err = {true, true};
 if MDSINFO.usePython
@@ -48,7 +52,7 @@ if ~all(ok)
         disp('Unable to connect to MDSplus using java bridge, using python bridge instead')
         disp(strcat(' Java error: ', err{2}.message))
     else % ~ok(1)
-        disp('Unable to connect to MDSplus using python bridge, using python bridge instead')
+        disp('Unable to connect to MDSplus using python bridge')
         disp(strcat(' Python error: ', err{1}.message))
     end
 end
--- a/matlab/mdsconnect.m
+++ b/matlab/mdsconnect.m
@@ -21,6 +21,7 @@ function [ status ] = mdsconnect( host )
                 else
                     MDSINFO.connection.mdsdisconnect();
                 end
+            end
 
             if MDSINFO.usemdsthin
                 error("Using mdsconnect('local') with mdsthin is not supported")
--- a/matlab/mdsconnect.m
+++ b/matlab/mdsconnect.m
@@ -21,7 +21,11 @@ function [ status ] = mdsconnect( host )
                 else
                     MDSINFO.connection.mdsdisconnect();
                 end
+
+            if MDSINFO.usemdsthin
+                error("Using mdsconnect('local') with mdsthin is not supported")
             end
+            
             MDSINFO.isConnected = false;
             MDSINFO.connection = [];
             MDSINFO.connectedHost = host;
--- a/matlab/mdsdisconnect.m
+++ b/matlab/mdsdisconnect.m
@@ -4,10 +4,10 @@ function [ status ] = mdsdisconnect( )
 %      described routines to their local behaviors
     global MDSINFO
     if MDSINFO.usemdsthin
-        MDSINFO.connection = []
+        MDSINFO.connection = [];
         MDSINFO.connectedHost = '';
         MDSINFO.isConnected = false;
-        status = 1
+        status = 1;
     else
         status = mdsconnect('local');
     end
--- a/matlab/mdsdisconnect.m
+++ b/matlab/mdsdisconnect.m
@@ -2,6 +2,14 @@ function [ status ] = mdsdisconnect( )
 % MDSDISCONNECT  disconnect from  a remote mdsplus data server.   
 %      mdsdisconnect will destroy this connection, reverting the above
 %      described routines to their local behaviors
-    status = mdsconnect('local');
+    global MDSINFO
+    if MDSINFO.usemdsthin
+        MDSINFO.connection = []
+        MDSINFO.connectedHost = '';
+        MDSINFO.isConnected = false;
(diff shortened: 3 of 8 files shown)


matlab/private/pythonActivate.m
```diff
@@ -5,7 +5,11 @@
     err = cache;
 else
     try
-        MDSINFO.ispy2 = logical(py.MDSplus.version.ispy2);
+        if MDSINFO.usemdsthin
+            MDSINFO.ispy2 = false
+        else
+            MDSINFO.ispy2 = logical(py.MDSplus.version.ispy2);
```
WhoBrokeTheBuild, line 11: We should consider removing this check altogether, the last matlab version that supported python 2 was R2022b

mwinkel-dev: I concur.

I hope there are no customers still using MATLAB R2022b.   Disabling this check would be a way to find out (grin).   Plus, it would be easy to undo if the user base clamors for Python2 and ancient MATLAB releases.

mwinkel-dev:
I have reviewed the changes and they are all straightforward.

However, I am holding off on the approval until I've installed this commit on my test systems and used it for a few days.

matlab/mdsInfo.m
```diff
@@ -48,7 +52,7 @@
         disp('Unable to connect to MDSplus using java bridge, using python bridge instead')
         disp(strcat(' Java error: ', err{2}.message))
     else % ~ok(1)
-        disp('Unable to connect to MDSplus using python bridge, using python bridge instead')
```
mwinkel-dev, line 51: This message should end with `using java bridge instead`.

During the investigation of #2925, it was noticed that if the specified bridge is not available, then the code tries the alternate bridge.   It is important to notify the user that if the Python bridge fails to launch, that it is then running the Java bridge.

matlab/mdsInfo.m
```diff
@@ -8,11 +8,15 @@
     info.connection = [];
     info.connectedHost = '';
     info.usePython = false;
+    info.usemdsthin = false;
     info.isPythonConnection = false;
     MDSINFO = info;
 end
 if nargin > 0
     MDSINFO.usePython = logical(varargin{1});
+    if nargin > 1
```
mwinkel-dev: To guard against bogus input that attempts to use `mdsthin` with the Java bridge . . .

 ```
mdsInfo(false, true)
```

. . . the `if` statement should probably be as follows.

```
if (nargin > 1) && MDSINFO.usePython
```

**Comment by mwinkel-dev** [comment]
Although it is correct that `mdsInfo(true)` does enable the Python bridge, the documentation instead recommends that customers use the wrapper function, `mdsUsePython()`.

Perhaps the `mdsUsePython()` function should also accept a second parameter to specify use of `mdsthin` with Python.    Suggested changes for the `mdsUsePython.m` file follow.

```
   if nargin == 0
       mdsInfo(true, false);
   elseif nargin == 1
       mdsInfo(varargin{1});
   else
       mdsInfo(varargin{1}, varargin{2});
   end
```

**Comment by mwinkel-dev** [comment]
I have confirmed that the `mdsthin-bridge` works fine on a M2 MacBook with MATLAB R2024b.   All it requires is the `$MDSPLUS_DIR/matlab` directory.

Here are the configuration steps . . . 

1. Install `mdsthin` via `pip`.
2. The `$MDSPLUS_DIR/matlab` directory was copied to a `thin_bridge` directory. 
3. MATLAB's search path was then modified so it no longer referenced `$MDSPLUS_DIR/matlab`, but instead referenced the `thin_bridge` directory.
4. Logged out and logged back in to get a clean shell environment (i.e., no references to $MDSPLUS_DIR or related environment variables).
5. Started MATLAB.   (Didn't need graphical output, so used `matlab  -nodisplay` because it launches faster.)
6. At the MATLAB `>>` prompt typed the following statements

```
mdsInfo(true, true)  % specifies the "mdsthin" bridge
mdsconnect("<server>")
mdsopen("cmod", 1090909009)
mdsvalue("\ip")
mdsdisconnect
```

Next step is to install the `mdsthin-bridge` on Arsene's M2 MacBook.

**Comment by mwinkel-dev** [comment]
Note that `mdsInfo()` does not mean "use Java"; it merely returns the current `MDSINFO` structure.   Yes, that structure is initialized to use the Java-bridge.   However, if the user has switched to the Python-bridge and then wants to switch back to the Java-bridge, `mdsInfo()` won't do the job.   They must instead explicitly use `mdsInfo(false)` or `mdsUsePython(false)`.

File: matlab/mdsInfo.m [hunk 4]
```diff
@@ -48,7 +52,7 @@
         disp('Unable to connect to MDSplus using java bridge, using python bridge instead')
         disp(strcat(' Java error: ', err{2}.message))
     else % ~ok(1)
-        disp('Unable to connect to MDSplus using python bridge, using python bridge instead')
+        disp('Unable to connect to MDSplus using python bridge')
         disp(strcat(' Python error: ', err{1}.message))
     end
 end
```
mwinkel-dev, line 61: If control flow makes it beyond the final `end`, then a bridge has been successfully opened.   In which case, the function should return the current state of the `MDSINFO` structure.

```
info = MDSINFO;
```

mwinkel-dev (dismissed):
Confirmed this PR has all the changes made in the prototype installed on Arsene's MacBook.   (His computer is Apple Silicon and has been running the mdsthin-bridge for a month or so with no problems.)  This PR also has a useful enhancement regarding the `mdsdiscconect.m` script (so it doesn't throw an error when disconnecting from the mdsthin-bridge).   Approving this PR.

mwinkel-dev:
The testing done two hours ago used the wrong version of the `mdsdisconnect.m` file (and probably other files too).   After updating to the current version of the files in this PR, other errors were also found.   A full retest will be done later today and the results posted to this PR.

matlab/mdsconnect.m
```diff
@@ -21,7 +21,11 @@
                 else
                     MDSINFO.connection.mdsdisconnect();
                 end
```
mwinkel-dev, line 23: There should be another 'end'  here to close out the 'if MDSINFO.isConnected' block.   Without this additional 'end', none of the three bridges will connect to a server.

matlab/mdsdisconnect.m
```diff
@@ -2,6 +2,14 @@
 % MDSDISCONNECT  disconnect from  a remote mdsplus data server.   
 %      mdsdisconnect will destroy this connection, reverting the above
 %      described routines to their local behaviors
-    status = mdsconnect('local');
+    global MDSINFO
+    if MDSINFO.usemdsthin
+        MDSINFO.connection = []
```
mwinkel-dev: Best to have a trailing ";" on the "... connection = []" statement.

matlab/mdsdisconnect.m
```diff
@@ -2,6 +2,14 @@
...
+        MDSINFO.connectedHost = '';
+        MDSINFO.isConnected = false;
+        status = 1
```
mwinkel-dev: The "status = 1" should also have a trailing ";" to suppress output to the screen.   Having the ";" is also consistent with the "else" clause below.

mwinkel-dev:
Thorough testing with the correct *.m files revealed some problems (noted in previous posts).

There is also a design question.   If using the mdsthin-bridge and then issuing a `mdsUsePython()`, should it remain in mdsthin mode or switch to the standard Python mode?   Current behavior is shown below.

```
>> mdsUsePython(true, true)
>> mdsInfo()

ans = 

  struct with fields:

           isConnected: 0
            connection: []
         connectedHost: 'local'
             usePython: 1
            usemdsthin: 1
    isPythonConnection: 0
                 ispy2: 0

>> mdsUsePython(true) 
>> mdsInfo()

ans = 

  struct with fields:

           isConnected: 0
            connection: []
         connectedHost: 'local'
             usePython: 1
            usemdsthin: 1
    isPythonConnection: 0
                 ispy2: 0

>> 
```

Unrelated to this PR, a pre-existing problem was also found. If issue `mdsclose()` when a "local" tree is open via Python, it returns error code 265388200 = FD180A8 = TreeNOT_OPEN. However, when using Java, `mdsclose()` correctly returns the TreeSUCCESS status.

mwinkel-dev (dismissed):
The third commit to this PR fixes the problems encountered with the previous commit.   The manual tests all pass, thus approving this PR.

The new documentation should explain how to install the mdsthin-bridge for MATLAB.   (Some users don't want a full MDSplus installation; they just want the simplest way to get MATLAB to read data from an MDSplus archive server.)  The easiest installation is probably to do the following:

1. install mdsthin,
2. download the appropriate *.tgz file from GitHub, and
3. extract just the `mdsplus/matlab` directory.

mwinkel-dev (approved):
A separate *.tgz file for the mdsthin-bridge is an excellent idea!   Thanks for changing the Jenkins file accordingly.   It all looks good, so approving this PR.

heidthecamp (approved):

Issue #2926, opened 2025-06-30
Have to restart MATLAB every 10 minutes because it is dropping the connection to the MDSplus server


mwinkel-dev:
**Affiliation**
DIII-D / GA (submitted on behalf of Francesca T.)

**Version(s) Affected**
TBD

**Platform(s)**
TBD

**Installation Method(s)**
TBD

**Describe the bug**
Have to restart MATLAB 2023b every 10 minutes because of dropped connections to the MDSplus server.   This problem occurs on the Omega-a cluster; it was never seen on the Iris cluster.

The problem arises when using the `getalldat()` and `getmds()` functions.   It works for one iteration of a loop and then fails on the next iteration, even though accessing the same shots.   When the connection is dropped, the following error is reported.   No data is retrieved and no other errors are reported.

```
Error during login: recv NULL
```

**To Reproduce**
Run the sample script supplied by Francesca.

**Expected behavior**
The mdsip connection should remain active regardless of how many iterations there are of the loop.

**Screenshots**
n/a

**Additional context**
This is surely a duplicate of Issue #2925.   However, it is initially being tracked as a separate issue because it involves a different version of MATLAB and a different researcher (thus likely a slightly different workflow).   When it is proven that this issue is a duplicate, it will be closed and folded into #2925.

mwinkel-dev:
Hi Francesca -- Which version of MATLAB do you use when you run the example script on the Iris cluster?

leighchery:
(the notifications work!)
At the moment I am using 2023b, the one that doesn’t have the bug on the “updater”, so I have to be on omega-a where I have my 2023b license. I could also try it on the standard version on omega-b if that is useful, but that will have the other issue of the constant messages about not being able to run the updater, which may muddy the waters.

leighchery:
One more detail that may or may not be relevant: on iris I had to use this commend

global MDSplus_persistent
MDSplus_persistent = 1;

to go through many instances of getmds data. I have tried to use this on omega, I have also tried commenting it out, and the issue reported in this thread does not seem affected.

mwinkel-dev:
Hi @leighchery -- Thanks for the additional information.

The note about `MDSplus_persistent` is a useful clue.   I just scanned the entire MDSplus source code and it does not contain that environment variable.   Thus, `MDSplus_persistent` is surely controlling a DIII-D customization.   (I will follow up with my counterparts at DIII-D to confirm that is the case.)

[...]

PR #2897, opened 2025-03-29
Fix: correct miscellaneous typos, add comments, disable a debug

mwinkel-dev:
Miscellaneous corrections to typos, add some comments, and comment out an unneeded `DEBUG` define.

PR #2936, opened 2025-07-16
Fix: mdsdisconnect for MATLAB java bridge no longer uses Java finalize()

mwinkel-dev:
This is a fix for Issue #2925.

To enable MATLAB to access MDSplus, there are two bridges: Python and Java.   The Java bridge was leaving dangling "connection" objects in MATLAB and dangling mdsip connections on the MDSplus archive server.   This is because the `mdsdisconnect` provided to MATLAB was not immediately terminating the mdsip connection to the server.   It was instead releasing a reference to the "connection" object, which was in turn relying on Java's garbage collector to run the `finalize()` method to destroy the connection.   However, the garbage collector runs infrequently and cannot be controlled by the application program.   Which means that MATLAB accumulates dangling "connection" objects and the MDSplus archive server accumulates dangling mdsip connections.   This was causing MATLAB and/or the archive server to eventually choke.   The mdsip connections would only be destroyed when MATLAB's garbage collector ran, MATLAB crashed, or the user shutdown MATLAB.

Note that the Java `finalize()` method was deprecated in Java 9 (in part because of this issue with the garbage collector).   Oracle's most recent advice is to remove the `finalize()` method from classes (see the entry for java.lang.Object.finalize in the following document).
https://docs.oracle.com/en/java/javase/22/docs/api/deprecated-list.html

This PR only fixes the MATLAB Java bridge `mdsdisconnect` issue.   Future PRs should remove `finalize()` from all other Java classes in MDSplus.

This PR was tested manually by using two virtual machines: an archive server and a client.   The client ran a MATLAB program that had a loop that iterated 1,000 times, with each iteration doing a pair of `mdsconnect` and `mdsdisconnect` calls.   When the MATLAB test finished, the archive VM was checked to see how many mdsip connections were present.   Prior to implementing this fix, the test showed that there was a dangling mdsip connection for every iteration of the loop.   With the fix, all mdsip connections were closed.

### DIFF
--- a/java/mdsobjects/src/main/java/MDSplus/Connection.java
+++ b/java/mdsobjects/src/main/java/MDSplus/Connection.java
@@ -64,8 +64,7 @@ public class Connection
 		isConnected = true;
 	}
 
-	@Override
-	protected void finalize()
+	public void mdsdisconnect()
 	{
 		if (sockId >= 0)
 			disconnectFromMds(sockId);
--- a/java/mdsobjects/src/main/java/MDSplus/Connection.java
+++ b/java/mdsobjects/src/main/java/MDSplus/Connection.java
@@ -71,6 +71,13 @@ public class Connection
 		isConnected = false;
 	}
 
+	// Deprecated in Java 9 to Java 22; will be removed in future Java.
+	@Override
+	protected void finalize()
+	{
+		mdsdisconnect();
+	}
+
 	public void openTree(java.lang.String name, int shot) throws MdsException
 	{
 		openTree(sockId, name, shot);
--- a/matlab/mdsconnect.m
+++ b/matlab/mdsconnect.m
@@ -15,6 +15,9 @@ function [ status ] = mdsconnect( host )
     else
         if strcmpi(host, 'LOCAL') == 1
             MDSINFO.isConnected = false;
+            if ~MDSINFO.isPythonConnection
+                MDSINFO.connection.mdsdisconnect();
+            end
             MDSINFO.connection = [];
             MDSINFO.connectedHost = host;
             status = 1;


mwinkel-dev:
The fix for Issue #2925 should also include PR #2897 to turn off a debug statement in the `mdstcpip/mdsipshr/MdsIpThreadStatic.c` file.   That statement was generating the many messages that the users reported with Issue #2925.

mwinkel-dev:
Note that this fix merely immediately disconnects from the archive server, but it does not destroy the MATLAB "connection" objects.   They remain dangling objects that only get cleared when the Java garbage collector eventually runs or the user shuts down MATLAB.

Nonetheless, this fix made a dramatic difference on the test systems.   Prior to the fix, the test virtual machines could only iterate 64 times before MATLAB and/or the archive server choked.

With the fix, it easily handled 1,000 iterations of the `mdsconnect` / `mdsdisconnect` pairs.

joshStillerman (dismissed):
this looks great!

mwinkel-dev:
This PR has been tested on GA's Omega cluster and works OK.

Here are the details . . . 

This simple MATLAB script reproduced the problem reported by the GA researchers in issues #2925 and #2926.   If the loop is changed to run many iterations, and run with MDSplus alpha-7-139-59, then each iteration leaves a dangling object in MATLAB and a dangling connection on the MDSplus archive server.   As those dangling objects / connections accumulate, eventually MATLAB and/or the server choke, thereby forcing the user to exit MATLAB and start over.

```
% Name:   bug_2925.m
% Purpose:   This tests the PR #2936 fix of Issue #2925.

% Issue 2925 was caused by relying on the Java garbage collector
% to terminate connections and delete connection objects.  However,
% there is no control over when the garbage collector runs, thus
% dangling objects / connections are created -- especially when
% the mdsdisconnect is in a loop. 

% On development system, this program would drop the mdsip connection
% if the loop ran more than 64 iterations, whereupon it also displayed
% the various error messages reported in Issue #2925.


% Cannot use the Python-bridge because Omega's MATLAB requires
% installing Python 3.9 (or newer), but Omega only has Python 3.7.
% mdsUsePython(true)

for i = 1:5
   status = mdsconnect('atlas');
   if status == -1
      break
   end
   mdsopen('RF', 203131);
   d = mdsvalue('LHCD.LH_COORDS');
   mdsclose();
   mdsdisconnect;
   disp(i)
end
```


**STATUS QUO:**

This is the MDSplus version GA has installed on Omega.   It is from the "alpha" branch.

```
$ mdstcl
TCL> show version


MDSplus version: 7.139.59
----------------------
  Release:  HEAD_release_7.139.59
  Browse:   https://github.com/MDSplus/mdsplus/tree/HEAD_release_7.139.59
  Download: https://github.com/MDSplus/mdsplus/archive/HEAD_release_7.139.59.tar.gz
```

This is the output of the MATLAB script when run in MATLAB 2024b.   Note that the `quit` causes MATLAB's garbage collector to run, which destroys the dangling objects, thereby generating the many messages plus also terminating the associated mdsip connections.

```
$ matlab -nodisplay

To get started, type doc.
For product information, visit www.mathworks.com.
 
>> bug_2925
     1

     2

     3

     4

     5

>> quit
D, 1752786698.807:  buffer_free()                 Connection(id=4, state=0x80, protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1752786698.807:  buffer_free()                 Connection(id=3, state=0x80, protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1752786698.807:  buffer_free()                 Connection(id=2, state=0x80, protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1752786698.807:  buffer_free()                 Connection(id=1, state=0x80, protocol='tcp', info_name='tcp', version=3, user='(null)')
D, 1752786698.807:  buffer_free()                 Connection(id=0, state=0x80, protocol='tcp', info_name='tcp', version=3, user='(null)')
$ 
```



**PR 2936 FIX:**


A "developer" build of MDSplus (current "alpha + PR 2936") was put in a tar file and installed in my home directory on Omega.   (This installation has no impact on the cluster's main MDSplus.)   A shell script was run to set environment variables in my session to point to the MDSplus dev build.   Also a `javaclasspath.txt` file was used to add the dev build's Java classes to the beginning of MATLAB's static javaclasspath list.

The `Date:` field shows that this is indeed the dev build of MDSplus that is being used.   (Ignore the version number; it is bogus on this dev build.)

```
$ mdstcl
TCL> show version


MDSplus version: 7.139.17
----------------------
  Release:  alpha_release-7-139-17
  Date:     Wed Jul 16 19:02:26 UTC 2025
  Browse:   https://github.com/MDSplus/mdsplus/tree/alpha_release-7-139-17
  Download: https://github.com/MDSplus/mdsplus/releases/tag/alpha_release-7-139-17


TCL> 
```

This shows the output of the MATLAB program when running with the dev build of MDSplus.   Although not shown here, the loop was set to 200 iterations and worked fine.

```
$ matlab -nodisplay

To get started, type doc.
For product information, visit www.mathworks.com.
 
>> bug_2925
     1

     2

     3

     4

     5

>> exit
$ 
```

**NOTES:**

An attempt was made to run GA's `getalldat.m` MATLAB script with the dev build of MDSplus, but it involved too much fiddling with environment variables. 

Upgrading MDSplus on a cluster that has many users can be a big task.   GA might find it simpler to modify its MATLAB scripts -- `getmds.m`, `getalldat.m`, etcetera -- to skip the `mdsdisconnect` call if GA's `MDSplus_persistent` flag is set.   Doing so would re-use the existing connection, and eliminate dangling connection objects in MATLAB and dangling mdsip connections on the MDSplus server.

java/mdsobjects/src/main/java/MDSplus/Connection.java
```diff
@@ -64,8 +64,7 @@ public void initialize(java.lang.String mdsipAddr) throws MdsException
 		isConnected = true;
 	}
 
-	@Override
-	protected void finalize()
```
WhoBrokeTheBuild, line 68: Overall this looks good and the logic behind it is sound, so LGTM
My one question, could we leave `finalize()` and just have it call `mdsdisconnect()` that would leave the existing functionality for all consumers of this API (which is not just MATLAB) while making sure we can close connections when needed?

WhoBrokeTheBuild (dismissed):

java/mdsobjects/src/main/java/MDSplus/Connection.java
mwinkel-dev: Hi @WhoBrokeTheBuild,

Yes, we can do as you suggest.   I will make that change.   Note though that a future Java release (a few years from now?) will remove the `finalizer`, at which time we will likely need to remove the `finalize()` method from the MDSplus Java classes.

Now for the details . . . 

I had considered that option too and vacillated on whether to retain `finalize()` or not.   I ultimately decided to remove `finalize()` simply because of the recent guidance from Oracle.   That guidance indicates that they hope to remove the `finalizer` feature in an upcoming release of Java.

However present JVMs still call the `finalize()` method of the ultimate base class when the garbage collector runs (i.e., typically as part of the shutdown process).   The MDSplus Java classes are subclasses of Java's base class, so the connection objects are still dangling until the garbage collector runs.   

Here are the pros / cons of retaining the `finalize()` method at this time.

- Pro: It ensures that there are **_eventually_** no dangling mdsip connections on the server, if an error occurs that disrupts normal program flow thus bypassing the `mdsdisconnect` call.   (Note though that the garbage collector runs infrequently, so dangling objects and connections can still accumulate for long periods of time -- such as until the program crashes or is shut down.)

- Con: Regardless of whether the MDSplus connection object has a `finalize()` method or not, when the object is destroyed, the server surely detects that the TCP connection has been broken and gets rid of the associated dead mdsip process.   Furthermore, retaining the `finalize()` in the MDSplus source code is a trivial form of technical debt that will eventually have to be deleted in some future year.

Here is my recommendation for improving this PR . . . 

1. Make the change you suggested, and
2. Add a comment to the retained `finalize()` method that it has been deprecated since Java 9 and will be removed in some future Java release.

WhoBrokeTheBuild: Gotcha, now I understand the full context. I think that's a good solution, and then we can remove all the `finalize()` in one commit with the appropriate context.

Thanks!

mwinkel-dev:
Restored the `finalize()` method but flagged it as deprecated (i.e., will be removed in a future Java).

Note that when a Java process ends, the garbage collector is usually not run (i.e., instead the operating system cleans up all the system resources for the process).   The Java garbage collector thus typically runs when the process is under memory pressure.   The Java `System.gc()` method can be used to **_suggest_** that the JVM do garbage collection, but there is no guarantee that it will do so.

Thus, it is difficult to create a test case that proves that the MDSpluss Connection object's `finalize()` method has actually been called.

Nonetheless, this revised version of this PR works fine in manual testing.   It compiles OK, immediately disconnects when the `mdsdisconnect()` method is called, and does not leave dangling mdsip processes on the server.

WhoBrokeTheBuild (approved):

PR #2957, opened 2025-10-06
Build: Allow adjusting the number of `-j` threads per agent

WhoBrokeTheBuild:
This allows for the `$THREADS` environment variable to be set on a jenkins agent node to define the value passed to `-j`, with a default of 8.
This way, the number of executors and the number of threads per executor can all be controlled from within Jenkins, without requiring a commit.


mwinkel-dev (approved):
Straightforward change that matches the description.   Thus approving this PR.

This change is useful because the build system uses multiple computers, each with different hardware specifications, thus some can support more threads than others.

Issue #2605, opened 2023-08-30
Potential memory leak in mdstcpip


zack-vii:
**Affiliation**
IPP - HGW Germany

**Version(s) Affected**
current alpha 8f6754d16d8fe3d157181b62052c3999eebf3c92

**Platform**
jenkins slave for address-sanitizer

**Describe the bug**
Potential Memory leak

**To Reproduce**
1. run tditest unit test test-mdsip under address-sanitizer
2. you may sometimes detect:
```=================================================================
==9713==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 52 byte(s) in 1 object(s) allocated from:
    #0 0x7f586fbf6667 in __interceptor_malloc (/lib64/libasan.so.6+0xb0667)
    #1 0x7f586c1bdaa6 in _send_response /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:506
    #2 0x7f586c1bdaa6 in send_response /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:559
    #3 0x7f586c1c3d81 in return_status /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:571
    #4 0x7f586c1c3d81 in standard_command /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:1036
    #5 0x7f586c1c3d81 in ProcessMessage /git/mdsplus/mdstcpip/mdsipshr/ProcessMessage.c:1332
    #6 0x7f586c1cdbae in ConnectionDoMessage /git/mdsplus/mdstcpip/mdsipshr/DoMessage.c:46
    #7 0x7f586c1d18bb in io_listen /git/mdsplus/mdstcpip/io_routines/IoRoutinesThread.c:66
    #8 0x7f586f3dc431 in start_thread (/lib64/libpthread.so.0+0x9431)

Direct leak of 48 byte(s) in 1 object(s) allocated from:
    #0 0x7f586fbf6667 in __interceptor_malloc (/lib64/libasan.so.6+0xb0667)
    #1 0x7f586c1b955f in GetMdsMsgTOC /git/mdsplus/mdstcpip/mdsipshr/GetMdsMsg.c:115
    #2 0x7f586c1cdad8 in ConnectionDoMessage /git/mdsplus/mdstcpip/mdsipshr/DoMessage.c:45
[...]

PR #2959, opened 2025-10-08
Build: Temporarily disable the address sanitizer

WhoBrokeTheBuild:
The updated address sanitizer is (probably correctly) upset about issues that we've seen in the leak sanitizer for a while, which are being tracked in Issue #2605

Unfortunately, there's no way to blacklist functions in asan, so the only way to have stable builds in the meantime is to disable asan altogether


mwinkel-dev (approved):
As was described in the PR, we can deal with the address sanitizer later.   This PR is thus approved.

heidthecamp (approved):

PR #2960, opened 2025-10-09
Build: Publish to the staging directory

WhoBrokeTheBuild:
The nightly rsync is currently disabled, but will be enabled once we verify the packages in the staging directory

Also remove "cmake" from the `confi/distributions` file generated by `debian_publish.py`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -300,7 +300,7 @@ pipeline {
                             unstash "packages-${os}"
                             unstash "dist-${os}"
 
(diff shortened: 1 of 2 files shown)


mwinkel-dev (dismissed):
Approving this simple change.

mwinkel-dev (approved):
The CMake branch is now merged to alpha, thus can be removed from the debian_publish.py file.   Approving the second commit to this PR.

PR #2963, opened 2025-10-14
Build: Fix macosx publishing

WhoBrokeTheBuild:
Add a mkdir to `build.py` to make `dist/homebrew` or `dist/macports`
This will cause `shutil.copy2` to properly place the tgz in the directory
Add flavor as a subdirectory to `macosx_publish.py` so that alpha and stable packages are separated


mwinkel-dev (approved):
Straightforward changes, thus this PR is approved.

santorofer (approved):
It looks all good.

PR #2950, opened 2025-09-07
(#2937) MdsConnection implements AutoCloseable

leewin12:
This PR refactors MdsConnection to implement AutoCloseable as a preliminary step towards replacing [deprecated finalize()](https://github.com/MDSplus/mdsplus/issues/2937) across the codebase.
This change enables deterministic, cascading resource cleanup.

It also introduces the Mockito framework for easier and more succinct testing.
There are some TODO comments for discussion.

As I am new to this project, I am eager to learn and would appreciate any feedback or guidance on my approach.

PR #2966, opened 2025-10-15
Build: Fix Java test $CLASSPATH

WhoBrokeTheBuild:
Unset `CLASSPATH` for all tests to ensure Java tests use the built JAR files, not the system ones

Swap `path_list_append` to `path_list_prepend`
Note: This would also solve the above problem but I don't want to rely solely on the order of the `$CLASSPATH` environment variable

This issue was found while testing a rebase on #2950


mwinkel-dev (approved):
Prepending the built *.jar file is an excellent change, so approving this PR.

PR #2971, opened 2025-10-17
Feature: Actually include HDF5 tools/libraries in debian builds

WhoBrokeTheBuild:
Currently the debian hdf5-bin packages contain nothing, and the hdf5 packages contain TDI code that won't work without the libraries

This should be merged after the latest commit to the docker repo is built successfully, which adds libhdf5-dev to all debian images


heidthecamp (dismissed):

mwinkel-dev (approved):
The changes to the compiler options are consistent across platforms.   The same is true of the packaging changes.   So, approving this PR.

PR #2974, opened 2025-10-21
Fix: fix brother of subtree nodes.

### LINKED ISSUE #2972: Issues with release build and brother_of (and similar) functions
**Affiliation**
MIT PSFC

**Version(s) Affected**
Alpha 7.154.0

**Platform(s)**
Ubuntu 24.04, others

**Installation Method(s)**
mdsplus.org repository

**Describe the bug**
There are issues with brother_of and other similar functions being optimized to the point of breaking.

This might be affecting nodes that come after subtree nodes, or specifically subtree nodes that follow other subtree nodes, it's difficult to say.

**To Reproduce**
0. Run a release build, from the packages or `-DCMAKE_BUILD_TYPE=Release` or `RelWithDebInfo`
1. `tdic`
2. `treeopen('cmod', 1090909009)`
3. `getnci(\top, 'children_nids')`

**Expected behavior**
Expect to see a long list of nodes, see only 2

joshStillerman:
The optimizer is deciding that it does not need to fill in the brother nid of nodes that are subtree attachment points. Making the variable that holds it volatile forces the optimizer to do the right thing.

Fixes: https://github.com/MDSplus/mdsplus/issues/2972

### DIFF
--- a/treeshr/TreeOpen.c
+++ b/treeshr/TreeOpen.c
@@ -1187,7 +1187,7 @@ static void SubtreeNodeConnect(PINO_DATABASE *dblist, NODE *parent,
   node_to_nid(dblist, parent_of(dblist, parent), &parent_nid);
   parent->child = *(int *)&child_nid;
   if (brother) {
-    node_to_nid(dblist, brother_of(dblist, parent), (NID *)&brother_nid);
+    node_to_nid(dblist, brother_of(dblist, parent), &brother_nid);
   }
   subtreetop->brother = *(int *)&brother_nid;
   subtreetop->parent = *(int *)&parent_nid;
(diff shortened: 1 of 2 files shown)


WhoBrokeTheBuild (approved):
What a wild ride to find this, good work and let's hope there aren't any other bugs like this around

treeshr/TreeOpen.c
```diff
@@ -1174,17 +1174,23 @@ static int GetVmForTree(TREE_INFO *info, int nomap)
 static void SubtreeNodeConnect(PINO_DATABASE *dblist, NODE *parent,
                                NODE *subtreetop)
 {
-  NID child_nid, parent_nid, brother_nid = {0, 0};
+  NID child_nid, parent_nid = {0, 0};
```
WhoBrokeTheBuild, line 1177: One note here, is that the old line reads like all things are getting assigned a default, but only `brother_nid` was. So this is now changing it to give a default to `parent_nid` as well. This shouldn't be an issue, but is a change.

heidthecamp (approved):

PR #2976, opened 2025-10-22
Build: Rework Jenkinsfile handling of new_tag to fix GitHub releases

WhoBrokeTheBuild:
Instead of tagging in the head node, which confuses `create_github_release.py`, tag in each distribution

Otherwise `create_github_release.py` always sees the tag as pre-existing, and never makes a release.
Removing that check from `create_github_release.py` would cause it to attempt to make a release even if it already exists, which causes worse errors.


heidthecamp (approved):
I wish it was easier to share variables across sections. That said this works.

Issue #2965, opened 2025-10-15
Some Java tests fail when users build from source on Apple Silicon, likely is a configuration issue


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
Local builds of alpha-7-154-0 (and newer) on Apple Silicon

**Platform(s)**
macOS Tahoe

**Installation Method(s)**
Building from source using the Homebrew and MacPorts package managers.

**Describe the bug**
Local builds are failing some Java tests.   This issue has been reported by two users (MarkW and DarrenG).  It is odd that local builds are failing tests, but the official build on Jenkins is passing tests.   Perhaps the local builds are failing because of system configuration (Tahoe vs. Sequoia, different development tools, Python virtual environments, etcetera).

MarkW reports this failing test:
- Java 98 = `MdsTreeNodeTest` -- fails on the `testTimestampedSegments()` subtest because `putRow()` throws an exception.   Happens with both Homebrew and MacPorts.   MarkW runs this test sequentially and it consistently fails.   DarrenG reported one failure of the test.

DarrenG reports these failing tests:
- Java 102 = `MdsShr_Test`
- Java 107 = `Descriptor_A_Test`
- Java 108 = `Descriptor_CA_Test`

**To Reproduce**

1. Install the Homebrew or MacPorts package manager on a Mac computer that has an Apple Silicon CPU (aka the M-series CPUs).
2. Using the package manager, install the various third-party packages required by MDSplus (e.g., bison, flex, readline, etcetera).
3. Follow the `README.md` instructions for building / testing from source (i.e., the `deploy/build.py --os=macosx-macports --test` stuff).
4. Examine the test results to see if any of the Java tests failed.

**Expected behavior**
Local builds on Apple Silicon should pass all tests.

**Screenshots**
n/a

**Additional context**
Because the builds on the Jenkins server pass all tests, but local builds fail some tests, it is likely that the root cause is related to differences in system configuration.   If this conjecture proves to be true, then it is possible that the source code will remain unchanged.   Instead, we might need to write documentation about how to configure Apple Silicon computers for local builds.

mwinkel-dev:
There is a chance that the Java CLASSPATH stuff in PR #2966 is a factor in this issue.

WhoBrokeTheBuild:
[...]

PR #2978, opened 2025-10-27
Fix: issue with putRow in MdsTreeNodeTest caused by the optimizer

WhoBrokeTheBuild:
Mark `vars` (from `INIT_VARS`) as volatile to prevent it from being optimized into an issue

This serves as a partial fix for #2965


mwinkel-dev (approved):
Good catch!   This is a simple, but important change.   Looks good.

heidthecamp (approved):

PR #2980, opened 2025-10-28
Build: test release builds to find optimizer issues

heidthecamp:
Add stage for build and test release.

Note: `RelWithDebInfo` only uses `-O2` but this has found all optimizer issues so far.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,10 +65,20 @@ def testStage(os) {
     }
 
     return {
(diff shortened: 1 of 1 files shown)


mwinkel-dev:
This PR looks good.

WhoBrokeTheBuild (approved):
I'm approving this despite the failing tests, we should merge this now so we can sort out the failing tests that it found in other PRs

PR #2981, opened 2025-10-29
Build: add hashbang to scripts that need them for opinionated packaging systems

dgarnier:
Homebrew is quite opinionated about installing scripts in unix systems that are textual but do not have a hashbang (#!/bin/sh for example) start.  

These are the offending files that won't have their permissions set to executable unless they are amended.  These changes are inline with the other scripts, so making the entire system more self-consistent.

### DIFF
--- a/scripts/job_finish
+++ b/scripts/job_finish
@@ -1,3 +1,5 @@
+#!/bin/sh
+#
 # file: /usr/local/mdsplus/bin/job_finish
 #
 # shell script to signal completion of async daq task.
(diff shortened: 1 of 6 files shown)


mwinkel-dev:
Hi @dgarnier -- Thanks for the PR.   The changes look good to me.   Stephen has a bunch of post-CMake PRs, so I defer to him regarding approval of the PR (and schedule for merging it to alpha).

WhoBrokeTheBuild (approved):

PR #2985, opened 2025-11-13
Build: Use -fno-strict-aliasing to temporarily fix the optimizer issues

WhoBrokeTheBuild:
Revert packaging Debug builds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -576,6 +576,13 @@ else()
         COMPILER -Wextra
     )
 
+    # There are currently many issues when using strict aliasing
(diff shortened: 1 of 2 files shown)


heidthecamp (approved):

mwinkel-dev (approved):
Straightforward changes, so approving this PR.   The MDSplus code base is older than the "strict aliasing" rule, so it is no surprise that it has problems with "strict aliasing".

Correction: the "strict aliasing" rule was apparently present in the ANSI C89 standard.   Nonetheless, still not a surprise that it poses problems for MDSplus.

Issue #2977, opened 2025-10-27
mdstcl cli setevent fails on first attempt


rzorger:
**Affiliation**
PPPL

**Version(s) Affected**
stable_release-7-153-3

**Platform(s)**
Rocky Linux 9.6 (Blue Onyx)

**Installation Method(s)**
dnf install

**Describe the bug**
When starting the mdstcl command line and running the 'setevent' command, the first time it executes it has an error. If tried again it the command executes correctly.

**To Reproduce**
 mdstcl 
TCL> setevent TEST_EVENT
mdsdcl: --> failed on line 'setevent TEST_EVENT'
TCL> setevent TEST_EVENT
TCL> q

**Expected behavior**
setevent should work the first time it is called.

**Screenshots**

**Additional context**
I tried this on two different Rocky Linux 9.6 servers we have and it occurred on both of them.

joshStillerman:
Please paste output of : 
```
TCL> show version
```

mwinkel-dev:
Hi @rzorger -- Thanks for filing this issue in GitHub (in response to our discussion on Discord).   Josh and I have both confirmed that the problem doesn't occur in the current alpha release.   So to investigate this issue, we will have to install stable-7-153-3 on RHEL9.   We will do that experiment later today and post our findings here.

mwinkel-dev:
Hi @rzorger -- I was unable to reproduce this problem using Rocky Linux 9.6 and MDSplus stable-7-153-3.   

So now the current conjecture is that there is something about the environment on your two Rocky 9.6 servers that is triggering the `mdsdcl` error on the first `setevent`.   I will try some additional experiments to see if I can trigger this error.

rzorger:
TCL> show version


MDSplus version: 7.153.3
----------------------
  Release:  stable_release-7-153-3
  Date:     Mon Jun 30 23:00:29 UTC 2025
  Browse:   https://github.com/MDSplus/mdsplus/tree/stable_release-7-153-3
  Download: https://github.com/MDSplus/mdsplus/releases/tag/stable_release-7-153-3

mwinkel-dev:
Hi @rzorger -- Is it possible that your Rocky 9.6 servers have multiple versions of MDSplus installed?   If so, please open a terminal and run the following commands.   Check to see if all the paths are similar and what you expect.

```
$ echo $MDSPLUS_DIR
$ which mdstcl
$ which mdsdcl
```

mwinkel-dev:
[...]

PR #2986, opened 2025-11-18
Fix: TCP setevent no longer throws bogus error message

mwinkel-dev:
Fixes issue #2977 .

### DIFF
--- a/mdsshr/MdsEvents.c
+++ b/mdsshr/MdsEvents.c
@@ -908,7 +908,11 @@ int RemoteMDSEventCan(const int eventid)
 static int sendRemoteEvent(const char *const evname, const int data_len,
                            char *const data)
 {
-  int status = 1, i, tmp_status;
+  int status, i, send_status, ans_status;
+  int tries;
+  const int MAX_TRIES = 3;
+  int bad_server = FALSE;
+  int bad_status;
   char expression[256];
   struct descrip ansarg;
   struct descrip desc;
@@ -927,36 +931,49 @@ static int sendRemoteEvent(const char *const evname, const int data_len,
   desc.dims[0] = data_len;
   ansarg.ptr = 0;
   sprintf(expression, "setevent(\"%s\"%s)", evname, data_len > 0 ? ",$" : "");
-  if (STATUS_OK)
+
+  for (i = 0; i < num_send_servers; i++)
   {
-    int reconnects = 0;
-    tmp_status = 0;
-    for (i = 0; i < num_send_servers; i++)
-    {
+    for (tries = 0; tries < MAX_TRIES; tries++) {
+      send_status = MDSplusERROR;
+      ans_status = MDSplusERROR;
+
       if (send_ids[i] > INVALID_CONNECTION_ID)
       {
-        if (data_len > 0)
-          tmp_status =
-              MdsValue_(send_ids[i], expression, &desc, &ansarg, NULL);
-        else
-          tmp_status =
-              MdsValue_(send_ids[i], expression, &ansarg, NULL, NULL);
-      }
-      if (tmp_status & 1) {
-        tmp_status = (ansarg.ptr != NULL) ? *(int *)ansarg.ptr : 0;
-      }
-      status = tmp_status;
-      if (!(tmp_status & 1))
-      {
-        if (reconnects < 3)
-        {
-          ReconnectToServer(i, 0);
-          reconnects++;
-          i--;
+        if (data_len > 0) {
+          send_status = MdsValue_(send_ids[i], expression, &desc, &ansarg, NULL);
+        } else {
+          send_status = MdsValue_(send_ids[i], expression, &ansarg, NULL, NULL);
+        }
+        if (IS_OK(send_status)) {
+          ans_status = (ansarg.ptr != NULL) ? *(int *)ansarg.ptr : MDSplusERROR;
         }
       }
-      free(ansarg.ptr);
+
+      if (IS_OK(send_status) && IS_OK(ans_status)) {
+        break;
+      } else {
+          ReconnectToServer(i, 0);
+      }
     }
+
+    if (tries >= MAX_TRIES) {
+      bad_server = TRUE;
+      if (IS_NOT_OK(send_status)) {
+        bad_status = send_status;
+      } else {
+        bad_status = ans_status;
+      }
+    }
+    free(ansarg.ptr);
+  }
+
+  // If one or more of the servers in the list cannot be reached, display the status of
+  // the last bad server.   Otherwise, status is the answer from the last good server.
+  if (bad_server) {
+    status = bad_status;
+  } else {
+    status = ans_status;
   }
   return status;
 }
--- a/mdsshr/MdsEvents.c
+++ b/mdsshr/MdsEvents.c
@@ -402,7 +402,7 @@ static void *handleRemoteAst(void *arg )
   for (idx = 0; idx < num_receive_servers; idx++)
   {
     receive_thread_ids[idx] = searchOpenServer(receive_servers[idx]);
-    if(receive_thread_ids[idx] < 0)
+    if(receive_thread_ids[idx] <= INVALID_CONNECTION_ID)
     {
       receive_thread_ids[idx] = ConnectToMds_(receive_servers[idx]);
     }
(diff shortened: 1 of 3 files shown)


mwinkel-dev:
This WIP must be carefully scrutinized to ensure it is not a breaking change for the "events" feature of MDSplus.   When reviewing this proposed fix, it will likely be useful to study the individual commits.

Commit 9874518:  This is the minimum change needed to fix the problem reported by PPPL.   The issue is that because PR #2288 redefined `INVALID_CONNECTION_ID = -1` instead of zero, the TCP `setevent` code was not correctly handling a connection ID of zero, thus it reconnected as connection ID = 1, which did send the event, but also generated a bogus error message.   The fix in this first commit was sufficient to prove that the root cause had been identified.   However, it is a partial fix because it only works if the `mds_event_target` environment variable contains just a single TCP event server.

Commit 23e2581:  Updates the `MdsEvents.c` file to use the `INVALID_CONNECTION_ID` define throughout, thereby eliminating the confusion caused by the old code (which incorrectly assumes zero is the invalid ID).

Commit 9366af2:  The `sendRemoteEvent()` function returns an MDSplus status code (32-bit integer with three fields) up the call stack, which is what the TCL `setevent` feature needs.   However, the separate `setevent` utility runs at the operating system prompt and thus should use C exit codes when terminating.   For clarity, this is done with the `C_OK` and `C_ERROR` defines.

Commits 6721c2e and 1079327:  For clarity, the separate `wfevent` utility was also changed to use `C_OK` and `C_ERROR`.

Commit 8d7e8192: Significant rewrite of the `sendRemoteEvent()` function to correctly process a list of servers specified by the `mds_event_target` environment variable.   The existing code was only retrying a connection on the first failed server that was encountered.   This rewrite assumes that an effort should be made to write to every server in the list even if some servers cannot be reached.   An alternate design would be to terminate as soon as the first failing server is encountered.

mwinkel-dev:
To manually test these changes, the following configuration was used.

- Two VMs, A and B, were used.   (The operating system doesn't matter.)
- Server A had four terminal windows open, with an `mdsip` process running in each window.   (Each `mdsip` process used a different port.).  When server A was a Rocky Linux VM, the software firewall, `firewalld`, was configured to allow TCP traffic on the various `mdsip` ports.  The firewall was also configured for UDP traffic on port 4000.
- Server B ran a developer build with this WIP.   To force TCP events to be used, this statement was executed: `export mds_event_target="<server A>:<port 1>;<A>:<port 2>;<A>:<port 3>;<A>:<port 4>"`.   Note that the double quotes and semicolon are important.
- On server B, two terminal windows were opened.   One was used to run the standalone `wfevent` utility to confirm that TCP events were indeed being sent.  The other terminal windows was used to run 1) TCL's `setevent` command and 2) the separate `setevent` utility.
- After running a test, the messages written by the `mdsip` processes on server A were also reviewed.
- To test a failed connection, B's terminal windows did `wfevent`, then on A one of the `mdsip` processes was terminated, and then on B the TCL `setevent` or separate `setevent` was issued.

mdsshr/MdsEvents.c
```diff
@@ -927,35 +931,49 @@ static int sendRemoteEvent(const char *const evname, const int data_len,
...
+      } else {
           ReconnectToServer(i, 0);
-          reconnects++;
-          i--;
-        }
       }
-      free(ansarg.ptr);
     }
+
+    if (tries >= MAX_TRIES) {
```
mwinkel-dev, line 960: Question: When a bad server is encountered, should the function continue sending the event to the remaining servers (as per this WIP)?   Or should the function instead immediately terminate?

setevent/setevent.c
```diff
@@ -32,12 +33,17 @@ int main(int argc, char **argv)
   if (argc < 2)
   {
     printf("Usage: %s <event> [optional-data-string]\n", argv[0]);
-    status = 0;
+    status = C_OK;
   }
   else
   {
     int len = (int)((argc > 2) ? strlen(argv[2]) : 0);
     status = MDSEvent(argv[1], len, argv[2]);
```
mwinkel-dev, line 41: Note that the old code was returning an MDSplus status code (32-bits, 3 fields) when the function terminates.   This is problematic because `MDSplusSUCCESS = 65545`, which when treated as a C exit code (a byte value) is a failure (i.e., `echo $?` displays `9` which is a non-zero value, thus denotes a C failure).

wfevent/wfevent.c
```diff
@@ -96,7 +96,7 @@ int main(int argc, char **argv)
         break;
       default:
         printhelp(argv[0]);
-        return 1;
+        return C_ERROR;
```
mwinkel-dev, line 99: Note that this is a change in behavior.   The define is `C_ERROR = -1`, which `echo $?` displays as `255`.

The conjecture is that very few users have shell scripts that check the status code of `wfevent`.   Thus, this change is likely very low risk.

Alternatively, we could add a new define, `C_FAILURE = 1`, and use it instead of `C_ERROR`.

mdsshr/MdsEvents.c
```diff
@@ -402,7 +402,7 @@ static void *handleRemoteAst(void *arg )
   for (idx = 0; idx < num_receive_servers; idx++)
   {
     receive_thread_ids[idx] = searchOpenServer(receive_servers[idx]);
-    if(receive_thread_ids[idx] < 0)
+    if(receive_thread_ids[idx] <= INVALID_CONNECTION_ID)
```
mwinkel-dev, line 405: Note that around line 413 below, an error condition is detected whereupon a socket is set to zero (i.e., stdin) which seems a bit odd.   The code works "as is", so it is probably OK to keep it.

mwinkel-dev:
There appears to be yet another issue with TCP events, thus this WIP definitely needs more work.

To test TCP events, it is advisable to use three virtual machines, call them A, B and C.

A sets `mds_event_target` to point to B.   B runs the `mdsip` process that relays the TCP event.   C sets `mds_event_server` to B.   On all VMs, make sure the software firewall permits TCP traffic on the desired port.

The test consists of the following:

1. On C, do `wfevent test_event`
2. On A, do `setevent test_event`
3. On C, check to see if the `test_event` was received
4. On B, check the messages displayed by the `mdsip` process to troubleshoot any issues.

The above test revealed that C would only receive the `test_event` if B's software firewall was configured to allow UDP traffic on port 4000 (the default for MDSplus UDP events).   Because the three VMs are on the same virtual network, it is probable that B is sending out UDP events that are received by A and C.

Josh has confirmed that TCP events should work without using UDP on any of the VMs.   (Similar behavior was seen months ago, but it was incorrectly assumed that TCP events were designed to convert to UDP at the target system.)

Thus, additional investigation is required.

mwinkel-dev:
Experiments prove that TCP events are indeed being converted to UDP events at the "target" system.   This occurs with both TCL's `setevent` command and the separate `setevent` utility.   Fixing TCP events will likely be complicated, so that task should probably should be split off as a new issue / PR.

Here are the details . . . 

Both TCL and the `setevent`utility call the `MDSEvent()` function of `MdsEvents.c`, which detects the `mds_event_target` environment variable and eventually causes `sendRemoteEvent()` to be called.   That function sends a TDI expression to the "target" system that causes it to evaluate the `setevent.fun` TDI function, which in turn causes the "target" to also call its `MDSEvent()` function.   However, on the "target", the `mds_event_target` and `mds_event_server` environment variables are undefined, and thus the target's `MDSEvent()` calls `MDSUdpEvent()`.   Thus the "target" converts TCP events into UDP events.

[references TDI function](https://github.com/MDSplus/mdsplus/blob/3360cd9ef6da595dfd482dba01faf63179f6cc33/mdsshr/MdsEvents.c#L929)

[setevent.fun](https://github.com/MDSplus/mdsplus/blob/3360cd9ef6da595dfd482dba01faf63179f6cc33/tdi/mdsshr/setevent.fun#L1)

[MDSEvent](https://github.com/MDSplus/mdsplus/blob/3360cd9ef6da595dfd482dba01faf63179f6cc33/mdsshr/MdsEvents.c#L1024)

mwinkel-dev:
A tentative conclusion from additional experiments is that UDP is the backbone of TCP events.   

Assume four servers: A, B, C, and D.   B and C are on the same network segment.   D executes a `wfevent test_event` command that uses TCP to listen to an mdsip process on C.   The A server does a `setevent test_event` that uses TCP to send an event to an mdsip process on B.   B converts the TCP event to a UDP event, which is seen by the mdsip process on C, which in turn then uses TCP to notify D.

It is difficult to test the above configuration when all the four systems are VMs are on the same virtual network.   However, based on experiments and perusing the source code, it seems likely that UDP is the backbone with TCP being used for the first and last network hop in the chain.

More investigation is required.

mwinkel-dev:
An additional experiment using debug print statements proves the preceding conjecture that TCP events are converted to / from UDP events.

WhoBrokeTheBuild (approved):
These all look like good changes, and I think the risks are minimal. LGTM

Issue #2987, opened 2025-11-19
The MATLAB mdstest.m function is broken


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
Developer build of current alpha

**Platform(s)**
Found in Ubuntu24 and macOS Tahoe, but surely applies to all platforms

**Installation Method(s)**
Dev build installed

**Describe the bug**
The `mdstest.m` script runs many tests.  Unfortunately, the `char` test is now failing.

The output from macOS Tahoe with MATLAB R2024b follows:

```
>> mdstest(false)  % test Java bridge
Error "string test": expected class char got string

>> mdstest(true)  % test Python bridge
Error "string test": expected class char got py.str
```

The output from Ubuntu24 with MATLAB R2025a was similar, but only failed on `mdstest(false)`.

**To Reproduce**
1. Install Java, Python and MATLAB
2. Install MDSplus
3. Configure MATLAB to use Java, Python and $MDSPLUS_DIR/matlab
4. Run MATLAB
5. At the MATLAB prompt, enter the `mdstest` commands shown above

**Expected behavior**
The tests should pass.

**Screenshots**
n/a

**Additional context**
n/a

mwinkel-dev:
A fix has been developed that works OK with MATLAB R2025a, OpenJDK21, and Ubuntu24.   However, this fix must also be tested on Windows and macOS.

Here are the details . . . 

MATLAB R2016b introduced a "string" datatype.   Prior to that text was represented as cell arrays of character vectors.

Presumably, the `mdstest.m` script was written prior to R2016b.   MDSplus probably doesn't need to maintain backwards compatibility with R2016b, thus it should be safe to rewrite the `mdstest.m` script for the "string" datatype.

Interestingly, only the Java bridge is returning the "string" datatype.   The Python bridge is still returning the older cell arrays of character vectors.

Also, because `mdstest.m` is written with the `&&` short-circuiting logical operator, it was masking two additional checks that failed.

mwinkel-dev:
Fixed by PR #2990 which was recently merged to alpha, so closing this issue.

PR #2990, opened 2025-11-25
Fix: upgrade the mdstest.m MATLAB script to work with newer software

mwinkel-dev:
This fixes issue #2987.

Surprisingly, MATLAB does not return the same "class" type across all platforms.   That caused the `mdstest.m` script to fail.

After the script was upgraded, it now passes all three bridges (Java, Python, mdsthin) on the following platforms.

- Ubuntu 24, MATLAB R2025a, OpenJDK 21.0.8, Python 3.12.3, NumPy 1.26.4
- macOS Tahoe, MATLAB R2024b, OpenJDK 21.0.8, Python 3.12.8, NumPy 2.2.2
- Win11, MATLAB R2025b, Java 1.8.0_471, Python 3.12.12, NumPy 2.3.5

### DIFF
--- a/matlab/mdstest.m
+++ b/matlab/mdstest.m
@@ -32,7 +32,9 @@ function result = mdstest(varargin)
     result = result && mdscheck('$ == $', 'uint8', [1, 1], 1, 2);
     result = result && mdscheck('QUADWORD_UNSIGNED(1:100)', 'uint64', [100, 1]);
     if info.usePython
-        if ismac || ispc
+        % Text can have different classes depending on the software versions
+        x = mdsvalue('"any_string"');
+        if strcmp(class(x), 'py.str')
             result = result && mdscheck('"string test"', 'py.str', [1, 11]);
         else 
             result = result && mdscheck('"string test"', 'char', [1, 11]);
--- a/matlab/mdstest.m
+++ b/matlab/mdstest.m
@@ -5,6 +5,7 @@ function result = mdstest(varargin)
 %   mdstest()   % tests current bridge
 %   mdstest(0)  % tests java bridge
 %   mdstest(1)  % tests python bridge
+%   mdstest(1,1)  % tests mdsthin bridge
 
     info = mdsInfo(varargin{:}); % update MDSINFO
     result = mdscheck('1BU', 'uint8', [1, 1]);
@@ -28,7 +29,7 @@ function result = mdstest(varargin)
     result = result && mdscheck('$ == $', 'uint8', [1, 1], 1, 2);
     result = result && mdscheck('QUADWORD_UNSIGNED(1:100)', 'uint64', [100, 1]);
     if info.usePython
-        if ismac
+        if ismac || ispc
             result = result && mdscheck('"string test"', 'py.str', [1, 11]);
         else 
             result = result && mdscheck('"string test"', 'char', [1, 11]);
--- a/matlab/mdstest.m
+++ b/matlab/mdstest.m
@@ -6,6 +6,7 @@ function result = mdstest(varargin)
 %   mdstest(0)  % tests java bridge
 %   mdstest(1)  % tests python bridge
 
+    global MDSINFO;
     mdsInfo(varargin{:}); % update MDSINFO
     result = mdscheck('1BU', 'uint8', [1, 1]);
     result = result && mdscheck('1WU', 'uint16', [1, 1]);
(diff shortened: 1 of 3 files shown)


mwinkel-dev:
The `mdstest.m` script needs more work to use the mdsthin-bridge.   Presently, the script is defaulting to the python-bridge even when the user requests use of the mdsthin-bridge.

mwinkel-dev:
**CORRECTION:**
Additional experiments prove that everything is working as intended.   The source of the confusion is that the new documentation states that `mdstest.m` is testing the bridges, but in reality it is merely testing MATLAB to see if it is configured to access the Java and Python programming languages.

Note that `mdstest.m` **does not** establish an `mdsip` connection to a remote MDSplus archive server, thus none of the three bridges (Java, Python, mdsthin) is exercised.

If `mdstest(true, true)` remains in the new documentation, then the comment line about that should remain in `mdstest.m`, even though `mdstest(true, true)` is equivalent to `mdstest(true)`.

Alternatively, if `mdstest(true, true)` is removed from the new documentation, then the associated comment line should be removed from `mdstest.m`.

matlab/mdstest.m
```diff
@@ -5,15 +5,15 @@
 %   mdstest()   % tests current bridge
 %   mdstest(0)  % tests java bridge
 %   mdstest(1)  % tests python bridge
+%   mdstest(1,1)  % tests mdsthin bridge
```
mwinkel-dev, line 8: Should this `mdstest(1,1)` comment line be deleted or retained?   

Note that this MATLAB script tests connection to the Java and Python languages; it does not establish a bridge connection via mdsip to a remote MDSplus archive server.

mwinkel-dev:
Retest this please

matlab/mdstest.m
WhoBrokeTheBuild: Perhaps the comment should be expanded, explaining why it won't work

matlab/mdstest.m
```diff
@@ -28,8 +28,19 @@
     result = result && mdscheck('$ : $', 'int32', [100, 1], int32(1), int32(100));
     result = result && mdscheck('$ == $', 'uint8', [1, 1], 1, 2);
     result = result && mdscheck('QUADWORD_UNSIGNED(1:100)', 'uint64', [100, 1]);
-    result = result && mdscheck('["a","b","c","d"]', 'cell', [4, 1]);
-    result = result && mdscheck('set_range(2,3,["a","b","c long string","d","e","f"])', 'cell', [2, 3]);
+    if info.usePython
+        if ismac || ispc
+            result = result && mdscheck('"string test"', 'py.str', [1, 11]);
```
WhoBrokeTheBuild, line 38: What an odd problem, does py.str work for all platforms or only mac/win?

WhoBrokeTheBuild (dismissed):

matlab/mdstest.m
heidthecamp, line 53: Just double checking my assumptions

This line doesn't cause issues because we are programmaticly passing the expected `result_class` and so it could be `py.str` but would still need to be converted to string later.

matlab/mdstest.m
mwinkel-dev: Good point.   I will revise / expand the comment for `mdstest(1,1)`.

matlab/mdstest.m
mwinkel-dev: My recollection (perhaps wrong) is that this `char` vs. `py.str` issue only applies to macOS and Windows.

As described in the initial post (way above), the Python version across all three platforms was 3.12.x, and the MATLAB versions were also in the same time range (R2024* and R2025*).   The only significant difference is the NumPy version, which was older on Ubuntu.

I will update the software on all three VMs and repeat the testing.   It might be that `py.str` now also applies to Linux.

heidthecamp (dismissed):
This looks good to me. Did you get a chance to check the the `mdstest.m` line51? ```if ~strcmp(class(x), result_class)```

matlab/mdstest.m
mwinkel-dev: Changed the code to make it more general.   When testing strings in the Python-bridge, it now accepts both `char` and `py.str`.

This a fix is a bit inelegant, but was deemed less risky than rewriting the `mdscheck()` function to support multiple classes.

matlab/mdstest.m
mwinkel-dev: Correct.

WhoBrokeTheBuild (approved):

mwinkel-dev:
Hi @heidthecamp -- Yes, the `~strcmp(class(x), result_class)` line is A-OK.   The `result_class` is passed in as either `char` or `py.str`.   The `class(x)` is a string representing the MATLAB class returned when it calls the local MDSplus to evaluate the input expression, `' "string test" '`.   If the returned class is not `char` or `py.str`, the associated `mdscheck()` test will throw an error which indicates that the `mdstest(1)` of the Python-bridge failed.

Elsewhere in the code, the `py.str` class is converted into a MATLAB `string`.

mwinkel-dev:
This improved fix just passed testing on three platforms: Ubuntu, macOS and Windows.   Thus, now merging it to alpha.

PR #2994, opened 2025-12-03
Fix: Add better handling of division by 0 (int and float) for MOD()

santorofer:
This is intended to be a discussion, there are several ways this can be handled for integers. Regardless, $ROPRAND is well suited for handling divide by zero errors for floating points.

Option 1: We could make both DIVIDE and MOD return the new TdiDIV_BY_ZERO error when dividing by zero
Option 2: We could make only MOD return this, and leave DIVIDE returning 0
Option 3: Make MOD return 0 as well and not add TdiDIV_BY_ZERO

This moves MOD (mod_float, mod_bin, OperateBin, Tdi3Mod) into TdiDivide.c
This currently adds a new TdiDIV_BY_ZERO error to tdishr_messages.xml (see above)

This is intended to fix the error handling for `0 % 0`, which currently segfaults.

### DIFF
--- a/tdishr/TdiMath2.c
+++ b/tdishr/TdiMath2.c
@@ -69,8 +69,6 @@ struct descriptor *out)
 #endif
 
 extern int CvtConvertFloat();
-extern double WideIntToDouble();
-extern void DoubleToWideInt();
 
 const double radians_to_degrees = 180. / M_PI;
 
@@ -155,15 +153,6 @@ const double radians_to_degrees = 180. / M_PI;
 
 static const int roprand = 0x8000;
 
-static inline double mod_float(double x, double m)
-{
-  if (m == 0.0)
-    return x;
-  double intpart;
-  modf(x / m, &intpart);
-  return x - intpart * m;
-}
-
 #define OperateFloatOne(dtype, routine, p1, p2)                     \
   {                                                                 \
     double a, b, ans;                                               \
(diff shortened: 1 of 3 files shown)


mwinkel-dev:
I've glanced at this PR and it looks good, however I will do a more thorough review later.   

Here are some initial comments . . . 

- I favor adding the new `DIV_BY_ZERO` status code for division even though it might be a breaking change for some customers.   However, as per this morning's team meeting, this proposed change merits additional discussion.

- The `$ROPRAND` constant (for missing data) should be included in the new TDI documentation.   (It is described in the existing Wiki, but not with the other constants.)

tdishr/TdiMath2.c
```diff
@@ -186,7 +175,7 @@ static inline double mod_float(double x, double m)
     {                                                          \
     case 0:                                                    \
     case 3:                                                    \
-      while (nout--)                                           \
```
heidthecamp, line 189: This won't break anything but isn't needed.

heidthecamp (changes requested):
Over all good. But there was an erroneous white space change

joshStillerman:
I concur that floating point behavior should remain.  It is complicated but useful that operators of $roperand return $roperand.

maybe we should return $roperand regardless of input types and throw a new error when trying to convert $ROPERAND to integer types?

Issue #3007, opened 2025-12-12
The "PreIDL52" directory dates to 1998 and should be removed from the IDL API


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
all versions

**Platform(s)**
all platforms

**Installation Method(s)**
all methods

**Describe the bug**
The `$MDSPLUS_DIR/idl/PreIDL52` directory contains 8 support functions that allow IDL 5.1 (and earlier) to execute code written for IDL 5.2 and later.   IDL 5.2 was released in 1998.   It is improbable (impossible?) that any customers are running versions older than IDL 5.2 at this time.   Therefore, the `PreIDL52` directory should be deleted.

**To Reproduce**
Using a web browser, go to the MDSplus GitHub site, click on the `idl/PreIDL52` directory, and examine the dates on the 8 functions.

**Expected behavior**
The `PreIDL52` directory should not be shipped to customers and should be removed from the repository.

**Screenshots**
n/a

**Additional context**
n/a

mwinkel-dev:
This issue can be closed when PR #3008 is reviewed and merged.

mwinkel-dev:
Fixed by PR #3008, so closing this issue.

PR #3008, opened 2025-12-16
Fix: delete unused functions to support IDL 5.1 and earlier

mwinkel-dev:
This fixes Issue #3007.

Deletes eight support functions written around 1998 to support backwards compatibility with early versions of the IDL programming language.


WhoBrokeTheBuild (approved):
Love to see some ~spring~ winter cleaning, this LGTM

PR #3012, opened 2026-01-12
Feature: Implement RESHAPE() in TDI

WhoBrokeTheBuild:
This builtin was never implemented, but the opcode was allocated and it existed in the original documentation. The similarity to numpy's reshape function means users might actually want to use this, and throwing a terse error is less than helpful.

This implements RESHAPE() with only the first two arguments of SOURCE and SHAPE. The optional arguments of PAD and ORDER described in the original documetnation are not implemented, and won't be included in the new documentation for this function.

```
TDI> reshape(1:6, [2, 3])
[[1,2], [3,4], [5,6]]

TDI> reshape(1:6, [3, 2])
[[1,2,3], [4,5,6]]

TDI> reshape(1:8, [2, 2, 2])
[[[1,2], [3,4]], [[5,6], [7,8]]]

TDI> reshape(reshape(1:8, [2, 4]), [8])
[1,2,3,4,5,6,7,8]

TDI> reshape([[1,2], [3,4]], [1, 4])
[[1], [2], [3], [4]]
```

### DIFF
--- a/tdishr/TdiReshape.c
+++ b/tdishr/TdiReshape.c
@@ -28,6 +28,7 @@ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #include <mds_stdarg.h>
 #include <mdsdescrip.h>
 #include <tdishr.h>
+#include <mdsshr.h>
 
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


tdishr/TdiReshape.c
```diff
@@ -49,7 +50,72 @@ int Tdi1Reshape(opcode_t opcode, int narg, struct descriptor *list[],
...
+        .scale = arr->scale,
+        .digits = arr->digits,
+        .aflags = arr->aflags,
+        .dimct = dimct,
+        .arsize = arr->arsize,
+      };
+
+      new_array.aflags.coeff = 1;
+      new_array.dimct = dimct;
+      new_array.a0 = new_array.pointer;
```
joshStillerman, line 104: it points at the same memory - good
Do you check if it is the right size for the dimensions ?

joshStillerman (commented):
it points at the same memory - good
Do you check if it is the right size for the dimensions ?

tdishr/TdiReshape.c
WhoBrokeTheBuild: Yep! That's done up on line 77, I compare the original count to the product of the shape, which should be the new size

joshStillerman (approved):

PR #3013, opened 2026-01-13
Fix: Make sure top node's parent state is on

### LINKED ISSUE #2953: It says parent node of top node is off, how to fix?
**Affiliation**
Columbia University

**Question**
The top node of my tree says that its parent node is off.  I cloned from a model tree that has this property.  How to turn the parent node of top to be on?? (getParent just returns itself).

joshStillerman:
Fixes: https://github.com/MDSplus/mdsplus/issues/2953

if the parent of a subtree is turned off, the parent_state of its top node is set to off.

Subsequently if the subtree is opened by itself, the top node's parent_state is off.  Since it has no parent in this scenario its parent_state is its own state.

Changed _TreeTurnOn to set the parent_state to on if the node in question is the top node.

Also changes SetNci flags, if it changes the state flag, then it also calls SetParentState


joshStillerman:
This turned out to not be a good solution.  Now addressed by:
https://github.com/MDSplus/mdsplus/pull/3018

Issue #3015, opened 2026-01-13
you can not rename nodes of usage subtree


joshStillerman:
**Affiliation**
MIT Plasma Science and Fusion Center

**Version(s) Affected**
All I presume.  I am running:
```
MDSplus version: 7.157.0
----------------------
  Release:  alpha_release-7-157-0
  Date:     Thu Nov 20 23:04:32 UTC 2025
  Browse:   https://github.com/MDSplus/mdsplus/tree/alpha_release-7-157-0
  Download: https://github.com/MDSplus/mdsplus/releases/tag/alpha_release-7-157-0
```

**Platform(s)**
All I presume.
I am running Ubuntu

**Installation Method(s)**
All I presume.
The behavior is there when built from sources and when installed from kits.

**Describe the bug**
mdstcl throws an error when you ask it to rename a node who's usage is subtree:
```
TCL> rename .subtree .subsub.subtree
Error: Problem renaming node .subtree to .subsub.subtree
Error message was: %TREE-E-INVPATH, Invalid tree pathname specified
mdsdcl: --> failed on line 'rename .subtree .subsub.subtree'
TCL> 
```

mwinkel-dev:
Josh says fixed by PR 3017, thus closing this.

PR #3017, opened 2026-01-14
Fix: can not rename nodes of usages subtree

joshStillerman:
This change addresses
    https://github.com/MDSplus/mdsplus/issues/3015

The check that members are not renamed to be children was also preventing subtrees from being renamed as children.

### DIFF
--- a/treeshr/TreeRenameNode.c
+++ b/treeshr/TreeRenameNode.c
@@ -139,7 +139,7 @@ int _TreeRenameNode(void *dbid, int nid, char const *newname)
    ************************************************/
   if (is_child)
   {
-    if (oldnode_ptr->usage != TreeUSAGE_STRUCTURE)
+    if ((oldnode_ptr->usage != TreeUSAGE_STRUCTURE) && (oldnode_ptr->usage != TreeUSAGE_SUBTREE))
     {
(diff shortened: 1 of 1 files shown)


joshStillerman:
retest this please

treeshr/TreeRenameNode.c
```diff
@@ -139,7 +139,7 @@ int _TreeRenameNode(void *dbid, int nid, char const *newname)
    ************************************************/
   if (is_child)
   {
-    if (oldnode_ptr->usage != TreeUSAGE_STRUCTURE)
```
WhoBrokeTheBuild, line 142: This looks good, can you change the comment above to say `non-STRUCTURE/SUBTREE` ?

WhoBrokeTheBuild (changes requested):

treeshr/TreeRenameNode.c
santorofer: Yes, the changes make sense and look good. I would change the comment per Stephen's comment.

santorofer (approved):

joshStillerman:
retest this please

WhoBrokeTheBuild (approved):

PR #3025, opened 2026-02-18
Build: Change server IDL tests point to

heidthecamp:
Point server for IDL tests to `alcdata` since `alcdaq6` is retired.


WhoBrokeTheBuild (dismissed):

mwinkel-dev (approved):
Approving this temporary change so that Jenkins builds can resume.

However, we should revisit this in the coming weeks.   We never did figure out why IDL tests behaved differently on mfedata-archives than on the old acldata-archives.   It now appears that the new alcdata also causes the IDL test to behave differently.   This is a low priority, but we should eventually investigate and find the root cause.

PR #3027, opened 2026-02-19
Feature: Change python __version__ to semver

### LINKED ISSUE #3023: PEP compliant `__version__` for python
**Affiliation**
General Atomics
**Description**
The current cmake build process yields PEP incompliant `__version__` strings that are separated by dashes instead of dots. This is not great for the conda-forge releases of MDSplus we maintain at GA . I was hoping by just setting 
`-DRELEASE_TAG=7.157.0` to a proper version string I could trick the `cmake` process but that just crashes. For now I will manually hack the `__version__` after the cmake build is complete but that is not a great forever solution.

WhoBrokeTheBuild:
Change the python __version__ variable to use standard semver format, e.g. "1.2.3" instead of the MDSplus release tag format "alpha_release-1-2-3"
Change the version check to use the release tag instead of __version__

This should fix #3023

### DIFF
--- a/python/MDSplus/__init__.py
+++ b/python/MDSplus/__init__.py
@@ -78,13 +78,14 @@ class libs:
 # check version
 if __name__ == "MDSplus":
     version_check = os.getenv("MDSPLUS_VERSION_CHECK", "on").lower()
-    version_check = not (
-        version_check == "0" or version_check == "off" or version_check == "no")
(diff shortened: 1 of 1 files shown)


mwinkel-dev (approved):
This change looks good.

heidthecamp (approved):
This looks good to me.

Issue #3024, opened 2026-02-17
mdstcl do/method initial call fails


rzorger:
**Affiliation**
PPPL

**Version(s) Affected**
MDSplus version: 7.153.3
----------------------
  Release:  stable_release-7-153-3

**Platform(s)**
Rocky Linux 9.7

**Installation Method(s)**
 mdsplus.org repository

**Describe the bug**

**To Reproduce**
I am seeing another fail on first attempt bug in mdstcl. I am testing the action nodes in our tree and the first attempt to 'do/method' in a mdstcl session always fails. Subsequent attempts work:
mdstcl 
TCL> set tree rf
TCL> do/method \RF::FW_H908_02 init
%CAMERR, module SFW_H908_02, bad status =           0

Error executing device method
Error message was: %TDI-E-ABORT, Program requested abort
mdsdcl: --> failed on line 'do/method \RF::FW_H908_02 init'
TCL> do/method \RF::FW_H908_02 init
TCL> do/method \RF::FW_H908_02 init
TCL> do/method \RF::FW_H908_02 init
TCL> q
[root@skylark2 dispatcher]# mdstcl 
TCL> set tree rf
TCL> do/method \RF::FW_H908_01 init
%CAMERR, module sfw_h908_01, bad status =           0

Error executing device method
Error message was: %TDI-E-ABORT, Program requested abort
mdsdcl: --> failed on line 'do/method \RF::FW_H908_01 init'
TCL> do/method \RF::FW_H908_01 init
TCL> do/method \RF::FW_H908_01 init
TCL> q

This is like issue #2977

**Expected behavior**
If this is a bug please fix and release a new stable version

**Screenshots**

**Additional context**
I need this to troubleshoot our action nodes and get the shot cycle to work under Rocky 9

mwinkel-dev:
Hi @rzorger -- I will use a Rocky 9.7 virtual machine later today to investigate this issue.   Thanks also for mentioning that it is similar to the closed issue, #2977, as that is likely an important clue.

mwinkel-dev:
Hi @rzorger -- I am assuming that failing action is associated with the `H908` device provided by "MitDevices".   If that assumption is incorrect (i.e., if it is instead a device driver that PPPL wrote), let us know.

mwinkel-dev:
Hi @rzorger -- I now have a Rocky 9.7 (Blue Onyx) VM with MDSplus stable-7-153-3.

This issue will be investigated with two parallel tracks:
[...]

PR #3028, opened 2026-02-20
Fix: the CAMAC libraries now accept connection IDs of zero

mwinkel-dev:
This fixes Issue #3024.

PR #2288 redefined `INVALID_CONNECTION_ID` to be `-1` (it used to be zero).   This fix changes two CAMAC libraries, `libCamShr.so` and `libRemCamShr.so`, to accept zero as a valid connection ID.   Developer builds of the libraries were tested by the user and confirmed to fix the problem.   This fix needs to be included in a new "stable" release.

### DIFF
--- a/camshr/RemCamMulti.c
+++ b/camshr/RemCamMulti.c
@@ -102,7 +102,7 @@ static int DoCamMulti(char *routine, char *name, int a, int f, int count,
   int serverid = RemoteServerId();
   int status = 0;
   int writeData;
-  if (serverid)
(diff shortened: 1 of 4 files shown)


mwinkel-dev:
Retest this please

WhoBrokeTheBuild (approved):

mwinkel-dev:
PPPL confirmed that the two MDSplus servers involved with the CAMAC devices will both be running the forthcoming stable release.   Thus, this CAMAC fix will work fine on those servers.

PR #3032, opened 2026-03-05
Codex/camshr sqlite alpha pr

joshStillerman:
Please do not review or merge - 
 This is a generated solution that may or may not be viable.


mwinkel-dev:
I've perused the AI generated code.   Interesting experiment.

Some observations:
- Large test suite.
- The AI code is readable, but at times obscure.
- AI doesn't comment the code it writes.
- This prototype maintains compatibility with the old `crate.db` and `cts.db` files, plus adds SQLite.
- The code base is thus expanded, therefore there are more lines of code to maintain.

Issue #3035, opened 2026-03-14
mdscts occasionally refuses to delete the last crate listed in the crate.db file


mwinkel-dev:
**Affiliation**
MIT PSFC

**Version(s) Affected**
The MDSplus Version(s) affected, if any.
Client & Server: alpha-7-158-0

**Platform(s)**
Ubuntu-24, but presumably affects other platforms too

**Installation Method(s)**
Developer build.

**Describe the bug**
The `mdscts` utility is used to add CAMAC crates to the `crate.db` file and CAMAC modules to the `cts.db` file.   In some circumstances, the CTS `DelCrate` command refuses to delete the last crate listed in the `crate.db` file.   Presumably, the same problem exists regarding the last module in the `cts.db` file.

**To Reproduce**

1) Install alpha-7-158-0 on Ubuntu 24.
2) Create a "/tmp/cts" directory that will contain the two *.db files.
3) Use a text editor to create `cts.db` file that contains two bytes, a space and a newline.
4) Use a text editor to create this `crate.db` file (it can also be created using CTS `AddCrate` commands)

```
GKA401:...:.:.:.
GKA402:...:.:.:.
GKA403:...:.:.:.
GKA404:...:.:.:.
GKA405:...:.:.:.
GKA406:...:.:.:.
GKA407:...:.:.:.
GKA408:...:.:.:.
GKA409:...:.:.:.
GKA410:...:.:.:.
GKA411:...:.:.:.
GKA412:...:.:.:.
GKA413:...:.:.:.
```
5) Run `mdscts` and execute the following commands:

```
CTS> delcrate gka413
CTS> delcrate gka412
CTS> delcrate gka405
CTS> delcrate gka411
CTS> delcrate gka410
Error: entry 'GKA410' not found
mdsdcl: --> failed on line 'delcrate gka410'
```

**Expected behavior**
Displaying the `crate.db` file shows that `GKA410` does indeed exist in the file.   Therefore, `mdscts` should find the crate and delete it.   No errors should be thrown.

**Screenshots**
n/a

**Additional context**
n/a

mwinkel-dev:
Experiments with the debugger reveal that there are few related causes.

- When a crate is deleted, the list of crates of course shrinks but the last record in the previous list is not erased correctly.
- Which causes another routine to incorrectly count the number of active records in the `crate.db` file.
- Which in turn can cause the binary search to skip beyond the list of active records and search for the crate in the erased inactive records.
[...]

PR #3038, opened 2026-03-17
Fix: the mdscts DelCrate now works, plus Add_Entry was cleaned up

mwinkel-dev:
Fixes issue #3035.

The `mdscts` utility for managing CAMAC crates and modules had an "off by one" issue that caused it to occasionally refuse to delete crates.   (Presumably, the same issue affected deleting modules.)  Records in the `crate.db` and `cts.db` files have zero-based indices; the variable that tracks the number of records in each `*.db` file is one-based.

The `*.db` files are kept in sorted order and are accessed by mapping them into memory.   When records are added, the files grow in chunks of 10 records (one has an entry, the other nine are blank).   If more than 10 records are entered, then an additional chunk of 10 records is added.   (Actually, a bigger file is created, the old file is copied into the new file, and then the old file is deleted.)

Suppose a `*.db` file contains 10 non-empty records.   And then the second record is deleted.   In that case, records 3 through 10 are shifted up in the file, and then the tenth record is overwritten with a blank record.  (Note that the file size never shrinks; as records are deleted the file just fills up with blank records.)

Because of the "off by one" problem, the program incorrectly assumed that there was an extra record in the file, thus attempted to access a region beyond the end of the memory mapped file.  The end result was that instead of the file being padded with blank records after records were deleted, it was being padded with records that contained the NULL character (ASCII code = 0x00).   The `get_file_count()` routine was searching for records that contained a space, never found one, so incorrectly reported more active records in the file than there actually were.   That in turn messed up the binary search that is used to find the record to be deleted.   When the binary search did its midpoint calculation, it could skip over the last active record in the file and end up in the region of the file that only had records filled with NULL characters.   Thus, mdscts would occasionally refuse to delete an entry.

After fixing the deletion issue, the source code was reviewed for other "off by one" issues.   That revealed that the `add_entry()` routine was doing unnecessary record manipulation prior to calling the insertion sort.

Also, some source lines had been split into two lines, so those were reformatted to improve readability.

### DIFF
--- a/camshr/add_entry.c
+++ b/camshr/add_entry.c
@@ -114,14 +114,9 @@ int add_entry(int dbType, char *newEntry)
     goto AddEntry_Exit;
   }
 
-  // shift current entries by one
-  if (numOfEntries) // ... only if any entries exist
-    for (i = numOfEntries - 1; i >= 0; --i)
-      memcpy((char *)dbptr + ((i + 1) * entrySize),
-             (char *)dbptr + (i * entrySize), entrySize);
-
-  // put new entry at head of list
-  memcpy((char *)dbptr, newEntry, entrySize);
+  // put new entry at end of list
+  // Indices are zero based, numOfEntries is one based so appends to the list.
+  memcpy((char *)(dbptr + numOfEntries * entrySize), newEntry, entrySize);
 
   // insertion sort
   if (numOfEntries > 0) // only insert if more than one entry exists already
--- a/camshr/remove_entry.c
+++ b/camshr/remove_entry.c
@@ -131,13 +131,14 @@ int remove_entry(int dbType, int index)
   char line[BUFFER_SIZE];
 
   // remove entry -- move appropriate entries one earlier in list
-  for (i = index; i < numOfEntries; ++i)
+  // The index is zero based, but numOfEntries is one based hence the -1.
+  for (i = index; i < (numOfEntries - 1); ++i)
     memcpy((char *)dbptr + (i * entrySize),
             (char *)dbptr + ((i + 1) * entrySize), entrySize);
 
   // 'blank' out last, old entry
   sprintf(line, fmt, " ");
-  memcpy((char *)dbptr + ((i + 1) * entrySize), line, entrySize);
+  memcpy((char *)dbptr + (i * entrySize), line, entrySize);
 
   // commit change to file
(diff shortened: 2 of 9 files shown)


camshr/add_entry.c
```diff
@@ -114,14 +114,9 @@ int add_entry(int dbType, char *newEntry)
     goto AddEntry_Exit;
   }
 
-  // shift current entries by one
```
mwinkel-dev, line 117: This entire section is unnecessary.   The insertion sort routine, `issort()`, is designed for a new item to be appended to the sorted list.

The insertion sort is a general purpose routine, so does work if the new item is instead prepended.   However, that involves shifting all records in the database file, which means that the insertion sort will usually end up doing more work than it needs to (compared to appending the new item).   This is especially true if the user enters records in ascending order (e.g., crate GKA401, GKA402, GKA403, etcetera).

camshr/remove_entry.c
```diff
@@ -131,13 +130,14 @@ int remove_entry(int dbType, int index)
   char line[BUFFER_SIZE];
 
   // remove entry -- move appropriate entries one earlier in list
-  for (i = index; i < numOfEntries; ++i)
```
mwinkel-dev, line 134: Because `numOfEntries` is one-based, the `i+1` in the loop body was actually attempting to access beyond the end of the memory mapped `*.db` file.   (The record indices are 0 through `numOfEntries-1`.)

camshr/remove_entry.c
```diff
@@ -131,13 +130,14 @@ int remove_entry(int dbType, int index)
...
+  // The index is zero based, but numOfEntries is one based hence the -1.
+  for (i = index; i < (numOfEntries - 1); ++i)
     memcpy((char *)dbptr + (i * entrySize),
             (char *)dbptr + ((i + 1) * entrySize), entrySize);
 
   // 'blank' out last, old entry
   sprintf(line, fmt, " ");
-  memcpy((char *)dbptr + ((i + 1) * entrySize), line, entrySize);
+  memcpy((char *)dbptr + (i * entrySize), line, entrySize);
```
mwinkel-dev, line 140: When the loop exits, `i` is already the index for the "last, old entry".

mwinkel-dev:
Retest this please

WhoBrokeTheBuild (approved):

Issue #3034, opened 2026-03-11
jScope log y works but log x doesn't seem to work


nelson-zapenergy:
**Affiliation**
Zap Energy

**Question**
I have a TDI-callable Python script called rfft.py that takes an MDSplus Signal Node from the Tree that is open, calls the SciPy rfft (fft of a real array, returns only positive frequencies), and returns an MDSplus Signal with the frequency as its dimension.

It works fine if I select "Log scale" for the Y axis, but if choose "Log scale" for the X axis (with or without "Log scale" selected for the Y axis) there are no data plotted.  If I autoscale, I can drag the Point selection through the panel and the crosshairs seem to track where the Signal _should_ be, _as if it was plotted but with the same color as the background_.  I used rfft(\signame)[100e3 : 3e8] to assure there was not a zero value in the X axis, but I still get the same result, as if the Signal was plotted in a white color.

Any ideas?

Thanks,
Brian Nelson

mwinkel-dev:
Hi @nelson-zapenergy -- Are you running jScope on a client computer or remotely on the MDSplus server?

Please provide more details about your configuration (operating system, MDSplus version, etcetera) and I'll see if I can reproduce the problem on my development system.

nelson-zapenergy:
Thanks.

I'm running jScope on Windows 11 (MDSplus version "stable_release-7-131-6" from an mdsplus.org download) connecting to an MDSplus server with a compiled version of MDSplus and I get the following:

$ tdic
TDI> MdsShr->MdsRelease:T()
"stable_release-7-153-3"

Note that with Network set to Local, I can make the Y and X both [10 .. 1000], and selecting "Log scale" for Y, e.g. 

<img width="811" height="548" alt="Image" src="https://github.com/user-attachments/assets/64e5c6b8-f849-4506-ab0b-89b3b9c182f4" />

I get:

<img width="1107" height="1086" alt="Image" src="https://github.com/user-attachments/assets/8da3dfe7-9a51-4c2e-8edf-20b529aa3f44" />

With both Y and X selecting "Log scale":

<img width="811" height="548" alt="Image" src="https://github.com/user-attachments/assets/1d2a9a64-f789-4a0a-8fe6-66a4d794aad3" />

I get:

<img width="1107" height="1086" alt="Image" src="https://github.com/user-attachments/assets/5588e3d9-f1c8-4cad-9f59-a0309e60d4b6" />

If I select "Pan" and move the cursor around, I can get this momentarily:

[...]

PR #3041, opened 2026-03-25
Fix: jScope now draws data when have logarithmic X axis

mwinkel-dev:
This fixes issue #3034.

The root cause of the issue was that when the data was being plotted on a log X axis, the logarithm was not being taken of the data points.   The fix was to add a new function, `XOnAxis(double x)`, that checks for the log X axis and does the necessary conversion.   Use of `XOnAxis()` also simplified some other routines.

Because the fix worked well on the X axis, the same design pattern was also applied to the Y axis by adding a new, `YOnAxis(double y)` function.

### DIFF
--- a/java/jscope/src/main/java/mds/wave/WaveformMetrics.java
+++ b/java/jscope/src/main/java/mds/wave/WaveformMetrics.java
@@ -73,14 +73,9 @@ public class WaveformMetrics implements Serializable
 				_ymax = MIN_LOG;
 			if (_ymin < MIN_LOG)
 				_ymin = MIN_LOG;
-			ymax = Math.log(_ymax) / LOG10;
-			ymin = Math.log(_ymin) / LOG10;
 		}
-		else
-		{
-			ymax = _ymax;
-			ymin = _ymin;
-		}
+		ymax = YOnAxis(_ymax);
+		ymin = YOnAxis(_ymin);
 		delta_y = ymax - ymin;
 		ymax += delta_y / 50;
 		ymin -= delta_y / 50.;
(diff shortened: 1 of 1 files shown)


java/jscope/src/main/java/mds/wave/WaveformMetrics.java
```diff
@@ -379,7 +369,7 @@ else if (last_y == max_y)
 					start_x = XPixel(x[j]);
 					max_y = min_y = y[j];
 					i = j;
-					if (sig.isIncreasingX() && x[j] > xmax)
+					if (sig.isIncreasingX() && XOnAxis(x[j]) > xmax)
```
mwinkel-dev, line 372: The loop terminated early for a log X axis because the logarithm was not taken of the data point.   Thus the raw data values quickly exceed the maximum of the log X axis.

java/jscope/src/main/java/mds/wave/WaveformMetrics.java
```diff
@@ -433,8 +423,20 @@ final public double XMin()
...
+			if (x < MIN_LOG)
+				x = MIN_LOG;
+			x = Math.log(x) / LOG10;
+		}
+		return x;
+	}
+
 	final public int XPixel(double x)
 	{
+		x = XOnAxis(x);
```
mwinkel-dev, line 439: Added this line to take the logarithm of the data point when have a log X axis.

mwinkel-dev:
Hi @GabrieleManduchi -- You are the jScope expert, so I defer to whatever you suggest regarding this proposed PR.

java/jscope/src/main/java/mds/wave/WaveformMetrics.java
GabrieleManduchi: It seems OK to me

GabrieleManduchi (approved):
It seems OK to me

mwinkel-dev:
Hi @GabrieleManduchi -- Thanks for the review.

PR #3043, opened 2026-03-25
Build: Handle multiple version tags

WhoBrokeTheBuild:
In preparation for the new release method, we need to support multiple version tags on a single commit
e.g. The `HEAD` of stable would have both `stable_release-1-2-3` and `alpha_release-1-2-3` and we would select the stable one

Update logic in both get_new_version.py and MDSplusProjectVersion.cmake to check all tags first

Remove flavor from deb and rpm wildcards in build.py


mwinkel-dev (approved):
The changes match the description and look good.   Thus approving this PR.

PR #3045, opened 2026-03-27
Feature: add python wheel to distribution

dgarnier:
When using with modern package managers and even opinionated newer versions of Ubuntu, it's useful to have MDSplus python as a wheel.  (For example, with uv, one can specify a directory to find wheels.)  This PR adds tooling to build the wheel and copy it into the `python/`directory.  It also adds some cleanup so the byproducts of building the wheel are not installed.  Finally, the pyproject.toml is updated as `python/MDSplus/tests` are not installed with CMake and a small warning of older license info is corrected.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,2 +1,28 @@
-
+# Add the MDSplus Python package subdirectory
 add_subdirectory(MDSplus)
+
+# Find Python interpreter for building the wheel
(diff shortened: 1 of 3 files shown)


dgarnier:
I made a less intrusive version in #3078 to not have a copy and less possibilities of mismatched python / shared libraries... so this one should close if that one is accepted.

PR #3046, opened 2026-03-30
Build: Explicitly set the RELEASE_TAG when building

WhoBrokeTheBuild:
Remove hacky new_tag file and instead pass the tag as an environment variable to the agents
Move error checking for invalid version to the publish step


mwinkel-dev (approved):
Straightforward change.   Looks good.

PR #3047, opened 2026-03-30
Build: Revert changes for handling multiple tags

WhoBrokeTheBuild:
The new approach uses Jenkins to properly label alpha/stable releases, so complicated parsing of multiple version tags isn't actually needed
Use the GIT_BRANCH instead of the RELEASE_BRANCH to always label builds after the branch you're on
Standardize the command for getting the current tag


mwinkel-dev (approved):
Yes, the new method is simpler.   This is a straightforward change to remove the GIT_TAG method.   Looks good.

PR #3048, opened 2026-03-30
Fix: Increase the MaxConnections for the mdsip.socket from 64

WhoBrokeTheBuild:
There is, unfortunately, no way to disable this, so setting it to 999,999 should hopefully suffice This has been implemented on the server used by the IDL tests, so those are reverted to their original working state

### DIFF
--- a/idl/testing/run_tests.py
+++ b/idl/testing/run_tests.py
@@ -574,168 +574,18 @@ if args.database_name != '':
     ''')
 
 
-    #---------------------------------------------------------------------------
+    # ---------------------------------------------------------------------------
     # https://github.com/MDSplus/mdsplus/issues/2625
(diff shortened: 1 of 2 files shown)


mwinkel-dev (approved):
It is a bit surprising that systemd's socket file has a default of 64 connections.   Regardless, this change to `MaxConnections` effectively eliminates that constraint.

As for the `idl-2625-loop` and `idl-2638-loop` tests, they had been duplicated and dialed back to 50 connections as a temporary workaround for the 64 connection limit.  (The original tests were then commented out.)  Now that the `MaxConnections` has been increased to a huge number, the duplicate tests have been deleted and the original tests uncommented.

These changes all look good.

PR #3056, opened 2026-06-01
Build: Add Ubuntu 26.04

WhoBrokeTheBuild:


mwinkel-dev:
Hi @WhoBrokeTheBuild -- Will we also support Ubuntu 26.04 on arm64?   If so, should arm64 be included in this PR?   Or is it cleaner to have arm64 be a separate PR?

mwinkel-dev:
A second PR for arm64 strikes me as a good approach.   My assumption is that we will want the amd64 version building smoothly for a few days before adding the arm64 build.  (However, it is also A-OK to include arm64 in this PR if you prefer.)

Regarding this PR, I will peruse the existing commits later today.   But will wait for the final commits before posting an official review.

mwinkel-dev (dismissed):
This is a straightforward change, so approving it.

Presently, it is failing on tests 24 and 39, which I'm assuming is just a glitch with the build system.   However, if resolving the two tests involves an additional commit to this PR, then I will re-review it.

heidthecamp (approved):

mwinkel-dev (approved):
The addition of the Ubuntu 26 `*.opts` files looks good.

PR #3068, opened 2026-07-10
Fix: Subscripting empty vectors

WhoBrokeTheBuild:
This removes attempts to access index 0 of empty vectors (and one array) in favor of an explicit NULL
This was found while testing #3056 for Ubuntu 26.04 (Thank you gcc 15)

Note: Lambda should probably be removed in favor of a simple freeDsc loop after the try/catch


mwinkel-dev (approved):
That is an interesting bug!

Anyway, the fix is straightforward and looks good.

santorofer (approved):
Very clear the changes. Looks good.

PR #3074, opened 2026-07-16
Feature: `Add $MDSPLUS_DIR/python` to `PYTHONPATH` in `etc/envsyms`

WhoBrokeTheBuild:
This removes the long-broken `MDSPLUS_SETUP_PY` and instead adds `$MDSPLUS_DIR/python` to the `PYTHONPATH` in `$MDSPLUS_DIR/etc/envsyms`.
This should allow users to `import MDSplus` in python after an install with no additional setup

Also add `import MDSplus` steps to the installer tests


mwinkel-dev (approved):
Simple change.   Looks good.   Approving this because I'm assuming Jenkins will rebuild it.

Note though that one of the "Test Packages" tests failed because it returned exit status 42.

dgarnier (commented):
I agree this is _consistent_ with the current use of PYTHONPATH and MDSplus.  However, I'm not a huge fan of this whole mechanism and at OpenStar we are doing something entirely different and explicitly not relying on PYTHONPATH.. preferring to use venv based  solutions (see my request to have MDSplus build its own wheel during the install instead).  

This change would effectively just have me modify our environment to take it back out... so I'm not a fan of this change.

dgarnier (approved):

dgarnier:
> I'm also not a fan of the wheel solution because it (as I understand it) places a copy of the python files in the venv rather than pointing to the `$MDSPLUS_DIR/python` folder, which makes sense for most packages but I don't like it for ours.
> 

Oh. That gives me an idea.  Let me see if it works.  I think a wheel can be made that does just what you want.. just installs a .pth that points to $MDSPLUS_DIR/python (or the absolute install directory which is probably actually what we want.  I'll send a PR if it works.

PR #3076, opened 2026-07-21
Fix: GetMany queries that succeed with a nil result (issue #3070)

### LINKED ISSUE #3070: mdsip: GetMany batch containing a query that succeeds with a NULL result crashes the server worker (NULL deref in getManyObj)
**Affiliation**
OpenStar Technologies (working against OpenStar's own copy of the LDX archive)

**Version(s) Affected**
Server Version: Alpha 7.160.1 (`alpha_release-7-160-1`). The faulty code path is unchanged on current `alpha` HEAD (726e9451f5b3); the code references below are pinned to the release tag.
Client: mdsthin 1.7.0 (the client only observes the crash; the fault is server-side and client-agnostic — the same TDI expressions are sent by the MDSplus python `Connection.getMany()`)

**Platform(s)**
Client: macOS. Server: Linux (OpenStar mdsip server hosting a copy of the LDX archive).

**Installation Method(s)**
Not relevant to the code path below (reproducible from source).

**Describe the bug**
If a `GetMany` batch contains a query that *succeeds* but evaluates to a NULL/empty result, the mdsip server worker handling the connection dies mid-reply (the reply never arrives; the TCP connection is left half-open). In one instance the whole mdsip service was wedged afterwards and stopped accepting new logins until restarted.

Queries that *fail* inside a GetMany are handled correctly (an `"error"` entry comes back). It is specifically a **successful evaluation yielding a null result** that kills the worker. The same expression sent through a plain `get()` harmlessly returns nil/None.

Concrete example: a node whose record is a Range with null begin/end, e.g. `\LDX::TOP.HARDWARE:DCLOCK_1.CHANNEL_1:CLOCK` in tree `ldx`, shot 100805020, whose record is `* : * : 5000E-9`. Then

```
DATA(BEGIN_OF(GETNCI(\LDX::TOP.HARDWARE:DCLOCK_1.CHANNEL_1:CLOCK,"RECORD")))
```

evaluates successfully to nothing — fine via `get()`, fatal inside `GetMany`.

**To Reproduce**

```python
import mdsthin  # same behavior expected with MDSplus.Connection

c = mdsthin.Connection('SERVER')
c.openTree('ldx', 100805020)  # any LDX shot with the DCLOCK_1 hardware should do

EXPR = 'DATA(BEGIN_OF(GETNCI(\\LDX::TOP.HARDWARE:DCLOCK_1.CHANNEL_1:CLOCK,"RECORD")))'

# 1. Plain get: returns None, no problem
print(c.get(EXPR))

# 2. Same expression inside GetMany: server worker dies, no reply ever arrives
gm = c.getMany()
gm.append('begin', EXPR)
gm.execute()   # hangs forever / connection dead
```

Any node whose record is a Range/expression with a missing part should do; presumably any expression that evaluates successfully to a NULL descriptor (e.g. `DATA(*)`) triggers it too, though we only verified against tree nodes like the one above.

**Expected behavior**
The result dictionary should carry a nil `"value"` for that query (so the client's `GetMany.get(name)` returns None, matching plain `get()`), or at worst an `"error"` entry — never a worker crash.

**Where it crashes (from reading `alpha_release-7-160-1`; identical on current `alpha` @ 726e9451f5b3 apart from a two-line offset in mdsdataobjects.cpp)**

`GetManyExecute` ([tdi/remote/GetManyExecute.fun](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/tdi/remote/GetManyExecute.fun)) → `GetManyExecute()` in [mdsobjects/cpp/mdsdata.c#L914-L926](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsdata.c#L914-L926) → `getManyObj()` in [mdsobjects/cpp/mdsipobjects.cpp#L122-L253](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsipobjects.cpp#L122-L253):

1. For each query, `getManyObj` calls `execute(expr)` / `executeWithArgs(...)` ([mdsipobjects.cpp#L238](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsipobjects.cpp#L238)).
2. `executeWithArgs` ([mdsdataobjects.cpp#L945-L981](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsdataobjects.cpp#L945-L981)) calls `compData->data()`, which evaluates with a *success* status but an **empty XD** when the result is nil ([mdsdataobjects.cpp#L409-L421](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsdataobjects.cpp#L409-L421)).
3. `convertFromDsc` then returns **NULL** ([mdsdata.c#L264-L269](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsdata.c#L264-L269)) — so `execute()` returns a NULL `Data*` without throwing.
4. Back in `getManyObj`, the NULL is stored as the answer: `answDict->setItem(valueKey.get(), currAnsw)` ([mdsipobjects.cpp#L240-L241](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsipobjects.cpp#L240-L241)).
5. `Dictionary::setItem` unconditionally calls `data->incRefCount()` ([include/mdsobjects.h#L2908-L2929](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/include/mdsobjects.h#L2908-L2929)) — NULL dereference, SIGSEGV in the worker.

Failing queries never reach that line — the `MdsException` is caught and turned into an `"error"` entry ([mdsipobjects.cpp#L243-L248](https://github.com/MDSplus/mdsplus/blob/alpha_release-7-160-1/mdsobjects/cpp/mdsipobjects.cpp#L243-L248)) — which is exactly the asymmetry we observe (failures fine, successful-null fatal).

Possible fixes: guard `currAnsw == NULL` in `getManyObj` (store nothing, or an explicit nil, under `"value"`), and/or make `Apd::setItem`/`appendDesc` NULL-safe.

**Additional context**
- Observed while bulk-archiving shots from OpenStar's copy of the LDX tree (batched metadata reads). Worked around client-side by keeping nullable expressions out of GetMany batches.
- A client-side consequence of the worker dying mid-reply: mdsthin clients spin at 100% CPU on the half-open socket — reported separately as MDSplus/mdsthin#8 (recv loop doesn't detect EOF).
- Possibly separate, mentioned for completeness: a GetMany whose *successful* result dictionary is very large (a segmented photodiode signal, ~58 MB data + ~115 MB timebase in one reply) also killed the worker. If desired we can file that separately once we can characterize it better.
- Claude (Anthropic's AI assistant) helped identify the crash path in the source and document this report.

dgarnier:
# Fix: GetMany queries that succeed with a nil result (issue #3070)

## Problem

A `GetMany` query whose expression evaluates to nil (e.g. `GETNCI(<node>, "RECORD")`
of a clock whose record is `* : * : 5E-6` — a Range with a missing begin) crashed
the mdsip worker instead of returning `{ 'value': missing }`, the way a plain
`get()` of the same expression does.

The crash had several layers; the root cause turned out to be broader than the
`Dictionary` accessors that first showed up in a debugger.

## Root cause

The C++ object layer uses a `NULL` `Data *` to mean **missing** (`*`) — it is what
`execute()` returns for a nil expression, and what `convertToApdDsc()` emits as a
null descriptor pointer. The machinery was inconsistent about tolerating that NULL:

1. **`Dictionary::getItem` / `setItem`** dereferenced the stored value without a
   NULL check, so storing/reading a missing value segfaulted. *(fixed in the header)*

2. **`deleteData(Data *)`** dereferenced `data->refCount` with **no NULL guard**.
   This is the real culprit and the one the fix below targets.

`deleteData(NULL)` is reached on ordinary cleanup paths, because `AutoData<T>` — the
RAII wrapper used throughout the mdsip object code — calls `deleteData(ptr)`
unconditionally in its destructor, and it routinely wraps the (nullable) result of
`getItem()`:

- **Server, `getManyObj`** (`mdsipobjects.cpp`): a query may legitimately omit the
  `args` key — the server explicitly supports that (`if (argsData.get() && argsData->len() > 0)`).
  But `AutoData<List> argsData((List *)currArg->getItem(&argsKey))` then wraps the
  `NULL` that `getItem` returns for the absent key, and destroying it at the end of
  each loop iteration calls `deleteData(NULL)`.
- **Client, `GetMany::get()`** (`mdsipobjects.cpp`): for a query that succeeds with a
  nil result, `AutoData<Data> result(dictionary->getItem(&valueKey))` wraps `NULL`, so
  the same `deleteData(NULL)` fires on the way out — i.e. the crash also reproduces on
  a real client reading back a nil value, not just inside the worker.

**Why it never showed up before:** the production C++ `GetMany` client always attaches
an `args` key (an empty `List` when there are no args), so `getItem(&argsKey)` was never
`NULL` in normal use. The new test exercises the args-less path on purpose.

## The fix, and why it lives in `deleteData`

```c
void MDSplus::deleteData(Data *data)
{
  if (!data)      // NULL == missing ("*"): accept it as a no-op
    return;
  ...
}
```

We considered fixing only `getManyObj`, but that would patch one call site and leave
the same defect live on the client (`GetMany::get()`) and at every other
`AutoData<X>(getItem(...))` in the codebase. The NULL reaching `deleteData` is not
specific to `getManyObj` — it is the general contract that `getItem()`/`execute()`
return NULL for "missing" and that `AutoData` is an RAII wrapper you can hand that NULL
to. `getManyObj` is correct as written; it was simply relying on a `deleteData` that
couldn't hold up its end.

So the fix belongs at the shared choke point. `deleteData` is the right one: it matches
C++'s own `delete nullptr` semantics, matches the already-NULL-tolerant helpers in the
same file (e.g. `getMember`), and covers both the `AutoData` destructor and the raw
`deleteData(descs[i])` calls. (`Apd::propagateDeletion` already guards its own elements,
which is exactly the convention we are now making universal.)

## Tests

`mdsobjects/cpp/testing/MdsGetManyTest.cpp` (new) covers three levels:

1. A `Dictionary` accepts a missing (NULL) value — `setItem`/`getItem`/replace round trips.
2. A missing value survives the serialize / deserialize round trip used to transport the
   result dictionary.
3. End-to-end: `GetManyExecute()` on a batch containing a nil-result query returns
   `{ 'value': missing }` (answered, not an error) instead of crashing.

## macOS test-harness changes (required to even see this failure)

Two changes to the Check-based C test harness were needed so that macOS reports crashes
and timeouts the way Linux does — without them, this bug manifested as a silent hang and
the test appeared to "pass" or stall rather than fail:

- **`testing/check_backend.c`** — On macOS a hardware fault (null deref, illegal
  instruction, …) is delivered to the task-level Mach exception port (the system crash
  reporter), which *parks* the faulting thread instead of terminating the process. The
  forked test child then never dies, the parent blocks forever in `waitpid()`, and the
  whole run stalls. In `__setup_child()` we now clear the child's exception ports
  (`task_set_exception_ports(..., EXC_MASK_BAD_ACCESS | EXC_MASK_BAD_INSTRUCTION |
  EXC_MASK_ARITHMETIC, MACH_PORT_NULL, ...)`), so a fault falls through to the default
  action and a crashing test dies promptly and is reported, just like on Linux.

- **`testing/backends/check/lib/macos_timer.h`** — macOS lacks POSIX `timer_create`
  delivery, so the harness emulates it with a dispatch timer. The default expiration
  handler previously called `signal(SIGALRM, NULL)`, which only reset the SIGALRM
  disposition and never actually raised the signal — so test timeouts silently never
  fired on macOS and a hung child stalled the runner indefinitely. It now does
  `kill(getpid(), SIGALRM)` to emulate real POSIX-timer delivery, so the harness'
  SIGALRM handler runs and kills the timed-out child.

Together these make a segfault/hang in a test surface as a reported failure on macOS
instead of a silent stall — which is how the remaining `deleteData` crash was found
after the header guards were added.

Fixes #3070.

### DIFF
--- a/include/mdsobjects.h
+++ b/include/mdsobjects.h
@@ -2892,13 +2892,17 @@ namespace MDSplus
       dtype = DTYPE_DICTIONARY;
     }
 
+    // A NULL value represents missing ("*"): it is what execute() returns
+    // for an expression that evaluates to nil, and convertToApdDsc() emits
+    // it as a null descriptor pointer
     Data *getItem(String *strData)
     {
       for (std::size_t i = 0; i < descs.size(); i += 2)
       {
         if (strData->equals(descs[i]))
         {
-          descs[i + 1]->incRefCount();
+          if (descs[i + 1])
+            descs[i + 1]->incRefCount();
           return descs[i + 1];
         }
       }
@@ -2915,9 +2919,11 @@ namespace MDSplus
           descs[i] = strData;
           strData->incRefCount();
 
-          descs[i + 1]->decRefCount();
+          if (descs[i + 1])
+            descs[i + 1]->decRefCount();
           descs[i + 1] = data;
-          data->incRefCount();
+          if (data)
+            data->incRefCount();
           return;
         }
       }
@@ -2925,7 +2931,8 @@ namespace MDSplus
       descs.push_back(strData);
       descs.push_back(data);
       strData->incRefCount();
-      data->incRefCount();
+      if (data)
+        data->incRefCount();
     }
 
     std::size_t len() { return Apd::len() / 2; }
--- a/mdsobjects/cpp/mdsdataobjects.cpp
+++ b/mdsobjects/cpp/mdsdataobjects.cpp
@@ -330,6 +330,11 @@ Data::~Data()
 
 void MDSplus::deleteData(Data *data)
 {
+  // A NULL Data represents missing ("*"). deleteData() must accept it as a
+  // no-op: RAII wrappers (AutoData) and Apd::propagateDeletion() hand it the
+  // NULL that getItem()/execute() return for a nil result.
+  if (!data)
+    return;
   if (data->refCount <= 1)
   {
     if (data->units)
--- a/mdsobjects/cpp/testing/CMakeLists.txt
+++ b/mdsobjects/cpp/testing/CMakeLists.txt
@@ -59,6 +59,7 @@ set(_test_source_list
     MdsEventSuppression.cpp
     MdsEventTest.cpp
     MdsConnectionTest.cpp
+    MdsGetManyTest.cpp
 
     # SKIP_TEST is not supported
     # MdsCallTest.cpp
--- a/mdsobjects/cpp/testing/MdsGetManyTest.cpp
+++ b/mdsobjects/cpp/testing/MdsGetManyTest.cpp
@@ -1,0 +1,137 @@
+/*
+Copyright (c) 2026, Massachusetts Institute of Technology All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+Redistributions of source code must retain the above copyright notice, this
+list of conditions and the following disclaimer.
+
+Redistributions in binary form must reproduce the above copyright notice, this
+list of conditions and the following disclaimer in the documentation and/or
+other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+#include <stdlib.h>
+
+#include <mdsdescrip.h>
+#include <mdsobjects.h>
+
+#include "testing.h"
+#include "testutils/unique_ptr.h"
+
+using namespace MDSplus;
+using namespace testing;
+
+// The mdsip server entry point for Connection.getMany() batches
+// (tdi/remote/GetManyExecute.fun -> MdsObjectsCppShr->GetManyExecute)
+extern "C" struct descriptor_xd *GetManyExecute(char *serializedIn);
+
+////////////////////////////////////////////////////////////////////////////////
+//  GetMany results that evaluate to missing (issue #3070)  ////////////////////
+////////////////////////////////////////////////////////////////////////////////
+
+int main()
+{
+  BEGIN_TESTING(GetMany);
+
+  { // A Dictionary must accept a missing (NULL) value: getManyObj() stores
+    // one whenever a query succeeds with a nil result. The rest of the Apd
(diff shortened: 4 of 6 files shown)


joshStillerman (approved):
Great Fix (same as mine but more thorough)

PR #3078, opened 2026-07-26
Feature: pip-installable MDSplus "local" wheel

dgarnier:
As promised in the discussion about #3074, I had an idea for a re-direct python wheel. 
The idea is just build a wheel that does the minimum and make a .pth.  Claude and
I got together and got a little more clever than that, so we don't need to run python
packaging tools during the build.

An optional build step (`ENABLE_PYTHON_LOCAL_WHEEL`, default on, easy to turn
off) that writes a small MDSplus wheel into `<prefix>/python/wheelhouse`. The
wheel is a redirect, not a copy: its only payload is a `MDSplus.pth` that adds
this install's `python/MDSplus` to a Python environment.

This is a nice-to-have, not a need-to-have. The usual `PYTHONPATH=$MDSPLUS_DIR/python`
works fine for most uses.  This method adds some other niceties.

- `pip install MDSplus --find-links $MDSPLUS_DIR/python/wheelhouse` into a venv pulls in `numpy` automatically, which a bare
  `PYTHONPATH` does not;
- another project can list `MDSplus` as a dependency and resolve it from the
  wheelhouse (`pip`/`uv --find-links`) rather than wiring `PYTHONPATH` in each
  project or CI job.

Kept deliberately small and self-contained so it's cheap to maintain or ignore:
one `python/local_wheel/` subdir, a stdlib-only `make_wheel.py` (no build
backend, no network), and the `.pth`/bootstrap. Nothing else depends on it, and
it's behind a flag. Metadata and dependencies are copied from
`python/MDSplus/pyproject.toml` so `pip show` reports the real package.

Behavior of the redirect at runtime:

- startup stays minimal — set `MDSPLUS_DIR` to the install the wheel was built
  for (or `$MDSPLUS_DIR` if it was relocated) and prepend its `python/` to
  `sys.path`;
- on the first `import MDSplus`, if needed, a one-shot hook sources that
  install's `setup.sh` to populate the environment (handy for pointing a
  throwaway venv at a specific build when testing);
- it never spawns anything at plain startup, is wrapped so it can't break the
  interpreter, and `MDSPLUS_LOCAL_WHEEL_DISABLE=1` disables it.

Draft while it gets more testing. See `python/local_wheel/README.md` for usage.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -1,2 +1,12 @@
 
 add_subdirectory(MDSplus)
+
+mdsplus_option(
+    ENABLE_PYTHON_LOCAL_WHEEL BOOL
+    "Build a pip-installable 'local' wheel whose only payload is a .pth file that points at the installed python/MDSplus directory."
+    DEFAULT ON
+)
+
+if(ENABLE_PYTHON_LOCAL_WHEEL)
+    add_subdirectory(local_wheel)
+endif()
--- a/python/local_wheel/CMakeLists.txt
+++ b/python/local_wheel/CMakeLists.txt
@@ -1,0 +1,40 @@
+# Build the MDSplus "local" wheel -- a pip-installable wheel whose only payload
+# is a MDSplus.pth (+ a _mdsplus_bootstrap.py it imports) that makes
+# `import MDSplus` resolve to the system-installed python/MDSplus package.
+#
+# The wheel is built at INSTALL time (install(SCRIPT) below) rather than build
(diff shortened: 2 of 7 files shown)


dgarnier:
Testing this patch with homebrew here:  

https://github.com/dgarnier/homebrew-plasma/releases/tag/mdsplus-alpha_release-7-160-2_1

can install on macOS with:

```
brew install dgarnier/plasma/mdsplus
```

dgarnier:
This has been working well for us... makes it easy to use in our own packages.

